PR #82, opened 2014-12-02
added @EnableDiscoveryClient, deprecated @EnableEurekaClient

spencergibb:
@dsyer What do you think of the abstraction for the mvp?  If we do, I'd officially like to add https://github.com/spring-cloud/spring-cloud-commons to the mvp and delete the s-c-client.\* from this project.


spencergibb:
@wgorder I think the docs may just need to be regenerated.

PR #115, opened 2014-12-22
Make zuul proxy route filter retryable

VanRoy:
Hello,

After make any tests on RC1 I have discovered that the Zuul proxy route does not implement the Ribbon retry.
If any instance registered in Eureka do not respond the proxy fail.

This commit activate the retry handler on for the RibbonCommand on ZuulProxy.

Do you think that is it necessary to make this feature configurable ( For example in a  zuul.routes.XXX.retriable properties ) ?
I can update my PR with this modification if you want.

Thanks a lot.
Julien.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
@@ -88,6 +88,7 @@ public class ProxyRouteLocator implements RouteLocator {
 		String targetPath = null;
 		String id = null;
 		String prefix = this.properties.getPrefix();
+		Boolean retryable = this.properties.getRetryable();
 		for (Entry<String, ZuulRoute> entry : this.routes.get().entrySet()) {
 			String pattern = entry.getKey();
 			if (this.pathMatcher.match(pattern, path)) {
@@ -106,11 +107,14 @@ public class ProxyRouteLocator implements RouteLocator {
 						prefix = prefix + routePrefix;
 					}
 				}
+				if(route.getRetryable() != null) {
+					retryable = route.getRetryable();
+				}
 				break;
 			}
 		}
 		return (location == null ? null : new ProxyRouteSpec(id, targetPath, location,
-				prefix));
+				prefix, retryable));
 	}
 
 	public void resetRoutes() {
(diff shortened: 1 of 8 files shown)


dsyer:
I don't think this is a good idea since it short circuits (unconditionally) the configuration a user might have set up for the `LoadBalancer` (e.g. my understanding is `ribbon.OkToRetryOnAllOperations=true` would have the same effect by default but would be overridable per client).

VanRoy:
Dave,

Even with "ribbon.OkToRetryOnAllOperations=true" property, the POST requests are not retriable in Zuul proxy.

In fact, the retriable boolean of Ribbon HttpRequest don't override user configuration.
On the RestClient Ribbon implementation, the retriable is a pre-requisite for retry handler.

It's visible on this line : https://github.com/Netflix/ribbon/blob/2.0-RC13/ribbon-httpclient/src/main/java/com/netflix/niws/client/http/RestClient.java#L709 

If we don't want hardcode httpRequest.setRetriable, we may add a new property like "zuul.routes.XXX.retriable properties"

What do you think ?

Thanks a lot.
Julien.

spencergibb:
based on that line in `RestClient.java` I think a configuration property would be fine.

dsyer:
Wow, that's a big change for a teeny tiny boolean property. Is it just me or is it really irritating having it spelled wrong? Can we call it "retryable" (even though the native Netflix property names are "retriable")?

vsafin-copart:
Is retry feature going to be available in RELEASE ?

dsyer:
If someone proposes a change where it is spelled correctly, possibly. To be perfectly clear: the retryable feature is there in Ribbon already, and we are only proposing to make it configurable using Spring Boot properties.

VanRoy:
Hi Dave,

I have pushed a spelling fix for this PR.

But I just wanted to clarify that this PR it's not only for make the retryable configurable.
It's mostly for fix the POST request that is not retryable in the current state.
Even with "ribbon.OkToRetryOnAllOperations=true" property, the POST requests are not retryable in the Zuul proxy.

In fact, the "retriable" boolean of Ribbon HttpRequest don't override user configuration.
On the RestClient Ribbon implementation, the "retriable" is a pre-requisite for retry handler ( expect for GET method that is retryable by default).

It's visible on this line : https://github.com/Netflix/ribbon/blob/2.0-RC13/ribbon-httpclient/src/main/java/com/netflix/niws/client/http/RestClient.java#L709

I really am desolated to make you lose a lot of time with this subject, it 's probably due to my English that is really bad.

Julien.

Issue #126, opened 2014-12-31
Turbine AMQP aggregator fails to de-serialize messages from Hystrix stream


ogunheper:
Turbine's new AMQP aggregation feature fails to read messages from queue with JsonParseException.

Deserialization fails since AmqpInboundChannelAdapter uses a SimpleMessageConverter. However, the message is in JSON format with "__TypeId__" being "java.lang.String". Issue can be fixed with using a message converter that supports JSON messages, instead of the simple one...

Full stack trace:

```
[...]

PR #127, opened 2014-12-31
Jackson2JsonMessageConverter added to Turbine amqp auto-configuration

ogunheper:
Message converter is not marked as a @Bean, since the only use is within the configuration class itself. This corresponds with the usage in org.springframework.cloud.bus.amqp.AmqpBusAutoConfiguration.java class.
No container provided ObjectMapper is used for the message converter, as this a JSON to java.lang.String converter.

Fixes gh-126


dsyer:
Sorry didn't see this. Fixed in 8f8e738.

PR #145, opened 2015-01-14
Polish

philwebb:
I'm not sure if you're interested in apply the same eclipse rules that Spring Boot uses or not? We've found them quite useful to enforce consistent formatting (but not everyone likes auto-formatted code).

This PR applies the rules and also provides some additional polish commits.

Fixes the following issues:

* Consistent use of tabs indentation (previously there was a mix)
* Apply cleanup rules such as using `this.` for field access
* Add copyright headers to all files
* Use BDD flavor of Mockito
* General formatting polish and tweaks


spencergibb:
Yeah, a lot of this was before I learned about eclipse formatter for intellij AND that it is per project, never went back and reformatted everything.

eclipse/eclipse-code-formatter.xml
```diff
@@ -0,0 +1,295 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
spencergibb: We reference this [here](https://github.com/spring-cloud/spring-cloud-netflix#code-conventions-and-housekeeping) and have [one in spring-cloud-build](https://github.com/spring-cloud/spring-cloud-build/blob/master/spring-cloud-build/eclipse-code-formatter.xml).  Maybe the docs should link to that one if they don't already.

Issue #185, opened 2015-01-27
Change ribbon to use LoadBalancerBuilder rather than setting class names in properties


spencergibb:
See [EurekaRibbonClientConfiguration.java](https://github.com/spring-cloud/spring-cloud-netflix/blob/84554725db0d8560ebc0cb4320888b70c8700868/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/eureka/EurekaRibbonClientConfiguration.java#L83-L92) for setting properties.

See [ConsulLoadBalancerClient](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLoadBalancerClient.java#L41-L48) for using LoadBalancerBuilder.

PR #186, opened 2015-01-28
use LoadBalancerBuilder so all parts of the LoadBalancer can be injected...

spencergibb:
... via Spring.

 This includes support for: ServerList, ServerListFilter, IRule, IPing and IClientConfig.
 Eliminates custom wrapping of ServerList.

 fixes gh-185

@dsyer will you do a quick review.  I want to make sure I didn't mess things up with the new RibbonClient stuff.  The integration tests work properly, overriding where needed with eureka.  I tested with s-c-s/tests (though I don't know if that helps), customer-stores, feign-eureka and the consul sample app.


dsyer:
Looks good to me. Maybe some updates to the docs needed?

Issue #218, opened 2015-02-19
Please allow for registering custom PolledConfigurationSource with Archaius


digitaljoel:
We have a custom PolledConfigurationSource to get our custom configuration system hooked in to Archaius.  With the ArchaiusAutoConfiguration I don't see a way to add our configuration source to the configuration setup in there.

dsyer:
Archaius is fairly primitive (lots of static singletons) so you can probably do it however you would in a native netflix app? Did you try something obvious and it failed? More details would help.

digitaljoel:
Sorry for the brevity in the initial submission.

In my application I attempted to register our configuration source with ConfigurationManager.install  before calling SpringApplication.run (within my main method).  On doing that it complained that when it attempted to do the ArchaiusAutoConfiguration that a configuration is already registered and the app failed to start.  I was hoping that there would a be a way to tell the auto configuration to also add in some custom list of configurations via injecting beans of the appropriate type and registering them if any are configured, or something like that.

Another option would be to simply check for an already-registered config in the ArchaiusAutoConfiguration and then conditionally install or append, such as:

      if ( ConfigurationManager.isConfigurationInstalled()) {
        ConfigurationManager.getConfigInstance().append(config);
      }
      else {
        ConfigurationManager.install(config);
      }

That would allow anyone to register their configuration as needed, then still use the auto configuration provided in spring-cloud-netflix-core.

icemanftg:
I agree with @digitaljoel on the check for ArchaiusAutoConfiguration part. Otherwise someone has to manually append the configuration after the spring context has been loaded.

dsyer:
Contributions would be more than welcome.

icemanftg:
I've just tried the approach above and it kinda solves the problem, the db properties are pooled by archaius, on the other hand, if there is another system source before the ArchaiusAutoConfiguration (with this modification) finishes, the properties from the config.properties file are not pooled anymore. 
```
// below come from ConfigurationManager.createDefaultConfigInstance()
DynamicURLConfiguration defaultURLConfig = new DynamicURLConfiguration();
try {
	config.addConfiguration(defaultURLConfig, URL_CONFIG_NAME);
}
catch (Throwable ex) {
	log.error("Cannot create config from " + defaultURLConfig, ex);
}
```
[...]

PR #238, opened 2015-03-02
Allowing to add Archaius configuration before the ArchaiusAutoConfiguration is executed

icemanftg:
This fix is in regard to https://github.com/spring-cloud/spring-cloud-netflix/issues/218

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
@@ -24,6 +24,7 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure</artifactId>
 		</dependency>
+
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-commons</artifactId>
@@ -147,6 +148,16 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-jpa</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.h2database</groupId>
+			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.netflix.archaius;
 
 import java.lang.reflect.Field;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.annotation.PreDestroy;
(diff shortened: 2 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
```diff
@@ -145,8 +139,7 @@ protected void configureArchaius(ConfigurableEnvironmentConfiguration envConfig)
 			DynamicURLConfiguration defaultURLConfig = new DynamicURLConfiguration();
 			try {
 				config.addConfiguration(defaultURLConfig, URL_CONFIG_NAME);
-			}
-			catch (Throwable ex) {
+			} catch (Throwable ex) {
```
spencergibb: Please use the [code formatting conventions](https://github.com/spring-cloud/spring-cloud-netflix/#code-conventions-and-housekeeping-1).

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
```diff
@@ -85,7 +81,7 @@ public ConfigurableEnvironmentConfiguration configurableEnvironmentConfiguration
 
 	@Configuration
 	@ConditionalOnClass(Endpoint.class)
-	protected static class ArchaiusEndpointConfuguration {
+	protected static class ArchaiusEndpointConfiguration {
```
spencergibb: :-)

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
```diff
@@ -165,14 +158,30 @@ protected void configureArchaius(ConfigurableEnvironmentConfiguration envConfig)
...
-		}
-		else {
+			addArchaiusConfiguration(config);
+		} else {
 			// TODO: reinstall ConfigurationManager
 			log.warn("Netflix ConfigurationManager has already been installed, unable to re-install");
 		}
 	}
 
+	private void addArchaiusConfiguration(ConcurrentCompositeConfiguration config) {
```
spencergibb: I think I would prefer spring to manage how configurations get injected rather than Archaius.

icemanftg: Can you give me more details on this (or an example :-) ) ? This method just extended the functionality of the ConfigurationManager.install(..) that was previously called from method configureArchaius(...) line 168

spencergibb: In your test you use `ConfigurationManager.install(abstractConfiguration);`.  I would rather inject a Collection of configuration beans and add them in this class rather than relying on the archaius static methods.

digitaljoel:
iceman, thanks for this. I was hoping to get to looking at it but hadn't had time.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
icemanftg: I see, I will look over it. As I see I don't think there will be a way to restrict the use of static archaius methods, but the collection of injected configuration beans should work also.

spencergibb: No there is isn't a way to restrict static usage, but almost all of the v1 netflix tools have them and we provide a "spring way" of consuming them.  I'd like to continue that.  Version 2 of the netflix tools move away from statics anyway, so when they do, the user experience here wouldn't have to change.  See https://github.com/Netflix/archaius/pull/216

icemanftg:
@spencergibb I've autwired a list of AbstracConfigurations for aditional configurations. As archaius does not have a single interface to extend AbstracConfiguration from apache, when they remove apache config from the project, we'll have to also adapt.

@digitaljoel Glad to help.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfiguration.java
```diff
@@ -40,18 +44,7 @@
...
-import com.netflix.config.DynamicURLConfiguration;
-
-import static com.netflix.config.ConfigurationBasedDeploymentContext.DEPLOYMENT_APPLICATION_ID_PROPERTY;
-import static com.netflix.config.ConfigurationManager.APPLICATION_PROPERTIES;
-import static com.netflix.config.ConfigurationManager.DISABLE_DEFAULT_ENV_CONFIG;
-import static com.netflix.config.ConfigurationManager.DISABLE_DEFAULT_SYS_CONFIG;
-import static com.netflix.config.ConfigurationManager.ENV_CONFIG_NAME;
-import static com.netflix.config.ConfigurationManager.SYS_CONFIG_NAME;
-import static com.netflix.config.ConfigurationManager.URL_CONFIG_NAME;
+import com.netflix.config.*;
```
spencergibb: Again, please follow our code conventions.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/archaius/ArchaiusAutoConfigurationTests.java
```diff
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.netflix.archaius;
 
+import static org.junit.Assert.*;
```
spencergibb: code conventions again

icemanftg:
@spencergibb  I've followed the convetions from [here](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style). 
If it is ok with you, I'd like to add the url tot the README.md file.

spring-cloud-netflix-core/pom.xml
```diff
@@ -149,5 +150,15 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
```
spencergibb: we use tabs, not spaces.

spencergibb:
merged 611c0b5c1c425f29e6492744e88fc4fe79d81e4d

Issue #271, opened 2015-03-20
Zuul  - ProxyRouteLocator does not use ZuulProperties servletPath


aivans:
This related to the big files transfer scenario where the DispatcherServlet is bypassed.
Basically, although the unit tests pass for Spring Cloud Zuul, the routing does not work in real scenarios because the ```servletPath``` is not set to the value held in ```ZuulProperties.getServletPath()```.

I will provide a pull request.

dsyer:
I didn't follow that sorry. What is a "real scenario"? The pull request appears to just break a test.

aivans:
Please see the comment on the pull request.

PR #272, opened 2015-03-20
fix for #271

aivans:


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/FormZuulServletProxyApplicationTests.java
```diff
@@ -65,7 +65,7 @@
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringApplicationConfiguration(classes = FormZuulServletProxyApplication.class)
 @WebAppConfiguration
-@IntegrationTest({ "server.port:0", "zuul.routes.simple:/zuul/simple/**" })
+@IntegrationTest({ "server.port:0", "zuul.routes.simple:/simple/**" })
```
spencergibb: Why change this test?

aivans: The context is the big files transfer through Zuul.
See https://github.com/spring-cloud/spring-cloud-netflix/issues/254#issuecomment-83958338

My problem is that the ```/zuul``` prefix is not set in the ProxyRouteLocator, meaning it will not be stripped from the path, so no path will be found.

The changed was written gfor this scenarion, but in my opinion the tested route should be declared without the ```/zuul``` prefix.

dsyer:
But this is a test specifically for the /zuul path. I still don't understand why we would change it.

aivans:
Hi Dave,

I don't know how to explain this better.
The route is actually ```/simple/**```, and the prefix comes from ZuulProperties.
So when a request is made to ```/zuul/simple/**```, ```/zuul``` is not part of the routing settings, it's just a prefixed introduced by the special code for handling big file requests.

You normally can have requests directly to ```/simple/**`` and still work with the Dispatcher Servlet.

dsyer:
> You normally can have requests directly to /simple/** and still work with the Dispatcher Servlet.

I know, but if we changed the test to go to that path it would not be testing the `ZuulServlet` route would it?

aivans:
Hi,

A route `/simple/**` is expected to work using in 2 cases:
* the normal flow by making requesting `host:port/simple/**`
* the Zuul flow by making requests to `host:port/zuul/simple**`

Bot of the above use the same routes configuration, basically the same `ProxyRouteLocator` instance.

My assumption is that `zuul.routes.simple:/zuul/simple/**` configures a route: anything that comes on `/zuul/simple/**` is routed to `simple`.
Now, I wonder how does this work in the other scenario described above where requests to `/simple` should also work.
So, in my opinion the route is: `/simple/**` and `/zuul` is the Zuul servlet context path which needs to be added as a prefix, but should not affect the actual routing table.

dsyer:
I don't follow you, sorry. My understanding is that `/simple/**` is a route that is served by 2 paths (so `/simple/` and `/zuul/simple/` are basically aliases and where they are routed to depends on the configuration). Please can you carefully explain what you have configured for `zuul.routes.*`, what you expected to happen, and what happened instead?

aivans:
```
zuul:
  ignored-services: "*"
  prefix: /api
  routes:
    upload-sample:
      path: /upload/**
      stripPrefix: false                   
```
Upload-Sample  is the origin service for uploading files.
Requests do work to `http://localhost:8765/api/upload`, but getting OOME for big files, see #254.
Requests to `http://localhost:8765/zuul/api/upload` are supposed to work, but getting the error `No route found for uri: /zuul/api/upload`

The reason this is happening is that config property `ZuulProperties.servletPath` is not injected in the `ProxyRouteLocator`.
My pull request handles exactly this:
```
if (StringUtils.hasText(servletPath)) { // a servletPath is passed explicitly
this.servletPath = servletPath;
} else {
//set Zuul servlet path
this.servletPath = properties.getServletPath() != null? properties.getServletPath() : "";
}
```

spencergibb:
I think I understand now.  The route shouldn't have `/zuul` in it because that comes from the `servletPath`.  The test still tries `/zuul/*` to test the servlet and large files.

aivans:
yes

dsyer:
OK, I see now. Did you fill in the contributors agreement (link in README)?

spencergibb:
@aivans thanks for being patient with us.

aivans:
:) no worries, glad I can help.

I will fill in the agreement in a few hours.

aivans:
Filled in the agreement.

PR #282, opened 2015-03-25
add support for ribbon http client in RestTemplate.

spencergibb:
fixes gh-280

@dsyer what do you think of using the ribbon http client here?  Maybe we should only use it if it is on the classpath?


dsyer:
`public static class A`?

Anyway, could we avoid the ugly `ThreadLocal` by pushing the code from the interceptor down into this request factory?

spencergibb:
@dsyer The request factory only has access to the resolved url.  I need to know the serviceId.  Any idea on how to get it at this point?

PR #298, opened 2015-04-11
Exlude mockito all

joshiste:
However turbine and zuul are depending on mockito-all in compile-scope. Not only I want to get rid of these (usually) test-dependencies, but also mockito-all contains some hamcrest-classes (in older versions) getting in conflict with the classes from spring-test


spencergibb:
@joshiste have you signed the [contributor agreement](https://support.springsource.com/spring_committer_signup)?

joshiste:
I did. confirmation number is: 93620140928055712

Issue #332, opened 2015-05-04
Zuul: SimpleHostRoutingFilter does not allow SSL connections


steve-oakey:
The HTTP client in SimpleHostRoutingFilter does not allow SSL connections to proxied entities if the keystore and truststore is set using the javax.net.ssl system properties. The SSLContext is built using a custom SSLSocketFactory which provided no key or trust store information. It also uses deprecated API to build the HttpClient.

spencergibb:
closed via f094ecfd889be7c22d26e939fa0650b6b6b3642e

PR #333, opened 2015-05-04
HttpClient is build using updated API. Support for SSL

steve-oakey:
The HttpClient is now built using non-deprecated API.

This version also uses the system properties to build the SSLContext.
This allows secure connections between the proxy and client. The
previous version would not allow SSL connections (failed hostname
verification) and did not allow a keystore or truststore to be added via
javax.net.ssl parameters.

fixes gh-332

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/RibbonLoadBalancer.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/RibbonLoadBalancer.java
@@ -21,6 +21,8 @@ import java.net.URI;
 import java.util.Collection;
 import java.util.Map;
 
+import org.springframework.web.util.UriComponentsBuilder;
+
 import com.netflix.client.AbstractLoadBalancerAwareClient;
 import com.netflix.client.ClientException;
 import com.netflix.client.ClientRequest;
@@ -47,12 +49,15 @@ public class RibbonLoadBalancer
 	private final int readTimeout;
 
 	private final IClientConfig clientConfig;
+	
+	private final boolean secure;
 
 	public RibbonLoadBalancer(Client delegate, ILoadBalancer lb,
 			IClientConfig clientConfig) {
 		super(lb, clientConfig);
 		this.setRetryHandler(RetryHandler.DEFAULT);
 		this.clientConfig = clientConfig;
+		this.secure = clientConfig.get(CommonClientConfigKey.IsSecure);
 		this.delegate = delegate;
 		this.connectTimeout = clientConfig.get(CommonClientConfigKey.ConnectTimeout);
 		this.readTimeout = clientConfig.get(CommonClientConfigKey.ReadTimeout);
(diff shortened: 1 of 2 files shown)


spencergibb:
Does ssl still work without a custom kestore or truststore?

steve-oakey:
I will verify tomorrow when I have some time, but I think SSL only worked
before because the HTTP client that was built had no host name verification
and no trust store. So the HTTP connection would trust anything including
self signed certificates. I preface this by saying I am not an expert in
SSL but I do deal with custom key stores and trust stores quite a bit at
work as all of our severs use certificates signed by our own CA.

Ideally, the filter would be configurable so that someone could provide
their own HTTP client implementation and decide whether they want to trust
self signed certificates or rely on a trust store and strict host name
verification.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -332,82 +299,43 @@ public void run() {
...
+									.setConnectTimeout(CONNECTION_TIMEOUT.get())
+									.setCookieSpec(CookieSpecs.IGNORE_COOKIES).build())
+					.setRetryHandler(new DefaultHttpRequestRetryHandler(0, false))
+					.setRedirectStrategy(new RedirectStrategy() {
+
+						@Override
+						public boolean isRedirected(HttpRequest request,
+								HttpResponse response, HttpContext context)
+								throws ProtocolException {
+							// TODO Auto-generated method stub
```
spencergibb: remove auto-gen comments

spencergibb:
By default the `hostnameVerifier = SSLConnectionSocketFactory.BROWSER_COMPATIBLE_HOSTNAME_VERIFIER;` 

To replicate what was here previously we need to allow users to do something like
```
 					.setHostnameVerifier(
							SSLConnectionSocketFactory.ALLOW_ALL_HOSTNAME_VERIFIER)
					.setSSLSocketFactory(new AcceptAllSocketFactory())
```

Maybe using some `Autowired(required = false)` fields and then only set the values in the builder if they are not null?

spencergibb:
Or maybe an optional customizer interface that is passed the builder before `build()` is called.

steve-oakey:
I agree with your proposal. Properties that allow customization of the
HttpClient would The problem is, though, is the HttpClient is a static
AtomicReference and newClient() is static. So we cannot use @Autowired
properties as there is no way to autowire static variables and no way to
access configuration data passed in a contructor.

I am not yet familiar enough with Spring Cloud Netflix or Zuul to
understand the need to have it created as a static AtomicReference so at
this time I am very reluctant to try to refactor this class to allow that
sort of customization.

- Steve

steve-oakey:
Sorry, meant to say "Properties that allow customization of the HttpClient would solve this issue."

spencergibb:
This is what I was thinking cdecf1fef43c263bf2e47591e942d482e01cac52.

steve-oakey:
This is much better. One suggestion would be to make ```newClient()``` protected so that if one wanted to, they could extend the class and configure their own HttpClient from scratch. It may never be necessary but as a consumer I would like that option.

steve-oakey:
Will work off https://github.com/spring-cloud/spring-cloud-netflix/commit/cdecf1fef43c263bf2e47591e942d482e01cac52 instead of leaving this open.

PR #334, opened 2015-05-04
Reason why Eureka instance is not found, is not visible

marcingrzejszczak:
With the following configuration at `bootstrap.yaml`

```yaml

spring:
  application.name: spring-cloud-hystrix-dashboard
  profiles.active: prod
  cloud:
    config:
      discovery:
        enabled: true
        serviceId: spring-cloud-server-config-with-eureka
server.port: 9003

turbine:
  appConfig: some-service
  clusterNameExpression: 'default'
```

I wasn't able to see why Turbine can't find the instance. The reason was related to SPEL:

```
org.springframework.expression.spel.SpelEvaluationException: EL1008E:(pos 0): Property or field 'default' cannot be found on object of type 'com.netflix.appinfo.InstanceInfo' - maybe not public?
```

Anyways there was nothing there in the logs so maybe it should not be silenced like right now?

### DIFF
--- a/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/EurekaInstanceDiscovery.java
+++ b/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/EurekaInstanceDiscovery.java
@@ -136,7 +136,7 @@ public class EurekaInstanceDiscovery implements InstanceDiscovery {
 			}
 		}
 		catch (Exception e) {
-			log.info("Failed to retrieve instances from Eureka");
+			log.warn("Failed to retrieve instances from Eureka", e);
 		}
 		return instances;
 	}


spencergibb:
@marcingrzejszczak `clusterNameExpression` is not needed, but you do need `turbine.aggregator.clusterConfig=SOME-SERVICE`. [see the docs](http://projects.spring.io/spring-cloud/docs/1.0.1/spring-cloud.html#_turbine).  Have you signed the [contributor agreement](https://support.springsource.com/spring_committer_signup)?

marcingrzejszczak:
Hi @spencergibb :) I saw the docs and picked this snippet

```
To use the "default" cluster for all apps you need a string literal expression (with single quotes):

turbine:
  appConfig: customers,stores
  clusterNameExpression: 'default'
```

and then I saw this: http://stackoverflow.com/questions/28365202/spring-boot-eureka-server-hystrix-with-turbine-empty-turbine-stream/29029202#29029202

and this comment

> I must admit, that I didn't believe that using new String('default') might change, but it actually did the trick :) Thank you @DaveSyer for your help, I can now move forward and put myself into new troubles :) –  Szymon Stepniak

but most likely I've made a mistake somewhere.

As for the `contributor agreement` I have just signed it ;)

PR #341, opened 2015-05-07
Add ignoredFilters properties functionality remove default filters.

ArekCzarnik:
Hi *,

current is not possible to disable default filters in zuul, the default filters imo are hardwired at initialization time. 

After the this pull request you have the possibility to add an ignore list to the properties, wich influence the initialization of the default filters and disable some of that. 

example:
```
zuul:
 ignoreFilters:
     - sendResponseFilter
 routes:
   users:
     path: /**
     url: http://localhost:8080
```
Thanks,
Arek

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,17 +66,17 @@
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-commons</artifactId>
-				<version>1.0.3.BUILD-SNAPSHOT</version>
+				<version>1.0.2.BUILD-SNAPSHOT</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-config-client</artifactId>
-				<version>1.0.3.BUILD-SNAPSHOT</version>
+				<version>1.0.2.BUILD-SNAPSHOT</version>
 			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-config-server</artifactId>
-				<version>1.0.3.BUILD-SNAPSHOT</version>
+				<version>1.0.2.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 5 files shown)


spencergibb:
@ArekCzarnik I like the idea, but have some issues with the implementation.  Why create the filter bean just to ignore it?  I'd prefer something like this:

```
@Bean
@ConditionalOnProperty(name = "zuul.filter.formBodyWrapper.enabled", matchIfMissing = true)
public FormBodyWrapperFilter formBodyWrapperFilter() {
   return new FormBodyWrapperFilter();
}
```

That way you could set `zuul.filter.formBodyWrapper.enabled=false` and for each filter you don't want.

ArekCzarnik:
thanks, your prefer way looks better... i will try refactor my stuff.

ArekCzarnik:
@spencergibb change the code to all filters with ConditionalOnProperty, the test is getting simple. 

what i found is  in the netflix stuff : 
```
public abstract class ZuulFilter implements IZuulFilter, Comparable<ZuulFilter> {
 
 private final DynamicBooleanProperty filterDisabled = DynamicPropertyFactory.getInstance().getBooleanProperty(this.disablePropertyName(), false);

    public String disablePropertyName() {
        return "zuul." + this.getClass().getSimpleName() + "." + this.filterType() + ".disable";
    }

    public boolean isFilterDisabled() {
        return this.filterDisabled.get();
    }
```
they already deal with disabling a filter...this can be used to turn off or turn on filter in a running application?

spencergibb:
@ArekCzarnik looks like it can, is that sufficient for your needs?

ArekCzarnik:
yes i already use it in my application.

ArekCzarnik:
@spencergibb: yes it is sufficient for my needs. We can drop the pull request, maybe it would be nice to add this into the documentation as an example for disable the filters. Thanks for your time...

---
zuul:
 SendResponseFilter:
   post:
     disable: true

Issue #377, opened 2015-06-05
The /pause endpoint should mark an instance as OUT_OF_SERVICE


heyjustin:
I believe the intent of the code in the _stop()_ method of _EurekaDiscoveryClientConfiguration_ is to mark an instance as OUT_OF_SERVICE:
```
@Override
public void stop() {
    log.info("Unregistering application " + this.instanceConfig.getAppname()
            + " with eureka with status OUT_OF_SERVICE");
    if (ApplicationInfoManager.getInstance().getInfo() != null) {
        ApplicationInfoManager.getInstance().setInstanceStatus(
                InstanceStatus.OUT_OF_SERVICE);
    }
    this.running.set(false);
}
```

This variant of the  _stop_ method is never executed since the _EurekaDiscoveryClientConfiguration_ implements the _SmartLifecyle_ interface, which states (referring to the _stop(Runnable)_ method):
> The LifecycleProcessor will call only this variant of the stop method; i.e. Lifecycle.stop() will not be called for SmartLifecycle implementations unless explicitly delegated to within this method.

Therefore, a solution to this problem is the delegation to the stop method as follows:
```
@Override
public void stop(Runnable callback) {
	stop();
	callback.run();
}        
```

However, having done that, and tested it locally, the instance is still not marked at OUT_OF_SERVICE in Eureka, even though I see the following message in the log files:
```
INFO 6672 --- [scoveryClient-1] com.netflix.discovery.DiscoveryClient    : DiscoveryClient_STATUS-APP/XXXXXX:status-app:PID-6672 - retransmit instance info with status OUT_OF_SERVICE
```

I will need to dig deeper to find out why not.

heyjustin:
Ok, I've dug deeper....

When we pause an instance (using the /pause endpoint) this is the flow of events:

**On the instance**
1. The _InstanceStatus_ is marked as OUT_OF_SERVICE (assuming the fix mentioned above is made to _EurekaDiscoveryClientConfiguration_) and flagged as dirty.
2. This is picked up by the _InstanceInfoReplicator_, which calls _register()_ on the _DiscoveryClient_ 
3. This make a remote call to the eureka server with _Action.Register_, i.e. POSTs to http://eureka/apps/{appName} with the payload containing the _InstanceInfo_

**On the eureka server**
1. The REST call is consumed by the method _com.netflix.eureka.resources.ApplicationResource#addInstance_
[...]

PR #390, opened 2015-06-10
Mark an instance as DOWN when the application context is stopped.

heyjustin:
Fixes gh-377


brenuart:
+1 for this PR.
An Eureka instance that don't want to receive traffic should set its status to DOWN.
OUT_OF_SERVICE should be used when traffic should be closed for administrative reasons or else (but the instance itself cannot report this status).

dsyer:
Looks useful. Can you fill in the contributor's agreement (link in README) please?

heyjustin:
Individual Contributor Agreement No: 124720150605045930

PR #356, opened 2015-05-19
upgrade eureka to take advantage of new DI refactoring

spencergibb:
things labeled "FIXME: eureka DI" are probably broken

fixes gh-349

PR #398, opened 2015-06-17
Upgrade eureka to 1.1.157

### LINKED ISSUE #349: upgrade eureka to 1.2.0
https://github.com/Netflix/eureka/releases/tag/v1.1.152
Will require changes because json is no longer rendered using xstream.

spencergibb:
Fixes #349 and #356. Unfortunately I can't separate the upgrade without getting rid of the static accessors.


brenuart:
Hi @spencergibb. I see you are working hard on this. Just want to provide you with some feedback on what we done regarding the initialisation of the Eureka DiscoveryClient.

First, we create a new instance ourselves via `new DiscoveryClient()`. This is now possible since some of their DI refactoring work. BTW, you may have noticed the `DiscoveryClient` takes care of registering itself in `DiscoveryManager.getInstance()` - so we don't care about the evil singleton anymore. We just have to make sure the Eureka client is instantiated before any other Netflix components are created.

Second, we actually subclassed the `DiscoveryClient` - this gave us access to their protected constructor where you can pass the `DiscoveryClientOptionalArgs` that may be required if you opt for the `EventBus` subscription (for CacheRefreshedEvents).

Third, we choose to override the `DiscoveryClient.onCacheRefreshed()` method to capture the event directly instead of having to instantiated yet another Netflix stuff... Once capture, the event is converted and directly rethrown into the Spring application context.

Fourth. The last issue was to support `server.port=0` For this, we concluded that the main problem was the DiscoveryClient attempts to register itself with the Eureka Server immediately after construction. So we managed to do such that the registration is delayed until we explicitly tell it to do so. This give us a chance to wait for the embedded servlet container. Until then, the client is usable in the sense that it fetches registry information - but hasn't yet started its registration and heartbeat work. Of course, this last point requires some awful reflection hacks...

Hope this feedback is of any help to you. Don't hesitate if you want us to share some code and contribute to your branch.

Cheers.

spencergibb:
@brenuart re point 2) the constructor is now public (one of my PRs).  3) is a good idea.  4) is going to be the hardest thing to do.

spencergibb:
@brenuart have you dealt with the eureka jackson codec?

brenuart:
Regarding 4), here is basically what we have done:

1. If `server.port=0` and 'shouldRegister=true' at the time we construct the DiscoveryClient, then we force `EurekaClientConfig.shouldRegister=false` temporarily to prevent any attempt to register.
2. We wait for the EmbeddedServletContainer and update the `InstanceInfo` accordingly.
3. Finally, we tell the DiscoveryClient to start the registration and heartbeat work as follows:

3.1. update the EurekaClientConfig instance (the one given to the client at construct time):

    clientConfig.setRegisterWithEureka(true)
    clientConfig.setFetchRegistry(false)

3.2. invoke the method `initScheduledTasks()` (by reflection). Note that is important to disable the registry fetch (last line above) before calling the method (see method's code).

3.3. re-enable registry fetch if it was enabled previously

brenuart:
@spencergibb About the jackson codec: NO, all our work is based on eureka 1.1.147
We had so many issues that we though we had better to spend our time contributing our changes/ideas to both Netflix and SpringCloud, rather than maintaing a code base full of hacks in all directions that are likely to break at the next version upgrade anyway. Hence my recent activity here...

brenuart:
Just to save you some time...

Since you created `CloudEurekaClient` via an `@Bean` method, the following piece of code:

    public class EurekaDiscoveryClientConfiguration {
        @PreDestroy
        public void close() {...}

is not required anymore and will lead to duplicate shutdown (the second with errors) because of:

    public class DiscoveryClient {
       @PreDestroy
       public void shutdown() {...}

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/config/DiscoveryClientConfigServiceAutoConfiguration.java
```diff
@@ -53,17 +52,18 @@
...
-        if (DiscoveryManager.getInstance().getDiscoveryClient() != null) {
-            DiscoveryManager.getInstance().getDiscoveryClient().shutdown();
-        }
-		ApplicationInfoManager.getInstance().initComponent(this.instanceConfig);
-		DiscoveryManager.getInstance().initComponent(this.instanceConfig,
-				this.clientConfig);
+		eurekaClient.shutdown();;
+		// FIXME: reinit EurekaClient and ApplicationInfoManager
+		//applicationInfoManager.initComponent(this.instanceConfig);
+		//discoveryManager.initComponent(this.instanceConfig, this.clientConfig);
```
spencergibb: @dsyer not sure what I need to do here and what case this is fixing to test it.  Now all the objects held in `ApplicationInfoManager` and `DiscoveryManager` are now beans.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/DataCenterAwareJacksonCodec.java
```diff
@@ -0,0 +1,401 @@
...
+import com.netflix.appinfo.InstanceInfo;
+import com.netflix.appinfo.LeaseInfo;
+import com.netflix.discovery.converters.Auto;
+import com.netflix.discovery.converters.EurekaJacksonCodec;
+import com.netflix.discovery.converters.StringCache;
+
+/**
+ * @author Spencer Gibb
+ */
+public class DataCenterAwareJacksonCodec extends EurekaJacksonCodec {
```
spencergibb: The `EurekaJacksonCodec` is awful.  I'll try and submit a PR to help.

spencergibb: See https://github.com/Netflix/eureka/pull/555

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/config/DiscoveryClientConfigServiceAutoConfiguration.java
spencergibb: @dsyer what do I need to do here?

spencergibb:
This is pretty close

brenuart:
Cool - good work. And as far as I can see, code is getting more clear ;-)

spencergibb:
Cherry picked commits into #477

tunix:
Hi @spencergibb ,

Is `HeartbeatEvent` kept for backward compatibility? I created a subclass of `DiscoveryClient` (to override its `onCacheRefreshed` method to refresh only when the cache is really changed) but had issues as `EurekaRegistration` class requires an instance of `CloudEurekaClient` class. It worked when I subclasses `CloudEurekaClient` instead but I'm afraid of breaking something else within Spring Cloud.

The changes I made causes `HeartbeatEvent` to be published only when there is a new application registered. Is that a problem for parts that I'm unaware of?

spencergibb:
@tunix In the future please ask a question in a new issue (though I see you already asked on gitter, so no need to do that now) rather than commenting on a very old PR

Issue #407, opened 2015-06-19
Zuul - SimpleHostRoutingFilter forwards inbound request paths as unescaped


brandon-atkinson:
This issue exhibits itself much like #347, but the fix in bd5d00c addressed it only in RibbonRoutingFilter. 

Example of issue:
inbound request path: /thirst%20mutilator
outbound request path: /thirst mutilator

The cause is similar, the routing filter will use the "requestURI" that the PreDecoratingFilter (or really any other filter) puts into the RequestContext. PreDecoratingFilter happens to put an unescaped path into requestURI.

This problem really exists because the PreDecoratingFilter unescapes the uri before putting it into the context. Is having an unescaped path in context.requestURI a requirement? 

I have forked locally to fix this in the SimpleHostRoutingFilter, definitely wouldn't mind submitting a PR.

brandon-atkinson:
The reason this was not caught is that there was a single unit test which only ran requests through the ribbon path.

dsyer:
PR more than welcome thanks.

PR #408, opened 2015-06-19
Fix for #407 Zuul unescaped paths fixed for both Ribbon and SimpleHost routing

brandon-atkinson:
This pull request pulls adds a test to demonstrate the fact that the simple host routing was broken, pulls URI generation into the ProxyRequestHelper, and updates both the SimpleHost and Ribbon routing filters to use the helper.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommand.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommand.java
@@ -80,7 +80,7 @@ public class RibbonCommand extends HystrixCommand<HttpResponse> {
 		super(getSetter(commandKey));
 		this.restClient = restClient;
 		this.verb = verb;
-		this.uri = (StringUtils.hasText(uri))? UriComponentsBuilder.fromUriString(uri).build().toUri() : new URI(uri);
+		this.uri = new URI(uri);
 		this.retryable = retryable;
 		this.headers = headers;
 		this.params = params;
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -168,10 +168,7 @@ public class SimpleHostRoutingFilter extends ZuulFilter {
 		InputStream requestEntity = getRequestBody(request);
 		HttpClient httpclient = CLIENT.get();
 
-		String uri = request.getRequestURI();
-		if (context.get("requestURI") != null) {
-			uri = (String) context.get("requestURI");
-		}
+		String uri = this.helper.buildZuulRequestURI(request);
 
 		try {
 			HttpResponse response = forward(httpclient, verb, uri, request, headers,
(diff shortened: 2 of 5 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommand.java
```diff
@@ -80,7 +80,7 @@ public RibbonCommand(String commandKey, RestClient restClient, Verb verb, String
 		super(getSetter(commandKey));
 		this.restClient = restClient;
 		this.verb = verb;
-		this.uri = (StringUtils.hasText(uri))? UriComponentsBuilder.fromUriString(uri).build().toUri() : new URI(uri);
```
spencergibb: I still think the null check needs to happen here.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -16,34 +16,12 @@
...
-import java.util.concurrent.atomic.AtomicReference;
-
-import javax.annotation.PreDestroy;
-import javax.net.ssl.SSLContext;
-import javax.net.ssl.TrustManager;
-import javax.net.ssl.X509TrustManager;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
+import com.netflix.config.DynamicIntProperty;
```
spencergibb: Please update your rules for imports to match ours so no needless changes are made in imports.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommand.java
brandon-atkinson: The only thing I did here was remove the escaping logic to allow for it to occur outside the ribbon command. The logic in the original would still throw NPE if the RibbonCommand was constructed with a null URI string. 

Does it make more sense to do a null check or make a non-null uri a design constraint with an assertion?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
brandon-atkinson: Sorry, I'm not a fan of intellij's pushiness when it comes to import management. Are there any docs that describe the formatting used or do people just do it by trial and error?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommand.java
spencergibb: My bad, I was thinking the exceptional case was with `URI` if `uri` was empty, but it was with `UriComponentsBuilder`.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
spencergibb: We reccomend the [Eclipse Code Formatter Plugin])https://github.com/spring-cloud/spring-cloud-netflix#code-conventions-and-housekeeping) in the docs.  I use this file for imports in that plugin: `https://github.com/spring-cloud/spring-cloud-netflix/blob/master/eclipse/org.eclipse.jdt.ui.prefs`. Generally, if I see the imports are very different, I'll revert those back.

spencergibb:
@brandon-atkinson could you squash your commits?

spencergibb:
Closed via 283c0f7600ffe3bbc41a581534f675f1c829aa33

PR #409, opened 2015-06-19
Zuul: Fix #394 by removing deflate accept-encoding from outbound requests

### LINKED ISSUE #394: Content-Encoding: deflate is handled incorrectly
A header `Accept-Encoding: deflate, gzip` is appended to every origin request here:

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java#L96

The origin server returns a `Content-Encoding: deflate`, but this header is removed from response here:

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java#L158

There is  code to either decode `gzip` responses or add a `Content-Encoding: gzip` based on whether or not a client supports gzip:

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java#L115-L129

However all other encodings are left as-is and no `Content-Encoding` header is sent to the client.

In summary:
* Any response that has `Content-Encoding` other than `gzip` will not be forwarded to the client correctly because that header is removed while encoding is preserved
* Spring Cloud specifically requests `Content-Encoding: deflate` which it then cannot handle correctly

This also makes me wonder: why does Spring need to get involved with content encodings at all? Could it preserve the response body and headers as-is and leave content negotiation to the client and origin server?

brandon-atkinson:
The problem with the current code is that the proxy is taking control of accept-encoding to compress outbound requests and advertising an encoding it doesn't itself support correctly. 

One possible solution would be to support deflate in the SendResponseFilter. 

Another alternative would be to allow Accept-Encoding to pass through from the client to the origin-server. In order to do that, the proxy would have to treat the payload as opaque, which it does not currently do. Also, having the proxy allow gzip even for clients that don't advertise support for it can make proxying a bit faster.

The path I chose was to remove the advertisement of 'deflate' as it required the least change.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
@@ -93,7 +93,7 @@ public class ProxyRequestHelper {
 		for (String header : zuulRequestHeaders.keySet()) {
 			headers.set(header, zuulRequestHeaders.get(header));
 		}
-		headers.set("accept-encoding", "deflate, gzip");
+		headers.set("accept-encoding", "gzip");
 		return headers;
 	}
 
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelperTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelperTests.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.netflix.zuul.filters;
 
-import java.util.List;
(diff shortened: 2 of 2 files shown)


dsyer:
Thanks. Did you fill out the contributor's agreement (link in README)?

brandon-atkinson:
What project and project lead applies?

spencergibb:
@brandon-atkinson "Spring Cloud" project and myself (Spencer Gibb) or David Syer as project lead.

spencergibb:
Closed via 80206e6b3861015438c0dec43ac8af100a10c1a2

Issue #315, opened 2015-04-23
ZonePreferenceServerListFilter Does Not Use Domain As Zone When approximateZoneFromHostname Is Set To True


ryanjbaxter:
I have set `approximateZoneFromHostname` to true in my server to try and force my Ribbon clients to always use services in the same zone as the server.  However once setting that value my server still seemed to be going round robin between the various instances of the service that was registered with Eureka.  When I looked at `ZonePreferenceServerListFilter` it looks like `this.zone` is always set to `defaultZone` and not to the domain of the server as it should be when `approximateZoneFromHostname` is set to true.  The problem appears to be in the `initWithNiwsConfig` method which I believe should be setting `this.zone` to the domain of the server.

Since `this.zone` is set to `defaultZone`, `getFilteredListOfServers` never finds a server in the same zone.

ryanjbaxter:
Actually after thinking about this more the code in `getFilteredListOfServers` is OK, we should be setting the zone correctly somewhere else.  But not sure where that should be yet....

ryanjbaxter:
So it looks like the zone is being set in the deployment context in `EurekaRibbonClientConfiguration.preprocess`.  The problem I am facing now is how to get the host name of the service.  I cannot find a good way to get the host name of the service using Ribbon.  It seems like Eureka would be the right place to get the host name from, but then the question is how do I get the host name for the current service instance.  Still poking around...

PR #423, opened 2015-07-08
Fix for #315. Sets zone based on approximateZoneFromeHostname

ryanjbaxter:
See #315


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/eureka/ZoneUtils.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class ZoneUtils {
+
+	/**
+	 * Approximates Eureka zones from a host name. This method approximates the zone to be
+	 * everything after the first "." in the host name.
+	 * @param host The host name to extract the host name from
+	 * @return The approximate zone
+	 */
+	public static String extractApproximateZone(String host) {
```
spencergibb: How about an interface and implementation here?  I can imagine there might be other ways to approximate zone.  What do you think?

ryanjbaxter: Other ways to approximate the zone from the host name?  Right now there is only a property to enable this functionality based off the host name.  If you think there might be more than one way of doing that than I think an interface might make sense.

dsyer: I'm happy to merge this as a bug fix for now. Please send the interface idea as a separate pull request if you have time.

Issue #426, opened 2015-07-09
Cannot modify query string for static zuul routes


qbast:
If route is using RibbonRoutingFilter, it is easy to modify/add query string parameters to the request by accessing ctx.getRequestQueryParams() and making modifications. This parameters are later passed to RibbonCommand which uses them to construct query string.

However, if route is static (defined with url instead of service name), it uses SimpleHostRoutingFilter. It also retrieves parameters from ctx.getRequestQueryParams()  and passes them to forward() method, but then they are ignored and query string is taken directly from original request.

My use case is adding tenant id as query string parameter based on original host name (configuration file has hostname to tenant and keycloak realm mapping).

PR #434, opened 2015-07-11
Build query string from context instead of original request

qbast:
Recreate query string from either getCurrentContext().getRequestQueryParams()
or original request as fallback. This allows to override query string
params in pre filters and matches behaviour of RibbonRoutingFilter

gh-426

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -18,8 +18,10 @@ package org.springframework.cloud.netflix.zuul.filters.route;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.net.Socket;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.net.UnknownHostException;
 import java.security.KeyManagementException;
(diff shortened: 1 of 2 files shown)


spencergibb:
Two questions, how about a unit test to verify your fix?  Also, have you signed the [contributor agreement](https://support.springsource.com/spring_committer_signup)?

qbast:
1) I am unsure how to go about it - SimpleHostRoutingFilter is quite opaque and I don't see a way to mock HttpClient, override forwardRequest() or intercept constructed request in another way.
2) Yes.

spencergibb:
Don't mock make a real request ala https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/SampleZuulProxyApplicationTests.java

qbast:
I added two tests - one showing that overriding query string works and another to verify that it does not break normal case.

spencergibb:
@qbast very nice, thanks!

spencergibb:
Merged via 22ed8818053d3915200e422c6d56801af8a861d6

Xero314:
I know this is a closed issue, but I wanted to bring to everyone's attention that this implementation is going to be problematic because URLEncoder does not properly encode query parameters as it encodes "/" and "?" which are not necessary to be encoded in query parameter values and will cause some applications to fail as they do not expected an encoded value here.

spencergibb:
@Xero314 you're right it is closed. Please open a new issue if you think there is an issue.

PR #455, opened 2015-07-25
Specifying only the serviceId causes NPE

jmnarloch:
If you specify only the serviceId on @FeignClient annotation you will end with NPE.

I had aggreed to Contributor License Agreement.


spencergibb:
Closed via 42f6e4d1f0eb71a7f1ba215edadef82258602e60, I cleaned up the imports and added you as an author.

jmnarloch:
Cool, thanks!

PR #464, opened 2015-07-27
Ignoring patterns in zuul

### LINKED ISSUE #445: Looking for a way to block patterns in zuul
Currently it is possible to ignore an entire service in the zuulProxy:

```yaml
 zuul:
  ignoredServices: '*'
  routes:
    users: /myusers/**
```

I would like to go one step further and have a property `zuul.ignoredRoutes` or `zuul.ignoredPatterns` which could hold an ant pattern that when matched, doesn't forward the request.

The specific use case is where I am trying to allow actuator endpoints to be reachable directly on the microservice, but not through the gateway (to not expose them to the outside world).
While the actuator is the main reason, there would also be other resources to hide to the outside.

andreasevers:
…terns to ignore. These patterns are being evaluated at the end of the route location process, which means prefixes should be included in the pattern to warrant a match. Ignored patterns span all services and supersede any other route specification.

```yaml
 zuul:
  ignoredPatterns: */admin/**
  routes:
    users: /myusers/**
````

This means that all calls such as "/myusers/101" will be forwarded to "/101" on the "users" service. But calls including "/admin/" will not resolve.

closes #445

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -900,6 +900,23 @@ An application with the `@EnableZuulProxy` could act as a standalone
 server if you set a default route ("/"), for example `zuul.route.home:
 /` would route all traffic (i.e. "/**") to the "home" service.
 
+If more fine-grained ignoring is needed, you can specify specific patterns to ignore.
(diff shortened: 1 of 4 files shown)


spencergibb:
Closed via cb732763b5f6ddf4e9fa877cd0c531adae9d4a5a

PR #477, opened 2015-08-05
Upgrade to Eureka 1.2.0

spencergibb:
Implements forwarding of eureka CacheRefreshedEvent as HeartbeatEvent,
rather thank serialization hack.  Uses new Eureka DI functionality
(mostly there, some issues). Updates EurekaJacksonCodec to be
"DataCenterAware". Reference EurekaClient interface rather than
DiscoveryClient class.

fixes gh-349


pom.xml
```diff
@@ -188,6 +188,11 @@
 				</exclusions>
 			</dependency>
 			<dependency>
+				<groupId>com.netflix.governator</groupId>
```
spencergibb: https://github.com/Netflix/eureka/pull/606 will hopefully remove this dependency.

PR #478, opened 2015-08-07
Make EurekaClient available before smart lifecycle phase 0

jkschneider:
I would like to improve the usefulness of the EurekaClient as a utility I can inject around my application.  In one particular case using Titan (Graph DB) Cassandra, I found myself configuring the TitanGraph instance in a configuration class that was a SmartLifecycle phase 1 and quickly realized that anywhere else in my application where the graph was used, these beans too would have to be SmartLifecycles...

By moving the relevant beans from `EurekaDiscoveryClientConfiguration` to `EurekaClientAutoConfiguration` I believe we can alleviate this.

We should be able to strike this part from the docs:

> Don’t use the DiscoveryClient in @PostConstruct method or in a @Scheduled method (or anywhere where the ApplicationContext might not be started yet). It is initialized in a SmartLifecycle (with phase=0) so the earliest you can rely on it being available is in another SmartLifecycle with higher phase.

If it hasn't already been done, I suppose we should update the docs to refer to `EurekaClient` instead of `DiscoveryClient` as the injectable resource.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -21,6 +21,11 @@ import java.util.concurrent.ConcurrentMap;
 
 import javax.annotation.PostConstruct;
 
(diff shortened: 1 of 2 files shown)


spencergibb:
Polished (imports), squashed and merged via b9d1d5e6973c087ad8f3fe42d88ea883296f917a
Thanks @jkschneider!

Issue #337, opened 2015-05-06
LoadBalancerClient is not picking up secure url


beku8:
I've following eureka.instance config: 
```
eureka:
  instance:
    appname: ${spring.application.name}
    hostname: ${vcap.application.uris[0]}
    homePageUrl: https://${vcap.application.uris[0]}
    instanceId: ${vcap.application.instance_id}
    securePortEnabled: true
    nonSecurePortEnabled: false
```
which is deployed to PWS CloudFoundry, ```/configprops``` shows:
```
eurekaInstanceConfigBean: {
prefix: "eureka.instance",
properties: {
appGroupName: null,
nonSecurePortEnabled: false,
statusPageUrl: null,
initialStatus: "UP",
instanceEnabledOnit: false,
homePageUrl: "https://login-nd.cfapps.io",
securePort: 443,
securePortEnabled: true,
hostname: "login-nd.cfapps.io",
dataCenterInfo: {
name: "MyOwn"
},
appname: "login",
healthCheckUrlPath: "/health",
leaseExpirationDurationInSeconds: 90,
metadataMap: { },
preferIpAddress: false,
healthCheckUrl: null,
statusPageUrlPath: "/info",
virtualHostName: "login",
ipAddress: "10.254.0.26",
leaseRenewalIntervalInSeconds: 10,
homePageUrlPath: "/",
nonSecurePort: 80,
namespace: "eureka",
secureVirtualHostName: null,
secureHealthCheckUrl: null
}
}
```
which results in the eureka registry(```/eureka/apps```):
```
<application>
<name>LOGIN</name>
<instance>
<hostName>login-nd.cfapps.io</hostName>
<app>LOGIN</app>
<ipAddr>10.254.0.26</ipAddr>
<status>UP</status>
<overriddenstatus>UNKNOWN</overriddenstatus>
<port enabled="false">80</port>
<securePort enabled="true">443</securePort>
<countryId>1</countryId>
<dataCenterInfo class="com.netflix.appinfo.InstanceInfo$DefaultDataCenterInfo">
<name>MyOwn</name>
</dataCenterInfo>
<leaseInfo>
<renewalIntervalInSecs>10</renewalIntervalInSecs>
<durationInSecs>90</durationInSecs>
<registrationTimestamp>1430883647411</registrationTimestamp>
<lastRenewalTimestamp>1430883747423</lastRenewalTimestamp>
<evictionTimestamp>0</evictionTimestamp>
[...]

PR #486, opened 2015-08-12
Detect https from Ribbon Client Config

steve-oakey:
- RibbonLoadBalancer and RibbonLoadBalancer use IClientConfig IsSecure
property for service ID to determine if the service is secure. Replaces
scheme of URI of request with https if the request is secure.

Solves gh-337


spencergibb:
Closed and polished via d24bff02f3e3028924884ed408aaa1159d027051

Issue #489, opened 2015-08-17
Feign Content-Encoding gzip, deflate not supported


jmnarloch:
Enabling Accept-Encoding: gzip, deflate through custom Feign RequestInterceptor on the outgoing requests ends with exception when the response is compressed (or in other word the payload is content encoded):

```
Could not read document: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]; nested exception is com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]
feign.codec.DecodeException: Could not read document: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]; nested exception is com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]
	at feign.SynchronousMethodHandler.decode(SynchronousMethodHandler.java:150)
	at feign.SynchronousMethodHandler.executeAndDecode(SynchronousMethodHandler.java:118)
	at feign.SynchronousMethodHandler.invoke(SynchronousMethodHandler.java:71)
	at feign.ReflectiveFeign$FeignInvocationHandler.invoke(ReflectiveFeign.java:94)
Caused by: org.springframework.http.converter.HttpMessageNotReadableException: Could not read document: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]; nested exception is com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: java.io.PushbackInputStream@7169c7d4; line: 1, column: 2]
[...]

PR #490, opened 2015-08-18
Feign GZIP compression

jmnarloch:
After strugling with issue: https://github.com/spring-cloud/spring-cloud-netflix/issues/489#issuecomment-132094542 in version 1.0.3.RELEASE

I would like to propose this simple test case that verifies whether content-encoded responses succeeds, though it's not so simple to test that since the ApacheHttpClient will strip the Content-Encoding header :(

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/FeignContentGzipEncodingInterceptor.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/FeignContentGzipEncodingInterceptor.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2013-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.feign.encoding;
+
+import feign.RequestTemplate;
+
+import java.util.Collection;
+import java.util.Map;
+
+/**
+ * Enables the HTTP request payload compression by specifying the {@code Content-Encoding} headers.
+ *
+ * @author Jakub Narloch
+ */
+public class FeignContentGzipEncodingInterceptor extends BaseRequestInterceptor {
+
+    /**
+     * Creates new instance of {@link FeignContentGzipEncodingInterceptor}.
+     *
+     * @param properties the encoding properties
+     */
+    protected FeignContentGzipEncodingInterceptor(FeignClientEncodingProperties properties) {
+        super(properties);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void apply(RequestTemplate template) {
+
+        if (requiresCompression(template)) {
+            addHeader(template, HttpEncoding.CONTENT_ENCODING_HEADER, HttpEncoding.GZIP_ENCODING,
+                    HttpEncoding.DEFLATE_ENCODING);
+        }
+    }
+
+    /**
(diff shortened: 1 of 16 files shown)


adriancole:
looks like how to setup things, is that the purpose (eg to make sure the setup doesn't break)? I'm asking as I don't see any assertion anywhere that shows the response is indeed encoded.

jmnarloch:
This is problematic to do, because the Apache HttpClient will remove the Content-Encoding header, although when you run this tests against the version 1.0.1 you will get an exception during decoding.

So this is kind of binary test - either it passes without any error or it fails.

jmnarloch:
By the way I had build on top of that this little thingy: https://github.com/jmnarloch/feign-encoding-spring-cloud-starter, I could turn that into pull request, but I do not know whether you want it or not.

adriancole:
specifying gzip encoding seems like something desirable. @spencergibb wdyt about the impl? does  https://github.com/jmnarloch/feign-encoding-spring-cloud-starter match spring cloud/boot world well enough to convert to a PR?

spencergibb:
seems like it would be generally useful

adriancole:
iotw @jmnarloch go for it :) I suspect you could repurpose this PR or drop an open another.. up to you!

jmnarloch:
OK, it's done.

This triggers in fact the Feign build in support  by specifing the proper headers. I made an attempt to spring-ify this commit adjusting to your conventions, but if you prefer to rename some classes, or do any other changes this is fine with me.

I've myself started to think whether not to rename the annotations into something user friendly like @EnableRequestCompression / @EnableResponseCompression etc. but for now I've left the existing ones.

I've have any comments just shoot.

cc @adriancole @spencergibb

spencergibb:
@jmnarloch thanks! I'll take a look next week after I'm home from spring one.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/EnableFeignAcceptGzipEncoding.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Jakub Narloch
+ * @see FeignAcceptGzipEncodingInterceptor
+ * @see FeignAcceptGzipEncodingConfiguration
+ */
+@Documented
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Import(FeignAcceptGzipEncodingConfiguration.class)
+public @interface EnableFeignAcceptGzipEncoding {
```
spencergibb: Can we go with an auto-configuration enabled via properties, rather than an @Enable? Something similar to what spring-boot [does for tomcat](http://docs.spring.io/spring-boot/docs/current/reference/html/howto-embedded-servlet-containers.html#how-to-enable-http-response-compression).

spencergibb:
Looking good. I like what it does. See my comment about how to enable.

jmnarloch:
Sure, so since we have to distinct part the request/response compression controlled by two different HTTP headers my proposition is to introduce fallowing properties:

```
feign.compression.response.enabled=true # will be disabled by default

feign.compression.request.enabled=true # false by default
feign.compression.request.mime-types=text/xml,application/xml,application/json
feign.compression.request.min-request-size=2048
```
I've fallowed the naming convention from Spring Boot 1.3.0

What do you think?

spencergibb:
Closed via 0c2c564be54ba671338ef7fb4ff5138d229bc104.
@jmnarloch thanks! how about a followup PR that adds this to the docs?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/FeignContentGzipEncodingInterceptor.java
```diff
@@ -0,0 +1,111 @@
...
+        if (contentTypes == null || contentTypes.size() == 0) {
+            return false;
+        }
+
+        if (getProperties().getMimeTypes() == null || getProperties().getMimeTypes().length == 0) {
+            // no specific mime types has been set - matching everything
+            return true;
+        }
+
+        for (String mimeType : getProperties().getMimeTypes()) {
```
TrilokDevda: error in getProperties()

jmnarloch: What do you mean?

PR #493, opened 2015-08-19
Consolidates references to parent pom

adriancole:
Some projects were failing on `./mvnw install` due to a non-existent
local path. This makes all poms use the same approach.

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -6,7 +6,7 @@
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-netflix</artifactId>
 		<version>1.1.0.BUILD-SNAPSHOT</version>
-		<relativePath>..</relativePath>
+		<relativePath>..</relativePath> <!-- lookup parent from repository -->
 	</parent>
 	<artifactId>spring-cloud-netflix-core</artifactId>
 	<packaging>jar</packaging>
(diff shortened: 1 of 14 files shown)


adriancole:
Note that without this commit, the following works `./mvnw -s .settings.xml install`.

spring-cloud-netflix-core/pom.xml
```diff
@@ -6,7 +6,7 @@
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-netflix</artifactId>
 		<version>1.1.0.BUILD-SNAPSHOT</version>
-		<relativePath>..</relativePath>
+                <relativePath>..</relativePath> <!-- lookup parent from repository -->
```
spencergibb: tabs vs spaces

docs/pom.xml
```diff
@@ -6,7 +6,7 @@
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
 		<version>1.1.0.BUILD-SNAPSHOT</version>
-		<relativePath />
+		<relativePath>..</relativePath> <!-- lookup parent from repository -->
```
spencergibb: This one is pointing at `spring-cloud-build` not `spring-cloud-netflix` so probably want to keep this one.

spring-cloud-netflix-core/pom.xml
adriancole: hurt me to use a tab.. suspecting now the ones with tabs are the wrong ones? :) Anyway I'll change to spaces with much delight!

spencergibb: No, it should be tabs :-(

adriancole:
fixed the tabs!

PR #509, opened 2015-08-29
Eureka health check

jmnarloch:
This change adds a HealthCheckHandler that plugs in into the Boot Actuator's HealthIndicator system and propagates the instance health status to Eureka.

Fortunetly the well defined Actuator statuses maps 1:1 to Eureka's ones, though any custom status will be mapped to Unknown.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
@@ -16,15 +16,18 @@
 
 package org.springframework.cloud.netflix.eureka;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
-
+import com.netflix.appinfo.ApplicationInfoManager;
+import com.netflix.appinfo.EurekaInstanceConfig;
+import com.netflix.appinfo.HealthCheckHandler;
+import com.netflix.appinfo.InstanceInfo.InstanceStatus;
+import com.netflix.discovery.EurekaClient;
+import com.netflix.discovery.EurekaClientConfig;
 import lombok.extern.apachecommons.CommonsLog;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.actuate.endpoint.Endpoint;
+import org.springframework.boot.actuate.health.HealthAggregator;
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.OrderedHealthAggregator;
 import org.springframework.boot.actuate.metrics.reader.CompositeMetricReader;
 import org.springframework.boot.actuate.metrics.reader.MetricReader;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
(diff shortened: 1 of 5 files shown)


spencergibb:
Your blog: http://blog.jmnarloch.io/2015/09/02/spring-cloud-fixing-eureka-application-status

spencergibb:
I'm interested in @dsyer's opinion on this. At the very least, you should be able to opt out of this (an enabled=false with matchIfMissing=true property). I'm wondering if it should be opt in (matchIfMissing=false).

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
```diff
@@ -174,4 +175,17 @@ public EurekaHealthIndicator eurekaHealthIndicator(EurekaClient eurekaClient,
...
 
+	@Configuration
+	@ConditionalOnBean(HealthIndicator.class)
+	protected static class EurekaHealthCheckHandlerConfiguration {
+
+		@Autowired(required = false)
+		private HealthAggregator healthAggregator = new OrderedHealthAggregator();
+
+		@Bean
+		@ConditionalOnMissingBean
```
spencergibb: `@ConditionalOnMissingBean(HealthCheckHandler.class)` so user a supplied `HealthCheckHandler` doesn't get clobbered.

jmnarloch:
I will add the property condition. 

I had figure this out after playing with Consul, since Consul performs the health check itself, the Actuator statuses will be reflected in Consul, so it makes sense to reflect that in Eureka.

spencergibb:
Yeah, I finally wrapped my head around this when I realized it was the push version of the consul health check.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
```diff
@@ -174,4 +175,17 @@ public EurekaHealthIndicator eurekaHealthIndicator(EurekaClient eurekaClient,
...
 
+	@Configuration
+	@ConditionalOnBean(HealthIndicator.class)
+	protected static class EurekaHealthCheckHandlerConfiguration {
+
+		@Autowired(required = false)
+		private HealthAggregator healthAggregator = new OrderedHealthAggregator();
+
+		@Bean
+		@ConditionalOnMissingBean(EurekaHealthCheckHandler.class)
```
spencergibb: This needs to be `HealthCheckHandler.class`.

jmnarloch:
I added the property condition - it's disabled by default and some extra documentation.

Hmm, I had weird problem now - some test are failling due to missing constructor despite the lombok annotations being specified. Although I haven't change anything here.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
jmnarloch: Yeah I noticed that, thanks.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
```diff
@@ -174,4 +175,18 @@ public EurekaHealthIndicator eurekaHealthIndicator(EurekaClient eurekaClient,
 		}
 	}
 
+	@Configuration
+	@ConditionalOnProperty(value = "eureka.client.healthcheck", matchIfMissing = false)
```
spencergibb: Can you make it `eureka.client.healthcheck.enabled`? That matches our convention a bit better.

jmnarloch: Alright

jmnarloch: Patch on the way.

spencergibb:
Polished and merged via 304e836980adca29f20f76a6b95915527a847302

jmnarloch:
Cool, thanks.

Issue #441, opened 2015-07-15
@EnableFeignClients - NullPointerException encoder in FeignClientFactoryBean


steve-oakey:
I am using release train Angel.SR3. I am having the same issue as raised in gh-212. When `FeignClientFactoryBean` tries to build the Feign clients, `encoder` as well as all other autowired fields are null.

Basically I have:

* Main class annotated with `@SpringCloudApplication` and `@EnableFeignClients`
* An interface `UserClient` annotated with `@FeignClient`
* A class 'CompositeService` annotated @Service and autowiring `UserClient` (initializing this bean is where the failure occurs).

Looking at the DEBUG logs, the beans from `FeignAutoConfiguration` are not created until after the initialization of the above bean fails. I have tried putting `@EnableFeignClients` on the main class, a separate `@Configuration` class and a static inner class of another `@Configuration` class, all with the same result.

I have successfully used `@EnableFeignClients` in other sample projects I have made, so I am sure that the custom configuration I am providing is likely causing some timing issue. I am providing a number of custom auto-configuration classes that I am providing myself from an external library + 1 Bootstrap configuration.

I cannot produce a sample to reproduce because I am on a closed network and I am really having a hard time tracking down where the configuration problem may be in order to provide a reproducible sample. If someone could provide some ideas on what I should look for when debugging that would be really helpful. I can provide as much information as needed from logs and stack traces.

Thanks in advance.

Koizumi85:
Hello,

are you using a milestone of spring-boot 1.3.0? If so you will have to go down to 1.2.5.RELEASE, because spring-cloud does not support spring-boot 1.3 at the moment. See https://github.com/spring-cloud/spring-cloud-netflix/issues/326 for morge informations.

Hope it helped

steve-oakey:
No, I am using Spring Cloud Release Train Angel.SR3 which is using Spring Boot 1.2.4.

brenuart:
`FeignClientFactoryBean` throws a `NullPointerException` because its autowired dependencies have not been set at the time the factory method is invoked.

After some investigations it appears the `GenericBeanDefinition` created by `FeignClientsRegistrar` doesn't say anything about its autowire mode leaving it to the default  `AbstractBeanDefinition.AUTOWIRE_NO` which doesn't make it eligible for autowiring.

A possible solution is to slightly change `FeignClientsRegistrar.createBeanDefinition()` and add the following to the created bean definition:

[...]

PR #513, opened 2015-09-04
Change from @Autowired to context.getBean

spencergibb:
Solves problems where FeignClientFactoryBean is run before auto-configuration.

fixed gh-441

@dsyer @adriancole

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
@@ -16,14 +16,17 @@
 
 package org.springframework.cloud.netflix.feign;
 
-import java.util.List;
+import java.util.Map;
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
+import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.FactoryBean;
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
```diff
@@ -91,40 +65,67 @@ public void afterPropertiesSet() throws Exception {
...
+	}
+
 	protected <T> T loadBalance(Feign.Builder builder, Class<T> type, String schemeName) {
 		builder.logger(new Slf4jLogger(type)); // TODO: how to have choice here?
-		if (this.ribbonClient != null) {
-			return builder.client(this.ribbonClient).target(type, schemeName);
-		}
-		else {
-			return builder.target(LoadBalancingTarget.create(type, schemeName));
+		Client client = getOptional(Client.class);
```
spencergibb: This fixes the two projects I have that replicated the problem. I'll add the simpler one to spring-cloud-samples/tests

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
```diff
@@ -91,40 +65,67 @@ public void afterPropertiesSet() throws Exception {
...
+		if (client != null) {
+			return builder.client(client).target(type, schemeName);
 		}
+
+		throw new IllegalStateException("No Feign Client for loadBalancing defined. Did you forget to include spring-cloud-starter-ribbon?");
```
spencergibb: `builder.target(LoadBalancingTarget.create(type, schemeName))` would happen even if ribbon was not loaded, this is a problem too.

PR #515, opened 2015-09-07
Add X-Forwarded-Proto HTTP header to zuul proxy requests

### LINKED ISSUE #514: X-Forwarded-Proto is missing in zuul client requests
When using `@EnableZuulProxy` inside a spring boot application which itself is secured via SSL to act as a reverse proxy for other spring boot applications which are not secured, the client HTTP requests from zuul are missing the `X-Forwarded-Proto: https` HTTP header.

The HTTP headers `X-Forwarded-Host` and `X-Forwarded-Prefix` are present and their content is correct.

MichaelF25:
If the zuul reverse proxy and the destination services are using a different scheme
(for example HTTPS on proxy side and HTTP on service side) an additional HTTP
header is required according to RFC7239 [http://tools.ietf.org/html/rfc7239]

This PR fixes #514


spencergibb:
closed via bea48c515099f3b78fa9eb748016bbbecdacf0cb

PR #517, opened 2015-09-08
Added support for Spectator and Atlas, deprecated Servo

jkschneider:
There is a lot going on here, so some explanation is in order.  Expect a few more commits (docs, starter, etc) after you have a chance to provide feedback.

**Purpose** -  Servo is internally considered a legacy library at Netflix now, though not exactly deprecated just yet.  Current and future work should be directed at Spectator instead.  I also wish to add support for the Atlas metrics backend.  Atlas provides a unique stack language for generating real-time graphics from your metrics.  Since Atlas is a tag-based rather than hierarchical metrics backend, it provides a unique capability to Spring Cloud developers.

**Assorted Details**
1.  Spectator is compiled against Java 8.  This leaves me a little unclear as to whether or not we should fully remove the Servo behavior.  This is why you also see the bulk of the code in its own module.  The approach is intended to mirror the one you took with turbine-amqp.
2.  The current Servo behavior appears to be quite dangerous.  While I haven't spent the time digging up the exact root cause, `DefaultServoMetricNaming`'s prepending of "servo." on the front of metric names (that may have already been prepended with "servo.") seems to be periodically leading to a combinatorial explosion of metric names and a subsequent OOM.  For this reason and because we don't want to have two auto-configurations adding monitors to the Servo monitor registry, this PR puts `ServoMetricsAutoConfiguration` behind an explicit `@EnableServo` annotation.
3.  The Atlas integration provides a `Exporter` implementation so that the developer can fully control when metrics are pushed to the backend.
4.  `SpectatorMetricServices` is written against the new `CounterService`/`GaugeService` API rather than `MetricWriter`.  I'm only half sure this is the right idea.
5.  `SpectatorMetricReader` attempts to map the tag-based Spectator/Servo metrics onto a naming scheme that looks decent in the /metrics endpoint.
6.  Once Spring Boot is on Java 8, we can add a default method to `ClientHttpRequestInterceptor` to fetch the parameterized URI and do away with the AspectJ silliness you find in `RestTemplateUrlTemplateCapturingAspect`.

Cheers!


dsyer:
> prepending of "servo." on the front of metric names (that may have already been prepended with "servo.")

Looks like a bug in its own right. We should just fix it (check the metric name before prepending it)?

adriancole:
Thanks for making progress towards shiny things. Atlas is cool.

I there an issue to link to about "Once Spring Boot is on Java 8"?

jkschneider:
@dsyer - If the current Servo integration needs to remain in a deprecated state for some time, by all means we should fix it.  Since it is undocumented, does anybody yet depend on it and can it just be removed outright?  The feature does look incomplete.  `ActuatorMetricWriter` drops certain meters and counters (anything that passes through the increment method) and passes histograms and timers through to Servo as gauges.

jkschneider:
I didn't realize that `EurekaHealthIndicator` was dependent on the metric name `counter.servo.discoveryclient_failed`.  I suppose we will need to decide on the status of the existing servo integration to fix this.

jkschneider:
This PR is now preceded by work in https://github.com/spring-cloud/spring-cloud-netflix/pull/531, which removes the dependency `EurekaHealthIndicator` has on Servo.

spencergibb:
@jkschneider how about a rebase and squash (I see some of our old commits). I think we have a proper eureka version, correct?

jkschneider:
@spencergibb I'll got a significant amount of changes locally that I'll try to push up tonight if I get a chance.

* Re-added servo support and flushed it out for Java 7 users
* Generalized the interceptors over both servo and spectator
* Added starter module

spencergibb:
@jkschneider looking forward to it.

jkschneider:
This branch has gotten too messy and stale, closing and will reopen a new PR.

PR #520, opened 2015-09-08
Changes in Eureka InstanceStatus originating from the server cause ApplicationEvent

jkschneider:
Changes to a service's InstanceStatus can originate from the deployed environment.

The canonical example is when an ASG containing a collection of instances visible in Eureka is disabled from AWS (say, temporarily for a red/black type deployment pipeline).  In this case, the application on these various instances remain running, but the developer may want to disconnect the application from consuming messages from queues, etc.

Cheers!

### DIFF
--- a/spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
+++ b/spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
@@ -1,6 +1,7 @@
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
 org.springframework.cloud.netflix.archaius.ArchaiusAutoConfiguration,\
 org.springframework.cloud.netflix.config.EurekaClientConfigServerAutoConfiguration,\
+org.springframework.cloud.netflix.eureka.EurekaApplicationEventAutoConfiguration,\
 org.springframework.cloud.netflix.config.DiscoveryClientConfigServiceAutoConfiguration,\
 org.springframework.cloud.netflix.feign.ribbon.FeignRibbonClientAutoConfiguration,\
 org.springframework.cloud.netflix.feign.FeignAutoConfiguration,\
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/eureka/EurekaApplicationEventAutoConfigurationTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/eureka/EurekaApplicationEventAutoConfigurationTests.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2013-2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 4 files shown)


spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
```diff
@@ -15,7 +15,8 @@ org.springframework.cloud.bootstrap.BootstrapConfiguration=\
 org.springframework.cloud.netflix.config.DiscoveryClientConfigServiceBootstrapConfiguration
 
 org.springframework.cloud.client.discovery.EnableDiscoveryClient=\
-org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration
+org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration,\
+org.springframework.cloud.netflix.eureka.EurekaApplicationEventAutoConfiguration
```
spencergibb: Should this go under `org.springframework.boot.autoconfigure.EnableAutoConfiguration`?

jkschneider: I was just going to submit a separate PR to do the opposite.  `EurekaClientAutoConfiguration` should probably move down to run only under `EnableDiscoveryClient`.  As it stands right now, `EurekaClient` is initializing even when `@EnableEureka` is not applied, and requires an extra `eureka.client.enabled=false` in properties.  This is my fault - stemming from the work on bootstrapping `EurekaClient` before SmartLifecycle phase 0.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/eureka/EurekaApplicationEventAutoConfigurationTests.java
```diff
@@ -0,0 +1,79 @@
...
+    @Autowired
+    ApplicationListener<EurekaStatusChangedEvent> listener;
+
+    @Test
+    public void changesToInstanceStatusInEurekaCauseApplicationEventsToBeFired() throws InterruptedException {
+        when(eurekaClient.getInstanceRemoteStatus())
+                .thenReturn(InstanceInfo.InstanceStatus.UP)
+                .thenReturn(InstanceInfo.InstanceStatus.DOWN);
+
+        Thread.sleep(500);
```
spencergibb: Maybe use a real listener, rather than mocked with a `CountDownLatch`? Trying to avoid `sleep()` in tests.

spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
spencergibb: :-) I guess I'm surprised that putting extra configuration here works. It wasn't intended to work.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/eureka/EurekaApplicationEventAutoConfigurationTests.java
jkschneider: Will do

spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
spencergibb: That doesn't mean we can ensure that it continues to work.

spencergibb: We have at least one `@AutoConfigureAfter(EurekaClientAutoConfiguration.class)`, that requires it to be auto-configured before.

jkschneider: You've sold me on moving the `EurekaApplicationEventAutoConfiguration`.  I've made it conditional on `EurekaClient` and `@AutoConfigureAfter(EurekaClientAutoConfiguration.class)`.

PR #531, opened 2015-09-09
Eureka 1.2.2 - Removes need for MetricReader in EurekaHealthIndicator

jkschneider:
Eureka 1.2.2 adds a couple more methods to `DiscoveryClient` that give us the ability to detect a failing connection to the Eureka server without having to track the state of a Servo monitor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 		<spring-cloud-config.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-config.version>
 		<main.basedir>${basedir}</main.basedir>
 		<archaius.version>0.6.5</archaius.version>
-		<eureka.version>1.2.0</eureka.version>
+		<eureka.version>1.2.4</eureka.version>
 		<feign.version>8.10.0</feign.version>
 		<hystrix.version>1.4.14</hystrix.version>
 		<ribbon.version>2.1.0</ribbon.version>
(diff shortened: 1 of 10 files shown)


jkschneider:
Necessary precursor to https://github.com/spring-cloud/spring-cloud-netflix/pull/517.

spencergibb:
I guessed the upgrade to 1.2.2 would affect eureka server because of the upgrades to the jackson codecs. Our metadata serializers are having trouble. When I run in the IDE it works fine. Running from a jar fails.

spencergibb:
The failures I'm seeing are from this repo: https://github.com/spring-cloud-samples/eureka
Tests fail for the build there as well.

spencergibb:
There is currently no way I can see to plug in our own serializers/deserializers, even with hacking. With static singletons we could change them.  There is no DI here because `CodecWrappers` and `EurekaAccept` don't allow customization. /cc @dsyer

jkschneider:
@spencergibb I'll take a look and try to make accomodations in Eureka itself if necessary.

spencergibb:
We either need to be able to plug in serializers/deserializers, or all for a custom `DataCenterInfo` class ala `AmazonInfo`.

jkschneider:
@spencergibb Waiting on Eureka [#646](https://github.com/Netflix/eureka/pull/646) to be merged.

jkschneider:
Approach has changed much since this initial implementation - closing this PR and opening another.

Issue #412, opened 2015-06-28
Missing HTTP PATCH request method in Zuul RibbonRoutingFilter


sebastian-sch:
Do you see any possibility that the ```RibbonRoutingFilter``` will support HTTP PATCH. 

The problem is, that the enum ```Verb``` from ```com.netflix.client.http.HttpRequest``` don't support HTTP PATCH. 

If it is not possible, perhaps it should be documented, which HTTP Verbs Ribbon & Zuul supports.

This issues is related to #307.

zampettim:
This is going to be a problem for me as well, since I'm using Spring Data REST, and use of the PATCH method is the documented way to update resources without having to supply a fully populated object.

hrandika:
This is coursing trouble to me too when one service want to update something in other service. Any way around this ?

schoberwirt:
We use ApacheHTTP as the client for ribbon. 
Dependencies need to be pushed.
 <dependency>
      <groupId>com.netflix.feign</groupId>
      <artifactId>feign-httpclient</artifactId>
      <version>8.3.0</version>
    </dependency>

now you should be able to do sthg like.
Don't use okhttp if you also want to use newrelic. No introspection there.

 @Bean
 public Client feignRibbonClient() {
        return RibbonClient.builder().delegate(new ApacheHttpClient()).lbClientFactory(cachingLBClientFactory()).build();
 }

zampettim:
@schoberwirt Where is the ```cachingLBClientFactory()``` coming from?

zampettim:
Ok, here is what I had to do to get a FeignClient implemenation to work. I haven't tested in a Zuul case yet, but it should be similar. And that is next on my plate.

I had to up-rev the following dependencies.
Updated com.netflix.feign.{feign-core,feign-ribbon,feign-slf4j,feign-httpclient} to 8.5.0
I also had to add feign-httpclient to the dependencies.
That then forced the need to update org.apache.httpcomponents to account for missing methods.
So updated org.apache.httpcomponents.httpclient to 4.5 and updated apache.httpcomponents.httpcore to 4.4.1 (under the assumption that httpclient will work best with the latest httpcore.

Then I had to add the following to my Spring Application classes

    @Bean
    public Client feignRibbonClient() {
        return RibbonClient.builder().delegate(new ApacheHttpClient()).build();
    }

[...]

PR #543, opened 2015-09-16
implementation of an Apache http client zuul ribbon command

ChristianLohmann:
This is an implementation of a ```RibbonCommand``` and a factory which makes use of a loadbalancing http client based on apache's http-components http-client.
With this command/factory it is possible to route PATCH http requests via zuul/ribbon as well.

This would fix https://github.com/spring-cloud/spring-cloud-netflix/issues/412

Any recommendations/improvements are welcome


ChristianLohmann:
This one is a bit hacky, especially the use of the RibbonApacheHttpResponse. So, please share any ideas to make this better

spencergibb:
@ChristianLohmann could you move your new classes to an `apache` package?

ChristianLohmann:
@spencergibb alright, moved classes to a new package

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonApacheHttpResponse.java
```diff
@@ -0,0 +1,151 @@
...
+import java.io.InputStream;
+import java.lang.reflect.Type;
+import java.net.URI;
+import java.util.Collection;
+import java.util.Map;
+
+import org.apache.http.Header;
+import org.apache.http.HttpResponse;
+
+import com.google.common.collect.Lists;
```
spencergibb: Can you remove dependencies on `com.google.*` classes?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonApacheHttpResponse.java
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	@Override
+	public boolean hasEntity() {
+		return hasPayload();
+	}
+
+	@Override
+	public <T> T getEntity(final Class<T> type) throws Exception {
+		return null;
```
spencergibb: is this a problem?

spencergibb:
What do you consider hacky about `RibbonApacheHttpResponse`?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonApacheHttpResponse.java
ChristianLohmann: Sure, i will. Didn't have much time the last weeks.

ChristianLohmann:
With hacky, i meant exactly the ```RibbonApacheHttpResponse``` class which conveniently implements 2 interfaces to save another transfer object.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonApacheHttpResponse.java
ChristianLohmann: This is no problem because they are not in use, but a bit ugly. As described above these methods have their origin in ```com.netflix.client.http.HttpResponse``` which i am forced to implement.

spencergibb:
Squashed and merged via f46e8e8e1f6830334fd2f5651a402b65b9fb5e2d

yairogen:
@spencergibb when can we expect this pull request to be in a release?

spencergibb:
@yairogen It is available in SNAPSHOTS now and will be in M4, not sure when that will be.

PR #548, opened 2015-09-21
Hystrix spring cloud stream

spencergibb:
Move from amqp specific implementations of they hystrix/turbine stream to spring-cloud-stream based implementations.  To choose the broker, include one of `spring-cloud-starter-stream-{kafka|rabbit|redis}`. Tested locally with rabbit and redis.

This doesn't solve the multiple rabbit connection factory problem (ie not just which transport should stream use, but which rabbit connectionfactory if there are multiples).

/cc @dsyer @scottfrederick @mstine @markfisher


spencergibb:
Closed. Superseded by 639a87bfa1e0a6fc2531f1f96b5b4af7324be453.

Issue #288, opened 2015-03-31
Ability to configure feign.RequestInterceptor specific to a given feign client


xyloman:
There is a few use cases where we need to add different headers or authentication mechanisms on a per client basis.  Currently the FeignClientFactoryBean appears to apply its configuration to all feign clients that it constructs.  This was problematic in that some services that we want to use feign for require a different set of RequestInterceptors to one set of clients verse another.  Is there a way to accomplish this today?  We were able to work around this issue by using the feign builder directly, with the only thing we are customizing is the list RequestInterceptor from the other FeignClients.

spencergibb:
Good idea.  We have a similar mechanism for `@RibbonClient` we could reuse.

joaoevangelista:
Maybe we could register the `RequestInterceptors` through a value in `@RibbonClient` ? So it's clear which client use which interceptor

spencergibb:
@joaoevangelista we need to maintain separation between Feign and Ribbon.  I'm imagining the same type of facility that `@RibbonClient` uses for `@FeignClient`.  We could probably generalize it for use with other annotations.

joaoevangelista:
@spencergibb Really sorry! I meant `@FeignClient`, I got confused a bit, but yea, a generalized way would be nice too

marcosbarbero:
Hi everybody, how is this issue going?

dwelch2344:
Just wanted to throw in my vote on this. We've got a mixture of some ribbon backed FeignClients and a few that are hard coded to 3rd party APIs, so this would be a major win.

spencergibb:
Yeah, I see this as our biggest feature for feign.  I'm working on getting ready for a conference week after next, so it might not get started until after OSCON.

sbuettner:
Same as @dwelch2344. Would really like to see support for that. Maybe by leveraging springs @Qualifier annotation.

lpborges:
This is exactly what I need.
I need to setup a FeignClient that use Basic Authentication, and this feature can help to setup a Feign BasicAuthRequestInterceptor different for each client.
Maybe some options for authentication on the FeignClient annotation can help too.

I figured to setup Authentication using headers in the RequestMapping:
```
@FeignClient(url="${service.url}")
public interface ProductsClient {

    @RequestMapping(method = RequestMethod.GET, value = "/item/{idProduct}", consumes = "application/json",
            headers="Authorization: Basic ${service.auth}")
[...]

PR #572, opened 2015-10-02
Allow overriding default feign beans.

spencergibb:
Similar to `@RibbonClient`.

For review @dsyer @adriancole while I work on docs.

Allows you to attache a configuration class to `@FeignClient` and default configuration classes to `@EnableFeignClients`.

```
@EnableFeignClients(defaultConfiguration=MyDefaultFeignConfig.class)
//...

@FeignClient(name="foo", configuration=FooClientConfig.class)
interface FooClient {/*...*/}

@Configuration
class FooClientConfig {
  @Bean
  RequestInterceptor feignRequestInterceptor() {
    return new BasicAuthRequestInterceptor("user", "pass");
  }
}
```

Anything that can be used in the [`Feign.Builder`](https://github.com/Netflix/feign/blob/master/core/src/main/java/feign/Feign.java#L83) can be created as a bean.

*TODO*: docs and javadoc updates.

fixes gh-288

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -767,6 +767,67 @@ don't want to use Eureka, you can simply configure a list of servers
 in your external configuration (see
 <<spring-cloud-ribbon-without-eureka,above for example>>).
 
+[[spring-cloud-feign-overriding-defaults]]
+=== Overriding Feign Defaults
+
+A central concept in Spring Cloud's Feign support is that of the named client. Each feign client is part of an ensemble of components that work together to contact a remote server on demand, and the ensemble has a name that you give it as an application developer using the `@FeignClient` annotation. Spring Cloud creates a new ensemble as an
+`ApplicationContext` on demand for each named client using `FeignClientsConfiguration`. This contains (amongst other things) an `feign.Decoder`, a `feign.Encoder`, and a `feign.Contract`.
+
+Spring Cloud lets you take full control of the feign client by declaring additional configuration (on top of the `FeignClientsConfiguration`) using `@FeignClient`. Example:
+
+[source,java,indent=0]
+----
+	@FeignClient(name = "stores", configuration = FooConfiguration.class)
+	public interface StoreClient {
+        //..
+	}
+----
+
+In this case the client is composed from the components already in `FeignClientsConfiguration` together with any in `FooConfiguration` (where the latter will override the former).
+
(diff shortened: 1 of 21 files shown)


adriancole:
nice

hah funny thing is that feign started as dagger, removed DI, and now has DI again :) At least this time it is a less obscure DI :P

spencergibb:
@adriancole I like that it doesn't have it baked in, made my life way easier.

dsyer:
Isn't `@RibbonClient` different (designed to support multiple declarations per app)? Since this is an `@Enable*` annotation it would be nice if it was idempotent, and I have a feeling the current implementation would cause issues if there were more than one in the same app. Given that constraint, does it need to be a child context?

dsyer:
Looking at #288 it seems like we do need something akin to `@RibbonClient`. I was just misreading the example code above. Ignore me. If you're not ignoring me, LGTM.

joaoevangelista:
Pretty good, the possibility to declare configurations on both package level on @Enable and method level will make possible fine tunned configs. Just a thing that popped up, what happens if we declare more than one bean of the same type on a configuration class? It will pick up the first and ignore the rest, or apply all, maybe an exception is thrown?

dsyer:
Is a normal Spring application context so the same rules apply. Probably 2 beans of the same type will lead to a runtime autowiring error.

spencergibb:
You can create multiple `feign.RequestInterceptor` beans, since they get injected as a list.

spencergibb:
Closed via 810502c1f326fe9f56ae9b59e3c4dfaae0e1a869

xyloman:
Finally had a chance to look at this.  Appears to address all of our concerns.  What is the ETA for inclusion in a release?

spencergibb:
@xyloman thanks for the follow up. We're planning on another milestone shortly.

mikeyudin:
Any updates on when this will be in a release?

spencergibb:
It's part of Brixton which has had a number of milestones. M4 currently. Q1 2016

mikeyudin:
Great! Thank you.

kummerer94:
This does not seem to be included in M4, any plans when this will be released yet?

spencergibb:
It is in M4, there's an M5 too, RC1 hopefully this week. If you think there is a bug, please open a new issue.

kummerer94:
Thanks for the immediate answer. I was trying to use `@EnableFeignClients(defaultConfiguration=MyDefaultFeignConfig.class)`.

In terms of the versions given on this page http://cloud.spring.io/spring-cloud-netflix/, which version do I need to use?

spencergibb:
`1.1.0.M4` should work, the latest is `1.1.0.M5`

PR #574, opened 2015-10-05
Enables monitoring of multiple services on a single physical box.

micheal-swiggs:
By setting `turbine.instanceInsertPort.default=false` and `turbine.combineHostPort=true` services registered with Eureka running on the same server(ie same hostname) can be monitored with turbine. I'm not sure if this is a good way of enabling this. 

Another option is to change `TurbineConfiguration` to:

	@Autowired
	private InstanceDiscovery iDiscovery;

	@Override
	public void start() {
		PluginsFactory.setClusterMonitorFactory(new SpringAggregatorFactory());
		PluginsFactory.setInstanceDiscovery(iDiscovery);
		TurbineInit.init();
	}

And allow the user to inject their own `InstanceDiscovery` implementation, e.g a subclass of `EurekaInstanceDiscovery`.


spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/EurekaInstanceDiscovery.java
```diff
@@ -61,12 +61,14 @@
 
 	private final EurekaClient eurekaClient;
 	private final Expression clusterNameExpression;
+    private final boolean combineHostPort;
```
spencergibb: We use tabs.

spencergibb:
Have you signed the [contributor's agreement](https://support.springsource.com/spring_committer_signup)?

micheal-swiggs:
Just signed it. My confirmation number is 141620151006073646.

spencergibb:
@micheal-swiggs I changed turbine recently. I thought I'd just add your changes. When I added this test to `EurekaInstanceDiscoveryTest`:
```
	@Test
	public void testCombineHostPort() {
		turbineProperties.setCombineHostPort(true);
		EurekaInstanceDiscovery discovery = getEurekaInstanceDiscovery();
		String appName = "testAppName";
		int port = 8080;
		String hostName = "myhost";
		InstanceInfo instanceInfo = builder.setAppName(appName)
				.setHostName(hostName)
				.setPort(port)
				.build();
		Instance instance = discovery.marshall(instanceInfo);
		assertEquals("hostname is wrong", hostName+":"+port, instance.getHostname());

		String urlPath = SpringClusterMonitor.ClusterConfigBasedUrlClosure.getUrlPath(instance);
		assertEquals("url is wrong", "http://"+hostName+":"+port+"/hystrix.stream", urlPath);
	}
```

It fails with
```
java.lang.RuntimeException: Configured to use port, but port or securePort is not in host attributes

	at org.springframework.cloud.netflix.turbine.SpringClusterMonitor$1.getUrlPath(SpringClusterMonitor.java:108)
	at org.springframework.cloud.netflix.turbine.EurekaInstanceDiscoveryTest.testCombineHostPort(EurekaInstanceDiscoveryTest.java:99)
```

If you want to rework it so a test like that doesn't fail, go for it. Otherwise, you can now create your own `InstanceDiscovery` beans.

micheal-swiggs:
I will try to rework this. So if I add this test into my branch I should be able to reproduce the error?

spencergibb:
You should. the `getUrlPath` fails if there is no `port` metadata.

micheal-swiggs:
I've added the tests and fixed them.

spencergibb:
@micheal-swiggs, just looking at this and realized you've made this from the M1 branch, can you do rebase to master instead?

micheal-swiggs:
I will try.

micheal-swiggs:
I was not able to rebase this branch. So I have created a new pull request forked from master with the changes of this pull request.

codependent:
Last November I came across this same problem and found an [open issue](https://github.com/Netflix/Turbine/issues/88) in Turbine's project.

I finally submitted a [PR](https://github.com/Netflix/Turbine/pull/105) addressing it. The changes introduced here definitely fixed the problem but in my opinion it would make sense that the solution were on Turbine's side. Why? Because this PR only works when you have Eureka, but doesn't when using other discovery server such as Consul.

What do you think?

PR #577, opened 2015-10-08
Use instanceof for better readability

alexVengrovsk:


spencergibb:
Does it provide the same functionality as `parameterAnnotation.annotationType()`?

alexVengrovsk:
Yes. It compares the annotation's type with expected type to deside which block of code to execute

Issue #578, opened 2015-10-08
DiscoveryClient is automatically configured if spring-cloud-starter-eureka is used


olivergierke:
I've tried to build an app that uses service discovery in a Spring profile only. I have added `spring-cloud-starter-eureka` as dependency and see a `DiscoveryClient` now being created by default (complaining about no Eureka instance being available).

It looks like this is related to `EurekaClientAutoConfiguration` being enabled by default (due to `@ConditionalOnProperty(value = "eureka.client.enabled", matchIfMissing = true)`. I could turn that off by setting the property to `false` but I was originally expecting the `DiscoveryClient` not being created unless I explicitly enable it using `@EnableDiscoveryClient`.

aheusingfeld:
+1 on that!

brenuart:
+1

PR #580, opened 2015-10-08
Move DiscoveryClient creation so it is guarded.

spencergibb:
Move it back to EurekaDiscoveryClientConfiguration.

fixes gh-578

@dsyer I had to muck with the config server discovery stuff, so I'd like your eyes on it.

@jkschneider It was you that moved creation to `EurekaClientAutoConfiguration` to begin with. The trouble we're running into is that it's getting created even if `@EnableEurekaClient` isn't being used. I moved it to `DiscoveryClientConfiguration` which gets imported by `EurekaDiscoveryClientConfiguration`. You could import it if you need it earlier.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/config/DiscoveryClientConfigServiceBootstrapConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/config/DiscoveryClientConfigServiceBootstrapConfiguration.java
@@ -25,7 +25,9 @@ import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.cloud.config.client.ConfigClientProperties;
 import org.springframework.cloud.config.client.ConfigServicePropertySourceLocator;
+import org.springframework.cloud.netflix.eureka.DiscoveryClientConfiguration;
 import org.springframework.cloud.netflix.eureka.EurekaClientAutoConfiguration;
+import org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.event.ContextRefreshedEvent;
@@ -45,7 +47,7 @@ import lombok.extern.apachecommons.CommonsLog;
 @ConditionalOnClass({ EurekaClient.class, ConfigServicePropertySourceLocator.class })
 @ConditionalOnProperty(value = "spring.cloud.config.discovery.enabled", matchIfMissing = false)
 @Configuration
-@Import(EurekaClientAutoConfiguration.class)
+@Import({EurekaClientAutoConfiguration.class, DiscoveryClientConfiguration.class})
 @CommonsLog
 public class DiscoveryClientConfigServiceBootstrapConfiguration {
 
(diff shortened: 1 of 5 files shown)


jkschneider:
I'm a little unclear now if EurekaClient will be available before SmartLifecycle phase 0.  Surely we can achieve both goals, right?

spencergibb:
`EurekaDiscoveryClientConfiguration` implements smart lifecycle. The beans it creates I don't think are involved in `Lifecycle`. I may be wrong.

jkschneider:
Prior to 478 being merged, `EurekaClient` was a bean created by `EurekaDiscoveryClientConfiguration` and it was not available until phase 0.  It is unclear to me if beans created by `@Import`s are available before.

spencergibb:
Maybe the solution is to move the lifecycle to it's own bean rather than the configuration.

jkschneider:
Effectively I think that is what we need to do.  Alternatively, move `EurekaClient` creation to its own configuration, point `@EnableEurekaClient` at it, and condition the `EurekaDiscoveryClientConfiguration` on the presence of `EurekaClient`.

spencergibb:
pointing `@EnableEurekaClient` won't work. We have `@EnableDiscoveryClient` that doesn't know anything about s-c-netflix.

spencergibb:
`DeferredImportSelector` might be the culprit, not the fact that `EurekaDiscoveryClientConfiguration` implements `Lifecycle`.

> A variation of {@link ImportSelector} that runs after all {@code @Configuration} beans have been processed. This type of selector can be particularly useful when the selected imports are {@code @Conditional}.

spencergibb:
Also, `@CondtionalOnBean` shouldn't work on `@Configuration` classes.

spencergibb:
maybe using `@EnableEurekaClient` might work if we bypassed `@EnableDiscoveryClient`. How do you reproduce the missing `EurekaClient` problem?

jkschneider:
Simply inject EurekaClient into a @Component that is not a Lifecycle.

jkschneider:
> DeferredImportSelector might be the culprit, not the fact that EurekaDiscoveryClientConfiguration  implements Lifecycle.

This might be the case; it never occurred to me.

spencergibb:
8fae07de3c07c9efdcbc362427b699d720a75685 doesn't replicate the missing bean. What am I doing differently. Does it need to be in something else that is autoconfigured?

spencergibb:
injecting in to an auto-config `@Configuration` works for me as well.

jkschneider:
I'm stumped now.  Perhaps the bean was available but not fully initialized previously?  Or perhaps the @Import is resolved ahead of Lifecycle.  If you fold everything in `DiscoveryClientConfiguration` back into `EurekaDiscoveryClientConfiguration` does the test fail?

jkschneider:
Looks great :+1:

spencergibb:
from our slack conversation

> spencergibb [4:07 PM] so before, there used to be a proxy on `DiscoveryClient` when the old static singleton was in use.
> We don’t need that anymore with the new DI refactorings.
> jkschneider [4:07 PM] :thumbsup:
> spencergibb [4:07 PM] THAT wasn’t populated until the lifecycle started
> spencergibb [4:07 PM] so maybe it isn’t an issue anymore
> spencergibb [4:08 PM] I had forgotten about the proxy.

spencergibb:
Closed via a96666b902b2806f0082b482756419f198fecb6c

PR #585, opened 2015-10-09
fix #584

### LINKED ISSUE #584: Feign clients are not registered if new `clients` annotation element is unspecified

mbenson:
not sure of the best way to get this into the unit test suite


spencergibb:
Nice catch.

spencergibb:
Closed with added test via c051cb83d215dd882e13cd09c115436187dc9e0a

PR #587, opened 2015-10-12
Polish

alexVengrovsk:
It is good thing to make this checks because of expensive concatenation of String objects (debug messages).


alexVengrovsk:
I made a mistake by using java8's constructions

PR #588, opened 2015-10-12
Add isDebugEnabled() check

alexVengrovsk:
It is good thing to make this checks because of expensive concatenation of String objects (debug messages).


spencergibb:
Merged via ce4076621a4895aea1c8dca855243b0497a00341

PR #604, opened 2015-10-18
Feign OkHttp client configuration

jmnarloch:
There is another one - Feign OkHttp client configuration.

Pretty simple. One more thing that I was thinking was to somehow made the httpclient and okhttp setup exclusive, so when you turn one explicitly then you will disable the other.

At this point to enable okhttp you would have to setup your application as fallows:

```
feign.httpclient.enabled=false
feign.okhttp.enabled=true
```

It would be simpler if setting the feign.okhttp.enabled=true would automatically disable the httpclient.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -282,6 +282,11 @@
 				<version>${feign.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>com.netflix.feign</groupId>
+				<artifactId>feign-okhttp</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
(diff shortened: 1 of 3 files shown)


spencergibb:
Merged with a few changes via 966e03ba64d0dbc41860d480511d01c184cc1c41.

The most likely scenario is that you will only have one of `feign-httpclient` or `feigh-okhttp` on your classpath, so I changed the `@ConditionalOnClass(OkHttpClient.class)` to use the feign `OkHttpClient` rather than the okhttp one. The only scenario where you would need to disable one or the other is if both are on the classpath.

PR #608, opened 2015-10-20
Add missing metadataMap property in doc example

ryanjbaxter:
The example was missing the metadaMap property.


spencergibb:
Actually, this has changed with upgrades to eureka that I worked on with Netflix. metadataMap still works for backwards compatibility.

ryanjbaxter:
I was using 1.2.5 and it did not seem to working, I have not ta newer version yet.  Do you think the change in 1.2.5?

spencergibb:
Right, it is still valid for 1.2.5

ryanjbaxter:
Hmmm it doesnt seem to be working for me, here is my eureka config

```
eureka:
  client:
    serviceUrl:
      defaultZone: ${vcap.services.ocr-eureka.credentials.uri:http://localhost:8761/eureka/}
  instance:
    hostname: ${vcap.application.uris[0]:localhost}
    nonSecurePort: 80
    instanceId: ${vcap.application.instance_id:${spring.application.name}:${spring.application.instance_id:${server.port}}}
```
Am I missing something?  I tried with Brixton and Angel

spencergibb:
So metadataMap should still be used for but master.

ryanjbaxter:
Even with using the latest Brixton snapshot build it does not seem to work....I am not sure how old the latest snapshot is but I would think it is pretty recent right?

spencergibb:
Hmm, I tried it and it seems to be working for me: `eureka.instance.instanceId=customer2`.

spencergibb:
image

ryanjbaxter:
Still not working for me, here is the source code to the client app
https://dl.dropboxusercontent.com/u/58852600/demo%203.zip

I am probably doing something dumb here....

spencergibb:
your app doesn't do it, mine does. Wonder what the difference is?

spencergibb:
nevermind, now yours is working :-(

spencergibb:
with Brixton.BUILD-SNAPSHOT

ryanjbaxter:
Can you share your Eureka server and client so I can try it on my machine?

spencergibb:
I'm using our samples: 

- https://github.com/spring-cloud-samples/configserver
- https://github.com/spring-cloud-samples/eureka
- https://github.com/spring-cloud-samples/customers-stores (just the customers service)

ryanjbaxter:
Are there some changes in the Eureka server needed to make this work?  I noticed that once I moved my Eureka server up to the Brixton Snapshot it started working?

spencergibb:
Yes, that's the case now that I think about it.

spencergibb:
So maybe, the best thing to do is to mention if either component is using 1.0.x it will need to use metadataMap.

ryanjbaxter:
I did another test and this only works with the latest Brixton Snapshot builds.  Looks like the documentation isn't versioned (there isn't a separate copy for the different builds). Right?  If that is the case than I guess we will have to clarify in the documentation which versions supports this property.

ryanjbaxter:
@spencergibb what do you think?

spencergibb:
We need to clarify the docs on master. The [Angel docs](http://projects.spring.io/spring-cloud/docs/1.0.3/spring-cloud.html) still have the right stuff.

spencergibb:
Do you want to update this PR or shall I have a shot at it?

ryanjbaxter:
Oh somehow I clicked on the wrong link to get to the Angel docs because I thought both the Angel and Brixton docs linked to the same place.  (Or was that just updated....)  Anyways I think if the docs are correct for the right versions than there is no need to update the Brixton docs since it works there.  Unless we want to update the Brixton docs to say that in M1 you need to include the metadataMap property.

spencergibb:
Closing in favor of the outcome of #642

Issue #570, opened 2015-10-02
eureka client instance registers incorrectly if 'server.port' is set to 0


kdvolder:
I created basic service that registers itself with a local Eureka server. The application .yml of my client looks like this:

    server:
      port: 8001
    spring:
      application:
        name: date-service
    eureka:
      client:
        service-url:
          defaultZone: http://localhost:8761/eureka/

That works.

When I change 'server.port' to '0' so the port is chosen dynamically then the service doesn't appear to register itself correctly anymore. (It still registers, but with the wrong port, 80).

Being able to use dynamic port would be nice, if one wants to run multiple instances of a service locally without having to create multiple configurations of it (i.e. just relaunch the same thing a few times and have each running instance dynamically pick a port).

spencergibb:
I just tried it and it worked. What version are you using?

spencergibb:
I recreated it. This used to work, so it is a regression. Though the workaround is to set a port. Random ports were built for testing.

shivangshah:
Hi @spencergibb I just tested this with Brixton M4 which has 1.1.0.M4 of cloud-netflix (which is where I am guessing this fix went in?) but I am still having this issue. Here's a test service that you can start up and register to a eureka server. Port is random (server.port is 0) but when it registers multiple instances with the same name/ip (locally) it overwrites them. Here's the test service: https://github.com/shivangshah/test-service

shivangshah:
Another weird thing that I found when I was setting server.port to actual values, lets say 8080 and 8081, from the eureka UI, it indeed does show 2 instances running BUT when you click on those instances, they both point to the same url of "info" .. and that is of the one that registered first .. in my case it was 8080 .. Not sure if that's how it's supposed to be?

spencergibb:
@shivangshah could you please open a new issue.

Canos:
I'm trying with Dalston.SR1 same happens.
If server.port=0 (random) eureka only show (and gives traffic) to the last registered, leaving formers unusable.

Can we reopen the issue or should i create a new one?

spencergibb:
Have you set `eurkea.instance.instanceId`? Otherwise with port=0 eureka will think it is one instance

Canos:
Perfect, an instanceId with a random value works. 
Sorry, i'm still a newbie.

thanks

TimB0:
after that port number is generated....is there anyway i can get that value?  If I use "${server.port}"  ill get 0.

[...]

PR #611, opened 2015-10-22
fix support for registering random ports with eureka.

spencergibb:
fixes gh-570 ... again


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/eureka/DiscoveryClientConfiguration.java
```diff
@@ -83,8 +82,7 @@ public EurekaClient eurekaClient(ApplicationInfoManager applicationInfoManager,
 		@SneakyThrows
 		@org.springframework.cloud.context.config.annotation.RefreshScope
 		public EurekaClient eurekaClient(ApplicationInfoManager applicationInfoManager,
-										 EurekaClientConfig config, EurekaInstanceConfig instance) {
-			applicationInfoManager.initComponent(instance);
```
spencergibb: @dsyer `applicationInfoManager.initComponent(instance)` is deprecated and was resetting the `InstanceInfo` (with port = 0). What does this fix? I'm assuming it doesn't reset when you call `/refresh`? Eureka config first works with this change.

spencergibb: `/refresh` seems to work ok.

spencergibb:
Closed via 3dfb8b5dc1484248b50738dacacb7efabf96c0ee

Issue #612, opened 2015-10-23
Feign: 204 No Content and HTTP HEAD responses returns null ResponseEntity.


jmnarloch:
Feign client mapped to following REST endpoints:

```
		@RequestMapping(method = RequestMethod.GET, value = "/noContent")
		ResponseEntity noContent() {
			return ResponseEntity.noContent().build();
		}

		@RequestMapping(method = RequestMethod.HEAD, value = "/head")
		ResponseEntity head() {
			return ResponseEntity.ok().build();
		}
```
With contract defined as follows:
```
		@RequestMapping(method = RequestMethod.GET, value = "/noContent")
		ResponseEntity noContent();

		@RequestMapping(method = RequestMethod.HEAD, value = "/head")
		ResponseEntity head();
```

Are going to return null as a result during runtime. I would rather expect that the response entity would be created with empty payload with all of the headers being populated for those cases.

jmnarloch:
So the problem was rather related to the client contract and the fact that the ResponseEntity didn't specify any type.

spencergibb:
@jmnarloch so using `ResponseEntity<Void>` worked? Not sure I understand your last comment.

jmnarloch:
Just to sum up, the behavior in case of contentless response on the client side requires you to specify the generic type - so it does not matter what that would be and ResponseEntity<Void> is somehow closes in saying - "I want the response, but I don't care about it's body".

Alternatively you could declare that you return ResponseEntity - so also you not interested in the actual response type. The only problem is during runtime and matching since ResponseEntity is not longer instance of ParameterizedType, just a simple Class instance. So the check fails and the overall invocation result of Feign client method is null.

spencergibb:
@jmnarloch thanks for the clarification.

PR #613, opened 2015-10-23
Feign unparameterized Response entity handling

jmnarloch:
Tests for issue: https://github.com/spring-cloud/spring-cloud-netflix/issues/612


jmnarloch:
Well, this might be a kind of edge case when I think of this now. A workaround to the previous version would be to define the result as ResponseEntity<Void>.

spencergibb:
Merged with a little polish via dc1f28a74ece8f152c0b23ed3250210e1273c28c

Issue #617, opened 2015-10-26
Support object parameters in feign


nickcodefresh:
Does spring-cloud-netflix support POJO object parameters?

```
@RequestMapping(value = "/api/events", method = RequestMethod.GET,)
EventList getEvents(@ModelAttribute @Valid EventQuery eventQuery)
````

I don't think it does. Feign seems to map the EventQuery object to the body of the request.

spencergibb:
It doesn't. See #609

nickcodefresh:
They're different issues, no?

spencergibb:
Yes, but similar in that they both want to support a spring parameter annotation.

Dreampie:
https://github.com/Netflix/feign/issues/329#event-561389611
so，we can custom user encoding，like json... can you help?

nickcodefresh:
Yes, that'll work

Issue #609, opened 2015-10-22
Is @MatrixVariable supported in Feign?


nickcodefresh:
If I have a method defined as:
```
@RequestMapping(value = "/api/events{matrixParam:.*}", method = RequestMethod.GET)
EventList getEvents(@MatrixVariable(pathVar = "matrixParam") Map<String, List<String>> matrixParams)
```

When I try to call this I get the exception

```
java.lang.IllegalArgumentException: Illegal character in path at index 35: http://mio-event-service/api/events{matrixParam:.*}
	at java.net.URI.create(URI.java:859)
```

Which implies that matrix parameters aren't supported. The RequestMapping definating works fine in the severing web app though.

Any ideas?

Thanks

nickcodefresh:
Looking at line 129 of ```org.springframework.cloud.netflix.feign.support.SpringMvcContract I see```

	// TODO: support spring parameter annotations?

So the answer appears to be no at the moment

cforce:
Would be nie to have that

Nick Smith <notifications@github.com> schrieb am Do., 22. Okt. 2015 3:19 PM:

nickcodefresh:
I'll fork and try to fix

mperrault2:
@nickcodefresh will your fix include support for @ModelAttribute (as described in #617)?
Thanks

Aloren:
If someone needs quick workaround for that -- we used custom Encoder in our feign client context:
```

    @Bean
    public Encoder encoder() {
        Encoder.Default defaultEncoder = new Encoder.Default();
        return new Encoder() {
            @Override
            public void encode(Object object, Type bodyType, RequestTemplate template) throws EncodeException {
                if(bodyType == MyCustomParam.class) {
                    encodeCustomParam((MyCustomParam) object, template);
                } else {
                    defaultEncoder.encode(object, bodyType, template);
                }
            }

            private void encodeCustomParam(MyCustomParam object, RequestTemplate template) {
                    MyCustomParam param = ((MyCustomParam) object);
                    template.query("var1", String.valueOf(param.var1));
                    template.query("var2", String.valueOf(param.var2));
            }
        };
    }
```

and method in feign client:
```
    @RequestMapping(method = GET, value = "/api", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
    MyCustomResponse getByCustomParam(MyCustomParam param);
```

sify21:
[...]

PR #619, opened 2015-10-30
Feign annotated parameter processors

jmnarloch:
Hey,

Since I've seen a couple of feature request related to Feign contract and supported Spring annotations ( https://github.com/spring-cloud/spring-cloud-netflix/issues/609 and https://github.com/spring-cloud/spring-cloud-netflix/issues/617) I though it would be worth of refactoring this functionality at this point so that:

* it would be simpler to support different annotations
* we would give users the extension points that they might need in the future

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/AnnotatedParameterProcessor.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/AnnotatedParameterProcessor.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2013-2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/AnnotatedParameterProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Collection;
+
+/**
+ * Feign contract method parameter processor.
+ *
+ * @author Jakub Narloch
+ */
+public interface AnnotatedParameterProcessor {
+
+    /**
```
spencergibb: Tabs, not spaces.

spencergibb:
Can you update the formatting? Then I'll have a look.

jmnarloch:
Should be better now.

spencergibb:
Merged with a bit of polish via 4fdb85e6786cdc01a85e7930b200b655b2ba4572

jmnarloch:
Thanks cool, thanks!

Dreampie:
@jmnarloch how to use？where is the document or example？

spencergibb:
This was an internal refactor to make future additions easier.

Dreampie:
thanks，When can I expect to use it?

spencergibb:
Unless you are writing a feign annotation processor, you won't.

Dreampie:
yes,Can I use @RequestParam and rewrite RequestParamParameterProcessor.java?

like this:
```
@FeignClient("serviceId")
public interface NoticeService {
   @RequestMapping(value = NoticeConstants.NOTICES_PATH, method = RequestMethod.GET)
   ResponseEntity<List<NoticeEntity>> findAll(@RequestParam("entity") NoticeEntity entity);
}
```

PR #625, opened 2015-11-03
Improvements to SimpleHostRoutingFilter (continued)

liebowitz:
Hopefully this diff is more readable.  
Thanks.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -88,7 +88,7 @@ public class SimpleHostRoutingFilter extends ZuulFilter {
 			.getInstance().getIntProperty(ZuulConstants.ZUUL_HOST_CONNECT_TIMEOUT_MILLIS,
 					2000);
 
-	private static final Timer CONNECTION_MANAGER_TIMER = new Timer(
+	private final Timer CONNECTION_MANAGER_TIMER = new Timer(
 			"SimpleHostRoutingFilter.CONNECTION_MANAGER_TIMER", true);
 
 	private ProxyRequestHelper helper;
@@ -133,7 +133,7 @@ public class SimpleHostRoutingFilter extends ZuulFilter {
 
 	@PreDestroy
 	public void stop() {
-		CONNECTION_MANAGER_TIMER.purge();
+		CONNECTION_MANAGER_TIMER.cancel();
 	}
 
 	@Override
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -19,15 +19,9 @@ package org.springframework.cloud.netflix.zuul.filters.route;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
-import java.net.Socket;
 import java.net.URL;
 import java.net.URLEncoder;
-import java.net.UnknownHostException;
-import java.security.KeyManagementException;
-import java.security.KeyStore;
-import java.security.KeyStoreException;
-import java.security.NoSuchAlgorithmException;
-import java.security.UnrecoverableKeyException;
+import java.security.SecureRandom;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


spencergibb:
@liebowitz have you signed the contributors agreement?

spencergibb:
Explain to me how you fixed ignored headers?

liebowitz:
Yes, agreement has been signed.

The RequestContext is stored in a thread local. When ignored headers are set in the config phase (e.g. below), they are not visible to the servlet executor thread pool.
```
@Configuration
@EnableZuulProxy
public class ZuulProxyConfig extends ZuulProxyConfiguration {
    @Bean
    @Override
    public SimpleHostRoutingFilter simpleHostRoutingFilter() {
        ProxyRequestHelper helper = new ProxyRequestHelper();
        helper.addIgnoredHeaders("X-Frame-Options");
        return new SimpleHostRoutingFilter(helper);
    }
}
```

spencergibb:
I believe you've broken some existing functionality

	/**
	 * Zuul context key for a collection of ignored headers for the current request.
	 * Pre-filters can set this up as a set of lowercase strings.
	 */
	public static final String IGNORED_HEADERS = "ignoredHeaders";

This allowed "pre" filters to add ignoredHeaders to the context. You've swapped that out for a single list in `ProxyRequestHelper`.

I think it makes sense to have both, but you can't remove the old way of working.

liebowitz:
Got it now, but it might be confusing to have both "globally" ignored headers and per-request ignored headers - I'll revert the change. 

If we added a getter for ProxyRequestHelper in SimpleHostRoutingFilter, it could be set in the request:
```
class CustomHostRoutingFilter extends SimpleHostRoutingFilter {
	@Override
	public Object run() {
		getHelper().addIgnoredHeaders("X-Ignored");
		return super.run();
	}
}
```
Or, delegate it to a protected method:
```
class CustomHostRoutingFilter extends SimpleHostRoutingFilter {
	@Override
	public Object run() {
		super.addIgnoredHeaders("X-Ignored");
		return super.run();
	}
}
```
Thoughts?

spencergibb:
I'd prefer the delegate to a protected method.

liebowitz:
Checked in.  Thanks.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -138,9 +115,25 @@ public SimpleHostRoutingFilter(ProxyRequestHelper helper) {
...
+				}
+				connectionManager.closeExpiredConnections();
+			}
+		}, 30000, 5000);
+	}
+
 	@PreDestroy
 	public void stop() {
-		CONNECTION_MANAGER_TIMER.cancel();
+		CONNECTION_MANAGER_TIMER.purge();
```
spencergibb: Why change this to `purge`?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -185,9 +177,66 @@ public Object run() {
 		return null;
 	}
 
+	protected PoolingHttpClientConnectionManager newConnectionManager() {
```
spencergibb: Why did you move this as opposed to just removing static? I can't tell if you've made changes or just moved it.

liebowitz: I moved it up so the diff would be clearer.

spencergibb: It's not clearer, IMO it's worse as I can't tell what changes (if any) you made to this method.

liebowitz: The entire method body changed, now using the updated HttpClient builder API.

spencergibb: right, and you have to explain the changes, I can't see them side by side.

liebowitz: newConnectionManager()
- Switched from deprecated ThreadSafeClientConnManager to PoolingHttpClientConnectionManager (thread-safe)
- Switched from deprecated SchemeRegistry to RegistryBuilder
- Switched from deprecated SSLSocketFactory to SSLConnectionSocketFactory

newClient()
- Switched from deprecated DefaultHttpClient to CloseableHttpClient 
- Switched from deprecated HttpParams to RequestConfig class.

spencergibb:
Merged with a little polish via c93a19c0d1fbfd34bbc1aba2b1b6223872f62589

PR #633, opened 2015-11-10
Fix flaky Servo tests by avoiding DefaultMonitorRegistry

jkschneider:
Static state in DefaultMonitorRegistry made it too difficult to clear between independent test executions.  

Instead, we can use the `SimpleMonitorRegistry` that holds no static state.  To accomplish this in tests, we also have to allow for a `MonitorRegistry` to be injected onto `ServoMonitorCache`.


spencergibb:
Merged with some polish via 55667497424a2368fb35e2e97141a9f6e8e61734

PR #639, opened 2015-11-11
Rest template metrics collection fixes

### LINKED ISSUE #636: RestTemplateUrlTemplateCapturingAspect should only be applied if AspectJ is available
RestTemplateUrlTemplateCapturingAspect should only be applied if AspectJ is available. Or maybe aspectj should be a mandatory dependency? Either way there are test failures in https://github.com/spring-cloud-samples/tests.

jkschneider:
Conditionalize rest template metrics collection on presence of CglibAutoProxy:

* If the application has `@EnableAspectJAutoProxy(proxyTargetClass = true)`, collection is safe
* A tacit assumption of `@EnableAspectJAutoProxy` is that a direct dependency on **aspectjweaver** is also present

Fixes #636

The around advice applied by the rest template metrics collection aspect was also not returning the value of the advised method.  This has been fixed and the test updated.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/metrics/MetricsInterceptorConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/metrics/MetricsInterceptorConfiguration.java
@@ -15,6 +15,7 @@ package org.springframework.cloud.netflix.metrics;
 
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.actuate.metrics.reader.MetricReader;
+import org.springframework.boot.autoconfigure.aop.AopAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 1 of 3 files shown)


spencergibb:
@jkschneider @dsyer  does 36e4c7cf9307720dfca3758e796ee7521cab5d67 change the need to rely on `CglibAutoProxy`

dsyer:
I think we can still make changes here, but the change in 36e4c7c means that a) native proxies will work if necessary, but b) all beans of type `RestTemplate` will have that concrete type at runtime if user sets `spring.aop.proxy-target-class=true`.

spencergibb:
There's a merge conflict. Do I still need to include `@ConditionalOnBean(AopAutoConfiguration.CglibAutoProxyConfiguration.class)`?

jkschneider:
@spencergibb @dsyer's solution is better.  It is not needed

jkschneider:
@spencergibb No need to try to merge this necessarily, but we do need the change in 6e9045d, because currently the around advice causes `RestTemplate` to always return null!

spencergibb:
NP, I'll get it.

spencergibb:
Squashed and merged via 713e6b84615aa4f271f797f30289dafe1db00083

PR #656, opened 2015-11-19
Fix to allow RestTemplate to perform fail-over on REST calls.

tkvangorder:
This change will add fail-over support into the RestTemplate when Ribbon has been enabled.

A detailed description can be found here: 

[Stack OverFlow Link](http://stackoverflow.com/questions/33765049/spring-cloud-getting-retry-working-in-resttemplate/33786574#33786574)

- Removed the selection of a server in the RibbonClientHttpRequestFactory as this is done by RestClient
- Changed the RibbonHttpRequest to call exectuteWithLoadBalancer()

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
@@ -19,7 +19,6 @@ package org.springframework.cloud.netflix.ribbon;
 import java.io.IOException;
 import java.net.URI;
 
-import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.loadbalancer.LoadBalancerClient;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.client.ClientHttpRequest;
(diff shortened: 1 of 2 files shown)


spencergibb:
I'm hesitant to accept this because of the work done in `RibbonLoadBalancerClient` that you bypass.

tkvangorder:
Initially, I had left the code in the factory and just made the one change in the request. However, diving into the RestClient internals, it delegates to the LoadBalancerContext.getServerFromLoadBalancer() and was dropping into the else clause for a "Full Url Case" and since the code the factory had already resolved the URL to a concrete host, it did not consult the load balancer. 

The  "executeWithLoadBalancer()" is calling the method "LoadBalancerContext.reconstructURIWithServer() which looks similar to the code you have in the RibbonLoadBalancerClient. 


But, I have only had my head into this code for a few days, you would know a lot better than me.

tkvangorder:
One thing I noticed after the fact: LoadBalancerAutoConfiguration is decorating the RestTemplate with an interceptor (in addition to the RibbonAutoConfiguration setting the request factory). 
So with the changes in this pull request a REST call may go through two different calls to the load balancer? 

Once in the Netflix RestClient via the executeWithLoadBalancer.....and then again via the interceptor LoadBalancerInterceptor.intercept

In our application, I was creating a new RestClient that was NOT using the load balancing interceptor and relying only on the RestClient.

At the end of the day, if there is a better way to build Re-Try into the RestTemplate...that is really what I am looking for. 

Let me know if I can help.

spencergibb:
@tkvangorder yeah, thanks for the reminder. I'm going to close this, but keep your issue open and see what we can do.

tkvangorder:
Sounds good, I will continue playing with this on my end to vet out the use of the RestClient.

Thanks for your time and looking into this.

spencergibb:
NP, I appreciate the community involvement.

PR #661, opened 2015-11-22
Enables hystrix monitoring of multiple services on a single server.

micheal-swiggs:


spencergibb:
Original request with conversation #574

spencergibb:
I like more tests than code! Thanks again.

Issue #662, opened 2015-11-22
Two problems with gzip content in Zuul integration


jettro:
Hi I have two problems with gzip content. The response I try to proxy using Zuul contains a header in the response: content-encoding, not in the format with capitals like in the contant in the class:
https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java

Would it be possible to effectively make this check lowercase as well?

Another thing is in the class:
https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java

```
if (requestEncoding != null && requestEncoding.equals("gzip")) {
    isGzipRequested = true;
}
```
The response I get contains the encoding: gzip,deflate,sdch. Maybe we can change the equals to a contains?

PR #663, opened 2015-11-22
Fixes A problem I am facing with a header in the response containing …

jettro:
…content-encoding in lowercase letters. Now I first check for the format Content-Encoding and if that is not found I also check the lowercase version.

 Next to that there was an issue checking if a request is made for gzip content. If you ask for more than gzip, ie gzip,deflate,sdch. Now I check if what is asked contains gzip instead of equals to.

 Fixed gh-662

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
@@ -57,6 +57,8 @@ public class ProxyRequestHelper {
 
 	public static final String CONTENT_ENCODING = "Content-Encoding";
 
+	public static final String CONTENT_ENCODING_LOWER = "content-encoding";
+
 	private TraceRepository traces;
 
 	public void setTraces(TraceRepository traces) {
@@ -129,8 +131,14 @@ public class ProxyRequestHelper {
 			RequestContext.getCurrentContext().setResponseDataStream(entity);
 		}
 		boolean isOriginResponseGzipped = false;
-		if (headers.containsKey(CONTENT_ENCODING)) {
-			Collection<String> collection = headers.get(CONTENT_ENCODING);
+		if (headers.containsKey(CONTENT_ENCODING) || headers.containsKey(CONTENT_ENCODING_LOWER)) {
+
+			Collection<String> collection = null;
+			if (headers.containsKey(CONTENT_ENCODING)) {
+				collection = headers.get(CONTENT_ENCODING);
+			} else {
+				collection = headers.get(CONTENT_ENCODING_LOWER);
+			}
 			for (String header : collection) {
 				if (HTTPRequestUtils.getInstance().isGzipped(header)) {
 					isOriginResponseGzipped = true;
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
```diff
@@ -129,8 +131,14 @@ public void setResponse(int status, InputStream entity,
 			RequestContext.getCurrentContext().setResponseDataStream(entity);
 		}
 		boolean isOriginResponseGzipped = false;
-		if (headers.containsKey(CONTENT_ENCODING)) {
-			Collection<String> collection = headers.get(CONTENT_ENCODING);
+		if (headers.containsKey(CONTENT_ENCODING) || headers.containsKey(CONTENT_ENCODING_LOWER)) {
```
spencergibb: spring's `HttpHeaders` is the way to go for case insensitive headers, to avoid this kind of logic. It is used in `buildZuulRequestHeaders` above.

jettro: I am looking into this, there is an issue I see. The method you mention receives a HttpServletRequest. The method getHeader of this object is case insensitive. However the setResponse methods receives a MultiValueMap, this does not have the case insensitive get method. Also looking at the HttpHeaders I do not see lowercasing support. Can you me more hints, I like to fix this the right way, but currently I do not see how.

spencergibb: Do what `buildZuulRequestHeaders` does: create a new `HttpHeaders` add them, iterate through all of the response headers, and then do your logic know that you have a case insensitive map. It's likely that in the future we'll use other headers besides Content Encoding .

jettro:
@spencergibb I have added the HttpHeaders, though I do not really like the implementation. I tried to combine the two loops in the method but I got unit tests errors in: ServletPathZuulProxyApplicationTests. Could not figure out why there was a problem. Had to do with the response not being a gzip response while the headers indicate it was. Curious to know if this is what you meant.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
```diff
@@ -49,221 +49,231 @@
...
-			}
-			catch (Exception e) {
-				log.debug(
-						"unable to encode uri path from context, falling back to uri from request",
-						e);
-			}
-		}
-		return uri;
-	}
+    public void setTraces(TraceRepository traces) {
```
spencergibb: We use tabs instead of spaces, can you clean that up?

spencergibb:
@jettro have you signed the [contributor's agreement](https://github.com/spring-cloud/spring-cloud-netflix#sign-the-contributor-license-agreement)?

jettro:
Hi @spencergibb, yes I did sign the contributor's agreement.

spencergibb:
Squashed and merged with some polish via a00fd2c3488cffa6ff9a25c197c719895d16473f

PR #669, opened 2015-11-29
Added check for invalid hostname

zanella:
The exception class has to be "fixed"

You may register a serviceId with invalid URI characters (e.g.: '_'), but the URI::getHost() fails when such chars are found.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
@@ -48,6 +48,9 @@ public class RibbonClientHttpRequestFactory implements ClientHttpRequestFactory 
 	public ClientHttpRequest createRequest(URI originalUri, HttpMethod httpMethod)
 			throws IOException {
         String serviceId = originalUri.getHost();
+        if (serviceId == null) {
+            throw new IOException("Invalid hostname in the URI [" + originalUri.toASCIIString() + "]");
+        }
         ServiceInstance instance = loadBalancer.choose(serviceId);
 		if (instance == null) {
 			throw new IllegalStateException("No instances available for "+serviceId);


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
```diff
@@ -48,6 +48,9 @@ public RibbonClientHttpRequestFactory(SpringClientFactory clientFactory, LoadBal
 	public ClientHttpRequest createRequest(URI originalUri, HttpMethod httpMethod)
 			throws IOException {
         String serviceId = originalUri.getHost();
+        if (serviceId == null) {
+            throw new IOException("Invalid hostname in the URI [" + originalUri.toASCIIString() + "]");
```
spencergibb: If this is a real problem we should do some validation much earlier, during startup rather than runtime.

zanella: Using a serviceId w/ underscore was exploding way futher in the call stack on the ConcurrentMap, this at least makes it clear why it's failing. The problem here seems to be the mismatch between what you can register as a serviceId and the expectation that's a valid hostname here...

Can you point me to where I should change so it happens at startup ? I am a *real* noob at spring dev

spencergibb:
@dsyer @adriancole @marcingrzejszczak what do you think? If using `RestTemplate`+ribbon, serviceid's need to be formatted as though they are valid dns hostnames. It's essentially a requirement because an exception is already thrown.

spencergibb:
@zanella have you signed the [contributor's agreement](https://github.com/spring-cloud/spring-cloud-netflix#sign-the-contributor-license-agreement)?

zanella:
@spencergibb hadn't, did now.

marcingrzejszczak:
@spencergibb - I've already had one idea how to check if the value passed for Ribbon is a valid hostname (since in Ribbon as you said the value has to be a hostname - in Zookeeper where you can have a nested structure we have the aliases approach). 

You can try to resolve the value to an address like here: https://github.com/4finance/micro-infra-spring/blob/54a5a2b5013491511efb908ef575c3bbb11be149/micro-deps/micro-deps-spring-config/src/main/java/com/ofg/infrastructure/discovery/UrlValidator.java#L7 . It should involve the entries in `/etc/hosts` or sth like that so all the custom entries would work. For insance in `/etc/hosts` you can have entries that have Polish chars (I've tested that ;) ) and it will still route to a proper ip.

spencergibb:
I don't think it should need to resolve to a real domain using `/etc/hosts`.

marcingrzejszczak:
It won't access that file. I'm just saying that it will work for valid entries in `/etc/hosts`

spencergibb:
@marcingrzejszczak as far as I can tell in tests, the `InetAdress` method requires a resolvable hostname.

adriancole:
Agree with validating early as possible,  or at least not too late :p the
prefix original hints we can.

Wrt the URI/URL woes,  this is a normal  problem and why sometimes I avoid
these types. Here's one attempt to get the features without the problems:

https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/HttpUrl.java

spencergibb:
Merged with a test via 8d259b9d8f3a43d29df48d7b62f09029fdca34e0

PR #670, opened 2015-11-30
Servo, Spectator, and Atlas docs

jkschneider:
@spencergibb draft of the documentation for the new metrics support. 

@brharrington would you mind giving this a quick once over for accuracy?


jkschneider:
According to @copperlight, a single Atlas standalone node running on an r3.2xlarge (61GB RAM) can handle roughly 2 million metrics per minute for a given 6 hour window.

spencergibb:
@jkschneider It might be a nice guage for users to base their needs. Are you going to make any more edits?

jkschneider:
@spencergibb Probably more useful to have examples than more docs at this point.  Thoughts on where those should go?

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -1322,3 +1322,239 @@ info:
...
+
+WARNING: If your code is compiled on Java 8, please use Spectator instead of Servo as Spectator is destined to replace Servo entirely in the long term.
+
+In Servo parlance, a monitor is a named, typed, and tagged configuration and a metric represents the value of a given monitor at a point in time. Servo monitors are logically equivalent to Spectator meters. Servo monitors are created and controlled by a `MonitorRegistry`. In spite of the above warning, Servo does have a link:https://github.com/Netflix/servo/wiki/Getting-Started[wider array] of monitor options than Spectator has meters.
+
+Spring Cloud integration configures an injectable `com.netflix.servo.MonitorRegistry` instance for you. Once you have created the appropriate `Monitor` type in Servo, the process of recording data is wholly similar to Spectator.
+
+==== Creation of Servo Monitors
+
+Servo does not provide a mechanism to retrieve an existing monitor by the unique combination of name and tags (which together form a `MonitorConfig` instance), so you need to manage this yourself, generally with some simple construct like:
```
brharrington: There are [DynamicCounter](https://github.com/Netflix/spectator/wiki/Servo-Comparison#dynamiccounter) and [DynamicTimer](https://github.com/Netflix/spectator/wiki/Servo-Comparison#dynamictimer) utility classes in servo for this use-case.

brharrington:
> Atlas standalone node running on an r3.2xlarge (61GB RAM) can handle roughly 2 million metrics per minute for a given 6 hour window. Not sure if this kind of expectation should be written into the docs or not.

That's a rule of thumb we use, it varies quite a bit depending on the data characteristics. We definitely need to add more information about some of those details to the Atlas wiki. Then it might be better to just link to the wiki so there is a single place to keep the information up to date.

spencergibb:
@jkschneider we usually put a basic sample in src/main/tests. More complex samples have gone in a [spring-cloud-samples](https://github.com/spring-cloud-samples) project. I wouldn't be opposed to have a samples module in the current project though. That could come in a later PR.

@brharrington linking to the wiki sounds like a good idea.

docs/src/main/asciidoc/spring-cloud-netflix.adoc
jkschneider: Thanks @brharrington.  Hadn't found this yet.  I'll update the core code to reflect this new knowledge as well.

spencergibb:
@jkschneider just let me know when you think it's good to merge. Since I'm not the technical expert on your new docs, and I'm a horrible proof reader :stuck_out_tongue_winking_eye:.

spencergibb:
ping @jkschneider thinking of a release soon to coincide (shortly after) with the boot 1.3.1 release. Probably next week some time.

jkschneider:
@spencergibb ready to ship!

anthonyikeda:
So I'm trying to follow the docs online (http://cloud.spring.io/spring-cloud-netflix/spring-cloud-netflix.html) But I can't seem to get my application integrated with Atlas.

I've added in @EnableAtlas, specified a netflix.atlas.uri value of http://localhost:7101/api/v1/publish/ (tried in my bootstrap.yml and application.yml) and using spring cloud netflix 1.1.0.M3

Stack trace here: https://gist.github.com/anthonyikeda/0f01df1045c9b231044b

I have a feeling the netflix.atlas.uri is not getting injected as the NullPointer falls on a Regex to compare the URI to an in built pattern.

Issue #689, opened 2015-12-03
Spring Boot 1.3 @FeignClient Error with @RequestBody


mike-boddin:
I upgraded a working project from spring boot 1.2.x to 1.3.
Now I get this error on startup:
```
java.lang.NullPointerException: null
	at org.springframework.cloud.netflix.feign.support.SpringMvcContract.processAnnotationsOnParameter(SpringMvcContract.java:160)
	at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:104)
	at org.springframework.cloud.netflix.feign.support.SpringMvcContract.parseAndValidateMetadata(SpringMvcContract.java:71)
	at feign.Contract$BaseContract.parseAndValidatateMetadata(Contract.java:61)
	at feign.hystrix.HystrixDelegatingContract.parseAndValidatateMetadata(HystrixDelegatingContract.java:24)
	at feign.ReflectiveFeign$ParseHandlersByName.apply(ReflectiveFeign.java:137)
	at feign.ReflectiveFeign.newInstance(ReflectiveFeign.java:55)
	at feign.Feign$Builder.target(Feign.java:179)
	at feign.Feign$Builder.target(Feign.java:175)
	at org.springframework.cloud.netflix.feign.FeignClientFactoryBean.loadBalance(FeignClientFactoryBean.java:126)
	at org.springframework.cloud.netflix.feign.FeignClientFactoryBean.getObject(FeignClientFactoryBean.java:143)
	at ...
```

The problematic part of my application is the following FeignClient:
```java
@FeignClient("feign.sample")
@RequestMapping("/feign")
public interface SampleFeignClient
{
    @RequestMapping(value = "/test", method = RequestMethod.POST)
    public String testAction(@RequestBody String string);

}
```

When I remove the ``@RequestBody String string`` part everything works fine.
Is there anything wrong with the processing of the ``@RequestBody`` annotation?

extract from the dependency-tree:
``|  +- org.springframework.cloud:spring-cloud-netflix-core:jar:1.1.0.M3:compile``

mike-boddin:
@tknowak maybe this is related to your problem here: https://github.com/spring-cloud/spring-cloud-netflix/issues/672#issuecomment-161218335

Koizumi85:
I had this Exception too yesterday and also solved it with removing the @RequestBody.
I looked at the code where the Exception is thrown and how it was in 1.2.
[...]

PR #690, opened 2015-12-03
Prevent NPE when no AnnotatedParameterProcessor is found for an Annotation

mike-boddin:
This is related to #689 and provides backwards-compatibility to spring boot 1.2.
This will fix an issue with FeignClients which contains ``@RequestBody`` Annotations.

I am not sure, if this is the right way to do, but it seems legit to me.


spencergibb:
@mike-boddin have you signed the [contributor's agreement](https://support.springsource.com/spring_committer_signup)?

mike-boddin:
Yes, I did it right now :-)

spencergibb:
Looks like this was taken care of here a43d9d86df51242442e6a8e92fd4c0eb7cdda5ed. Thanks for the contribution, though!

Issue #691, opened 2015-12-04
Disable redirect handling in RibbonLoadBalancingHttpClient


snussbaumer:
I think redrect handling should be either disabled or configurable in the RibbonLoadBalancingHttpClient, otherwise the ribbon client will **follow redirects**. If a backend server sends a redirect it's probably for the final user agent to follow the redirect not the Zuul Proxy.

Redirect handling can be disabled here : 
https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java#L42

replacing

```java
private final HttpClient delegate = HttpClientBuilder.create().build();
```

with

```java
private final HttpClient delegate = HttpClientBuilder.create().disableRedirectHandling().build();
```

Redirects are a bit of a corner case in a REST world ... but they happen nevertheless (if you have a Spring Oauth provider behind the zuul proxy for example), and I can not think of a case where following the redirect would be the right thing to do for the proxy (I may be wrong of course !)

PR #694, opened 2015-12-08
Enable configuration of the way HttpClient handles redirects

snussbaumer:
Default value is now DefaultClientConfigImpl.DEFAULT_FOLLOW_REDIRECTS
(false), it was true before (because it's true by default in HttpClient)

Also fixed the setting of connectTimeout and readTimeout who were equal
to zero before.

fixes gh-691

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
@@ -29,29 +29,27 @@ import com.netflix.client.AbstractLoadBalancerAwareClient;
 import com.netflix.client.RequestSpecificRetryHandler;
 import com.netflix.client.RetryHandler;
 import com.netflix.client.config.CommonClientConfigKey;
+import com.netflix.client.config.DefaultClientConfigImpl;
 import com.netflix.client.config.IClientConfig;
 import com.netflix.loadbalancer.ILoadBalancer;
-
-import lombok.Setter;
 
 /**
  * @author Christian Lohmann
  */
 public class RibbonLoadBalancingHttpClient
-		extends AbstractLoadBalancerAwareClient<RibbonApacheHttpRequest, RibbonApacheHttpResponse> {
+		extends
+		AbstractLoadBalancerAwareClient<RibbonApacheHttpRequest, RibbonApacheHttpResponse> {
 	private final HttpClient delegate = HttpClientBuilder.create().build();
 
-	@Setter
 	private int connectTimeout;
 
-	@Setter
 	private int readTimeout;
 
-	@Setter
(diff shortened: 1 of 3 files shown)


spencergibb:
@snussbaumer NP. When I get a bit more time, I'll add some comments. We'll need you to make some changes. Have you signed the [contributors agreement](https://github.com/spring-cloud/spring-cloud-netflix#sign-the-contributor-license-agreement)?

snussbaumer:
@spencergibb : I've just signed the contributors agreement. NP for the changes, go ahead !

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
```diff
@@ -64,9 +62,28 @@ public RibbonLoadBalancingHttpClient(final ILoadBalancer lb) {
 	}
 
 	@Override
+	public void initWithNiwsConfig(IClientConfig clientConfig) {
```
spencergibb: You call this explicitly in tests. My guess is that this doesn't get called when not in a test, can you confirm?

snussbaumer: Yes it is called because `RibbonLoadBalancingHttpClient`'s superclass is `IClientConfigAware`. When `SpringClientFactory` is used to get the instance it [calls `instantiateWithConfig`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/SpringClientFactory.java#L186) to create the instance the first time, and `instanciateWithConfig` [calls  the object's `initWithNiwsConfig`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/SpringClientFactory.java#L157) if it is `IClientConfigAware`.

`HttpClientRibbonCommandFactory` uses a `SpringClientFactory` to get the `RibbonLoadBalancingHttpClient` instance, hence `initWithNiwsConfig` is called. This is why a removed the unnecessary `client.setClientConfig(this.clientFactory.getClientConfig(serviceId));` there.

I've updated the tests to use a `SpringClientFactory` instead of calling directly `initWithNiwsConfig` to make it clear it's working.

I've also tested and seen the expected behaviour in my own application adding the following property in my boostrap.yml : 
```
ribbon:
  FollowRedirects: true/false
```

spencergibb: Thanks for the update.

spencergibb:
Closed via 0ddfb42d4df41132774459bcd1eebfa5e9bd5408

PR #699, opened 2015-12-10
Add Zuul Proxy regex serviceId to route mapping

### LINKED ISSUE #693: Zuul Proxy routes and service name convention
Hello,

I have a need near issue #458 but slightly different. I would like to implements some convention between route and services that are dynamically discovered :

domain-servicename-v1 (registered id in discoveryClient) --> /v1/domain/servicename/**
or
rest-service-v1 --> /v1/rest-service

After some research in spring-cloud-netflix-core I think a good place to implements this enhancement is in ProxyRouteLocator[#around line 199](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java#L197-214). The way routes are currently bind is a bit static and hard to override without duplicating ProxyRouteLocator code :

```java
String key = "/" + serviceId + "/**";
```

We could add some "overridable" logic here
```java
protected LinkedHashMap<String, ZuulRoute> locateRoutes() {
...
// Need to be done after static route code
String key = "/" + serviceToPathPolicy(serviceId) + "/**"
...
}
protected String serviceToPathPolicy(serviceId) {
   return this.policy(serviceId);
}
```
this.policy is a custom object that transform a service id in a path
rest-service-v1 -> v1/rest-service -> /v1/rest-service/**

I could provide a pull request if I'm not too wrong in the way to do this.

stephaneLeroy:
Add Zuul Proxy regex serviceId to route mapping
closes #693

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
@@ -26,6 +26,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.cloud.netflix.zuul.filters.ZuulProperties.ZuulRoute;
+import org.springframework.cloud.netflix.zuul.filters.regex.RegExServiceRouteMapper;
 import org.springframework.util.AntPathMatcher;
 import org.springframework.util.PathMatcher;
 import org.springframework.util.PatternMatchUtils;
@@ -55,6 +56,8 @@ public class ProxyRouteLocator implements RouteLocator {
 
 	private String servletPath;
 
+	private ServiceRouteMapper serviceRouteMapper;
+
 	public ProxyRouteLocator(String servletPath, DiscoveryClient discovery,
 			ZuulProperties properties) {
 		if (StringUtils.hasText(servletPath)) { // a servletPath is passed explicitly
@@ -76,8 +79,17 @@ public class ProxyRouteLocator implements RouteLocator {
 			}
 		}
 
+        if(properties.isRegexMapper()) {
+            this.serviceRouteMapper = new RegExServiceRouteMapper(properties.getRegexMap().getServicePattern(), properties.getRegexMap().getRoutePattern());
+        }
 		this.discovery = discovery;
 		this.properties = properties;
+	}
+
+	public ProxyRouteLocator(String servletPath, DiscoveryClient discovery,
+							 ZuulProperties properties, ServiceRouteMapper serviceRouteMapper) {
+		this(servletPath, discovery, properties);
+		this.serviceRouteMapper = serviceRouteMapper;
 	}
 
 	public void addRoute(String path, String location) {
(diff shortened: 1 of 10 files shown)


stephaneLeroy:
Don't know what goes wrong.
HystrixStreamTests test seems to hang (not a modified module in this PR).

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -57,6 +57,10 @@
 
 	private boolean ignoreLocalService = true;
 
+	private boolean isRegexMapper = false;
```
spencergibb: The pattern to follow should be `regexMap.enabled`.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -57,6 +57,10 @@
...
+
+	private RegexMapper regexMap = new RegexMapper("(?<name>.*)-(?<version>v.*$)", "${version}/${name}");
```
spencergibb: `s/regexMap/regexMapper`. Shortening it confused me, thought it was an actual java `Map` of something.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
```diff
@@ -238,6 +250,13 @@ public void resetRoutes() {
 		return values;
 	}
 
+	protected String mapRouteToService(String serviceId) {
+		if(this.serviceRouteMapper != null) {
```
spencergibb: You could use auto-configuration if regex mapper is enabled to create one, otherwise the default would be a passthru rather than an `if` statement here.

spencergibb:
Looking good otherwise.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRouteLocator.java
stephaneLeroy: Ok, nice feedback. 
I wasn't too confident about my first integration in ProxyRouteLocator constructor.
Just to be sure : 
By "the default" do you mean a SimpleServiceRouteMapper class that just return the serviceId?
Can I add ServiceRouteMapper parameter in ProxyRouteLocator constructor?

stephaneLeroy:
code reviews point corrected

spencergibb:
Thanks for the changes! I think the only thing missing from my perspective is some documentation.

stephaneLeroy:
Documentation added!

spencergibb:
Nice! I think a lot of people will like this. I should get to it tonight or, more likely tomorrow morning (GMT -7).

spencergibb:
Squashed and merged via 4720b78db556940c11909fdce56c11642548efb8.

The most [significant change](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java#L127-L145) was to how the `ServiceRouteMapper` was created. This allows user implementations to be injected.

stephaneLeroy:
Great! Thank you

darkedges:
Just to confirm that this will replace the need to use `PatternServiceRouteMapper` ?

spencergibb:
the regex mapper was renamed to `PatternServiceRouteMapper`

PR #713, opened 2015-12-11
Improve path cleaning by removing multiple successive slash

### LINKED ISSUE #583: Zuul: URISyntaxException when setting route.url with trailing slash
Using spring-cloud-starter-zuul 1.0.3.RELEASE, the problem occurs if you define a route like this (using spring-boot in this example):

```
zuul:
  routes:
    foos:
      path: /foos/**
      url: http://localhost:8080/
```
The trailing slash of the ```url``` always gets prepended to the delegation request. E.g.

```curl -X GET http://zuulhost/foos/1```

results in a ```GET //1``` to ```url``` (instead of ```GET /1```). This particular example does work though, a call to root doesn't however:

```curl -X GET http://zuulhost/foo/``` --> ```URISyntaxException``` as the request is ```GET //```

The current workaround of course is to define ```url``` without trailing slash, then everything works as expected:

```
zuul:
  routes:
    foos:
      path: /foos/**
      url: http://localhost:8080
```

tl;dr:
```ZuulRoute``` should take care of trailing slashes in ```url``` on delegation

kakawait:
Should fix #583

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -241,7 +241,7 @@ public class SimpleHostRoutingFilter extends ZuulFilter {
 				requestEntity);
 		URL host = RequestContext.getCurrentContext().getRouteHost();
 		HttpHost httpHost = getHttpHost(host);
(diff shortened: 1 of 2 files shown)


stephaneLeroy:
Sorry, I miss read your regex... I even hesitate to change my own PR

kakawait:
If you missed read my regex that proove the regex was confusing. `/{2,}` is maybe more clear and natural

spencergibb:
Squashed and merged via b108de33ed677ccdc6aa07f2a16234dc1c7165b5

Thanks @kakawait!

PR #723, opened 2015-12-16
Update docs to indicate the need for spring-cloud-starter-atlas

jkschneider:
`spring-cloud-starter-atlas` adds the Jackson Smile library necessary to properly format the post to Atlas.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -1517,7 +1517,7 @@ Atlas was developed by Netflix to manage dimensional time series data for near r
 
 Atlas captures operational intelligence. Whereas business intelligence is data gathered for analyzing trends over time, operational intelligence provides a picture of what is currently happening within a system.
 
-No additional dependencies are necessary to send Spring Boot Actuator, Servo, and Spectator metrics to Atlas. Just annotate your Spring Boot application with `@EnableAtlas` and provide a location for your running Atlas server with the `netflix.atlas.uri` property.
+Spring Cloud provides a `spring-cloud-starter-atlas` that has all the dependencies you need. Then just annotate your Spring Boot application with `@EnableAtlas` and provide a location for your running Atlas server with the `netflix.atlas.uri` property.
 
 ==== Global tags
 


anthonyikeda:
Thanks for that Jon, I'll give that a run in the morning!

jkschneider:
@anthonyikeda working on a spring-cloud-sample right now to demonstrate, I'll share with you when done

anthonyikeda:
Let me update my gist with the gradle config and the application aspects to
show you what I have (event adding the atlas dependency I'm getting the
NullPointer, I addd smile separately originally)

anthonyikeda:
I've added in my bootstrap.yml, build.gradle (including the newly added
spring-cloud-starter-atlas), a manager implementation showing an example of
a Timer, and the main Application enabling Atlas. I recompiled but the unit
test I have on the DepartmentManagerImpl is still throwing the same
NullPointerException. I also have aloha instance of Atlas running.

Maybe I'm doing something wrong?

anthonyikeda:
https://gist.github.com/anthonyikeda/0f01df1045c9b231044b

anthonyikeda:
The application log is now added also for clarity.

anthonyikeda:
The Spock spec on the DepartmentManagerImpl fails when testing the
getDepartment(String id) method that uses the Timer.

When trying to skip tests and running the application, there seems to be a
wiring issue with the Spectator Metrics Services:

NoUniqueBeanDefinitionException: No qualifying bean of type
[org.springframework.boot.actuate.metrics.GaugeService] is defined:
expected single matching bean but found 2:
gaugeService,spectatorMetricServices

jkschneider:
Thanks, clearly a few different things to look at.  If I remember correctly, `@AutoConfigureBefore(EndpointAutoConfiguration.class)` was supposed to pre-empt the actuator GaugeService, but perhaps there is some other ordering issue here.

jkschneider:
I don't see the missing netflix.atlas.uri problem in my sample app, so there is something else going on here to find.

jkschneider:
@anthonyikeda this is what I'm working off of currently: https://github.com/jkschneider/spring-cloud-atlas-sample

anthonyikeda:
I'm going to try building off the samples project rather than apply to an
existing project to get a better understanding as I still can't get this to
work in my own project.

Anthony

anthonyikeda:
I think I'm making progress, in my examples I'm annotating the main class
with @SpringCloudApplication, not @SpringBootApplication. Swapping that
over has so far removed any errors when running the app as bootRun - the
Unit Test still report the NullPointerException.

Mind you my Spock Specs are written in Groovy - not sure if that would be a
problem.

When Running with @SpringCloudApplication it looks like there is some
wiring of the GaugeService already performed - thus the conflict:

Error creating bean with name
'org.springframework.cloud.netflix.hystrix.HystrixCircuitBreakerConfiguration$HystrixMetricsPollerConfiguration':
Injection of autowired dependencies failed; nested exception is
org.springframework.beans.factory.BeanCreationException: Could not autowire
field: private org.springframework.boot.actuate.metrics.GaugeService
org.springframework.cloud.netflix.hystrix.HystrixCircuitBreakerConfiguration$HystrixMetricsPollerConfiguration.gauges;
nested exception is
org.springframework.beans.factory.NoUniqueBeanDefinitionException: No
qualifying bean of type
[org.springframework.boot.actuate.metrics.GaugeService] is defined:
expected single matching bean but found 2:
gaugeService,spectatorMetricServices


Do you want me to create a new defect ticket?

Anthony

anthonyikeda:
Yup, @SpringCloudApplication adds in the @EnableHystrix annotation as part
of it's declaration. If I use @SpringBootApplication and @EnableHystrix I
get the same error.

jkschneider:
@anthonyikeda thanks for hunting down the reason -- fix is in #735

Issue #579, opened 2015-10-08
Split zuul, ribbon and feign out of core


spencergibb:
So s-c-zk and s-c-consul can have cleaner dependencies.

spencergibb:
I'm thinking moving eureka-client out of core is simpler and easier.

PR #739, opened 2015-12-23
Move eureka-client specific code to separate module.

spencergibb:
fixes gh-579

Rather than move ribbon, feign, zuul etc.. out of core I decided to move eureka-client out.


spencergibb:
Currently compiles, haven't run tests yet.

spencergibb:
I think I'm done with this. The brewery tests pass on my local machine. Not sure why things hang on integration tests.

@dsyer @adriancole @marcingrzejszczak

adriancole:
there are errors in the console output like this. Are they unrelated? Still curious how to see what's hung that results in 10m of no-output. Maybe there's a travis tool or methodology to capture details when a container is killed?

```
Caused by: java.io.FileNotFoundException: class path resource [org/springframework/cloud/netflix/test/NoopDiscoveryClientConfiguration.class] cannot be opened because it does not exist
```

adriancole:
We can also try commenting out the trusty beta container travis config. no reason to suspect it is involved, but little effort to try.

marcingrzejszczak:
I've sent an email to Travis support with regards to this case - I'll ping you once I get some answer.

marcingrzejszczak:
Actually - https://travis-ci.org/spring-cloud/spring-cloud-netflix the build has gone through but it just failed

PR #743, opened 2015-12-27
Abstraction for ZuulRoute storage

jmnarloch:
Hi, I have a additional pull request this time to Zuul proxy integration part. 

I would like to change two thing in the current implementation, first is loading the routes from external storage - that is for large organizations centralized text file (like yaml) is not really manageable, so I would like to introduce one more layer for the ProxyRouteLocator to be able to load the routes definitions from different sources like: RDMBS or NoSQL - like Cassandra for instance.

The second thing I would like to do is to introduce some strategy for route matching with possible extension points for more specialized ones - I will try to prepare a separate PR for that. 

For now I don't want to introduce any specific implementation except for the existing one, just introduce possible extension points.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulConfiguration.java
@@ -20,14 +20,17 @@ import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.web.ErrorController;
 import org.springframework.boot.context.embedded.ServletRegistrationBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.client.actuator.HasFeatures;
 import org.springframework.cloud.context.scope.refresh.RefreshScopeRefreshedEvent;
+import org.springframework.cloud.netflix.zuul.filters.PropertiesZuulRouteStore;
 import org.springframework.cloud.netflix.zuul.filters.RouteLocator;
 import org.springframework.cloud.netflix.zuul.filters.SimpleRouteLocator;
 import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
+import org.springframework.cloud.netflix.zuul.filters.ZuulRouteStore;
 import org.springframework.cloud.netflix.zuul.filters.post.SendErrorFilter;
 import org.springframework.cloud.netflix.zuul.filters.post.SendResponseFilter;
 import org.springframework.cloud.netflix.zuul.filters.pre.DebugFilter;
@@ -57,6 +60,9 @@ public class ZuulConfiguration {
 	@Autowired
 	private ZuulProperties zuulProperties;
 
+	@Autowired
+	private ZuulRouteStore routeStore;
+
 	@Autowired(required = false)
 	private ErrorController errorController;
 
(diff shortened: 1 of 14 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulRouteStore.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author Jakub Narloch
+ */
+public interface ZuulRouteStore {
+
+    /**
+     * Retrieves the list of {@link org.springframework.cloud.netflix.zuul.filters.ZuulProperties.ZuulRoute}.
+     *
+     * @return list of routes
+     */
+    Map<String, ZuulProperties.ZuulRoute> getRoutes();
```
spencergibb: `ZuulRoute` should move out of `ZuulProperties` to here.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulConfiguration.java
```diff
@@ -140,6 +146,16 @@ public ZuulFilterInitializer zuulFilterInitializer() {
 
 	}
 
+	@Configuration
+	public static class ZuulRouteStoreConfiguration {
```
spencergibb: Similar to other PR, unneeded static class.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulRouteStore.java
jmnarloch: Alright, but don't you think it maybe would make more sense to make ZuulRoute top level class? So that we could still use it ZuulProperties without need to use there ZuulRouteStore.ZuulRoute indirection.

spencergibb: Makes sense.

jmnarloch:
I've done the changes. 

Whichever PR you will decide to merge first I will have to either way merge the other one.

dsyer:
Thanks for the code. But why would the Spring Environment not be sufficient? We are trying to keep configuration in one place, and that seems like the right place to me, since it is used for everything. A route matching strategy is orthogonal, so maybe you could separate that out into a separate PR?

jmnarloch:
Hi Dave,

So we came up with conclusion that for organization having more then 50 services exposed through edge service it becomes hard to maintain and manage those in single file, so this tries to externalize loading those from configurable storage. Now this pull request doesn't do more then provide an extension point for loading the route definitions, and keeps the standard/default implementation that uses environment for that purpose.  Prior this pull request the only way to change this behaviour was to define everything from custom @EnableZuulProxy that would import the subclassed configuration and finally the altered ProxyRouteLocator, so having that extension point would be a plus from this perspective.

The abstraction for the matcher is in separate pull request: https://github.com/spring-cloud/spring-cloud-netflix/pull/763

dsyer:
I get why you would want to do this. I'm just quibbling with the implementation. If you want to use a database to store routes why not just write an `EnvironmentPostProcessor` for it?

jmnarloch:
I haven't though about the EnvironmentPostProcessor, though it looks a bit hacky, couldn't the ProxyRouteLocator open parts of it's API and define some extension points in similar manner as the Feign does? https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java

jmnarloch:
Closing the issue. Thanks for the effort to refactor the ProxyRouteLocator, I've taken a look on the changes and generally like the fact the in current version the ProxyRouteLocator can be replaced with custom implementation.

PR #735, opened 2015-12-18
Deconflict GaugeService provided by Servo/Spectator integration

jkschneider:
Addresses the issue brought up by @anthonyikeda on #723.

Latest commit to the sample project demonstrates the issue and fix: https://github.com/spring-cloud-samples/spring-cloud-atlas-sample/commit/a00908544cc37e2e3a320c3d496571106e09c583

PR #745, opened 2015-12-28
Servo and Spectator metrics defer to other Counter/GaugeServices

jkschneider:
@otrosien Reported this issue as a comment on #735.  While it has a similar stacktrace, the conflict here is different: in this case between `DropwizardMetricServices` and `ServoMetricServices`.  Since Servo support is being added automatically in spring-cloud-netflix and a developer may not be interested in using this feature, I think it makes sense to defer to other (intentionally added) metrics implementations like Dropwizard when both are present.


yairogen:
in which version can I test this?

yairogen:
@spencergibb is there a version out there I can test this against? Another milestone maybe?

spencergibb:
SNAPSHOT's are the only thing right now. I suspect we'll have another milestone or RC soonish (Jan).

PR #763, opened 2016-01-10
Extracted ProxyRouteLocator matching logic

jmnarloch:
Hi, [as promised](https://github.com/spring-cloud/spring-cloud-netflix/pull/743) I've made another pull request to Zuul.

This time I would to introduce thin abstraction layer so that there would be an extension point in Zuul route matching logic that could be replaced by more specialized versions.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
@@ -31,11 +31,13 @@ import org.springframework.cloud.client.discovery.event.InstanceRegisteredEvent;
 import org.springframework.cloud.client.discovery.event.ParentHeartbeatEvent;
 import org.springframework.cloud.context.scope.refresh.RefreshScopeRefreshedEvent;
 import org.springframework.cloud.netflix.ribbon.SpringClientFactory;
+import org.springframework.cloud.netflix.zuul.filters.AntPathZuulRouteMatcher;
 import org.springframework.cloud.netflix.zuul.filters.ProxyRequestHelper;
 import org.springframework.cloud.netflix.zuul.filters.ProxyRouteLocator;
 import org.springframework.cloud.netflix.zuul.filters.ServiceRouteMapper;
 import org.springframework.cloud.netflix.zuul.filters.SimpleServiceRouteMapper;
 import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
+import org.springframework.cloud.netflix.zuul.filters.ZuulRouteMatcher;
 import org.springframework.cloud.netflix.zuul.filters.pre.PreDecorationFilter;
 import org.springframework.cloud.netflix.zuul.filters.regex.RegExServiceRouteMapper;
 import org.springframework.cloud.netflix.zuul.filters.route.RestClientRibbonCommandFactory;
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
```diff
@@ -136,6 +141,16 @@ public ServiceRouteMapper serviceRouteMapper(ZuulProperties props) {
 	}
 
 	@Configuration
+	protected static class AntPathZuulRouteMatcherConfiguration {
```
spencergibb: Why a static configuration class?

jmnarloch: Probably without a really good reason - removed the wrapping class.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/AntPathZuulRouteMatcher.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * A base implementation of {@link ZuulRouteMatcher} that uses {@link AntPathMatcher} for matching the request URI
+ * with configured routes.
+ *
+ * @author Jakub Narloch
+ */
+@CommonsLog
+public class AntPathZuulRouteMatcher implements ZuulRouteMatcher {
```
spencergibb: Is a whole new abstraction needed? Could the functionality move to a protected method in `ProxyRouteLocator`?

jmnarloch: Yes, that's the whole point of this pull request, I would like to be able to introduce the configuration strategy for the path matching, at this point I am unable to do that without defining custom @EnableZuulProxy annotation: https://github.com/jmnarloch/zuul-trie-matcher-spring-cloud-starter/blob/master/src/main/java/io/jmnarloch/spring/cloud/zuul/support/ZuulProxyMatcherConfiguration.java

It would be much simpler I could be able to leave ProxyRouteLocator as it is with swapping only the parts of it's logic: so either the storage from where I read the routes or the matching approach.

I am aware that this makes ProxyRouteLocator a bit more complicated, but on other hand being able to extract that nicely separates the concerns.

dsyer:
I don't like this idea much either. It seems like `RouteLocator` should be the strategy. I'll have a look at beefing it up and using it more consistently.

jmnarloch:
Well if will just take a look at Zuul as a example itself it has this nice abstraction of between the general load balancing class in form of ILoadBalancer through ServerList - the server list provider and finally ServerListFilter and IRule the strategy for matching the servers. I would be happy to have at least some extension points in RouteLocator (or ProxyRouteLocator) that would mimics such separation of concerns.

Though I don't insist on one solution over another, simply at this point there is no way to plug there any alternate behaviour except to replace everything.

dsyer:
I think you mean Ribbon (the two abstractions for server list and rule)? I get that, but I don't understand why you need it here. You already have many choices (there are 2 `RouteLocator` implementations, and the routes can be configured through the Spring `Environment`, and all the extension points that offers, or via discovery). There is also a matching abstraction for servers (`ServiceRouteMapper`) - I don't like it much so I'm going to probably change it or remove it. Keep your eye on master and see if it is heading in a useful direction for you.

spencergibb:
@jmnarloch 9161c54c43bf615dd95f17691cb8203cf27c5bba has changed zuul significantly.  Take a look and see if it suits your needs.

jmnarloch:
Closing the issue.

PR #775, opened 2016-01-18
RxJava handling multiple Observable results

jmnarloch:
PR that handles multiple values returned by the Observable. 

I've seen lately remark that in order to use the current implementation you need to use Observable&lt;List&lt;T&gt;&gt;, which is quite counter intuitive since the Observable by itself can produce multiple results, so this PR tries to address this issue, by aggregating the results and postponing the actual moment when they are going to "applied" on the DeferredResult.

The only thing I was thinking of is whether arbitrary wrapping the values into a plain Object array is a good way to approach this.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/rx/ObservableReturnValueHandler.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/rx/ObservableReturnValueHandler.java
@@ -23,7 +23,11 @@ import org.springframework.web.context.request.async.WebAsyncUtils;
 import org.springframework.web.method.support.AsyncHandlerMethodReturnValueHandler;
 import org.springframework.web.method.support.ModelAndViewContainer;
 import rx.Observable;
+import rx.functions.Action0;
 import rx.functions.Action1;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Handles return values of type {@link rx.Observable}.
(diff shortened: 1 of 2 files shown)


rstoyanchev:
Indeed this code could use some improvement. `Observable` produces 0..N items while `DeferredResult` accepts a single result. That's a better fit for `rx.Single`. For an Observable, you can accumulate the items before rendering the response as in this PR, possibly even using `Observable.toList()` in the implementation.

The response of course would not be rendered until all items are accumulated. Alternatively for a more streaming scenario where each item is rendered with a message converter, something similar to the [ResponseBodyEmitterReturnValueHandler](https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/ResponseBodyEmitterReturnValueHandler.java) could be done.

spencergibb:
@rstoyanchev Interesting, so support `rx.Single` similar to the current `Observable` support and move `Observable` support to something like `ResponseBodyEmitterReturnValueHandler`.

rstoyanchev:
That would be more idiomatic I think. Otherwise with an `Observable` return type there is no way of telling whether the intent is to accumulate + render or stream (i.e. render each item).

sdeleuze:
:+1:

jmnarloch:
> Interesting, so support rx.Single similar to the current Observable support
and move Observable support to something like
ResponseBodyEmitterReturnValueHandler.

I can try to prepare a pull request later for that :)

2016-01-19 10:03 GMT+01:00 Sébastien Deleuze <notifications@github.com>:

spencergibb:
@jmnarloch that would be great! I think we should close this one in lieu of the new one.

jmnarloch:
I've made the PR, it's look promising: https://github.com/spring-cloud/spring-cloud-netflix/pull/778
Though maybe some extra tests would be useful.

rstoyanchev:
@jmnarloch, @spencergibb, after some giving this full consideration for including in Spring Framework 4.3. we've [reached a conclusion](https://jira.spring.io/browse/SPR-14046) that hopefully will be helpful to you to move forward. 

We've decided not to include built-in support in 4.3 and leave that to 5.0. While we could have easily and cleanly added support for `Single` and `Observable` for SSE scenarios, there are still questions surrounding the use of `Observable` more generally that can only be addressed fundamentally in [spring-reactive](ub.com/spring-projects/spring-reactive) and Spring Framework 5.

I did however make some changes that should make doing that for your internal needs easier. See [this pull request](https://github.com/spring-projects/spring-framework/pull/1005). Essentially just register a DeferredResultHandler with an rx.Single adapter and a ResponseBodyEmitterResultHandler with an rx.Observable adapter which creates an SseEmitter if "Content-Type:text/event-stream".

I should clarify I didn't think it necessary to expose timeout customizations because Observable and Single have built-in timeout related capabilities that are probably sufficient. For example in your adapter code you can create DeferredResult and SseEmitter with some expanded timeout (say from 10 to 30 seconds) and combine that with the timeout operator to supply default values with some fluctuation per use case.

For indicating SSE, it would be enough to return `ResponseEntity.ok().contentType(new MediaType("text/event-stream").body(observable)` or some shortcut resulting in the same.

As to how to treat an `Observable` return value, it's really your call. You can make more assumptions than we can but it's also easy enough in code to do `observable.toList().toSingle()` so it might be enough to support just `Single` and `Observable` for SSE.

Hope this helps.

spencergibb:
@rstoyanchev any reason you reopened this one instead of #778?

rstoyanchev:
No, no reason. I think this was linked from the PR somewhere.

spencergibb:
Thanks for your work! Will integrate the new changes with boot 1.4 and spring 4.3 and look forward to removing this with spring 5!

rstoyanchev:
No worries, this was overdue to consider with the 5.0 work progressing.

PR #778, opened 2016-01-19
RxJava streaming Observable

jmnarloch:
This PR alters the ObservableReturnValueHandler behavior, so that it would be possible to return multiple values. On transport layer it uses HTTP Streaming, so that multiple values returned from the Observable will be emitted to the client.

Beside that an additional support for rx.Single has been added that resembles Spring's DeferredResult or ListenableFuture.


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/rx/ObservableReturnValueHandler.java
```diff
@@ -17,55 +17,129 @@
...
+			}
+		}
+
+		@Override
+		public void onError(Throwable throwable) {
+			deferredResult.setErrorResult(throwable);
+		}
+
+		private <T> void send(T next) throws IOException {
+			final MediaType mediaType = response.getHeaders().getContentType();
```
jmnarloch: This is the only piece that might be a complication, the original ResponseBodyEmitter have information about the object media type, at this point I don't have anything similar from Observable.

jmnarloch:
I took some time to work on this, remembering what we had discuss in the previous pull request. I very like the idea of implementing customized ResponseBodyEmitter. Yet I've came to conclusion that this might provide support for rather narrow use case.

So I started from scratch from a bit different perspective, trying to introduce the general purpose API that could be used in a bit wider scenarios. I've ended in defining three basic types:

* SingleDeferredResult
* ObservableDeferredResult
* ObservableSseEmitter

The first two are specialized versions of DeferredResult - Single will always produce single entity. In case of Observable the values are going to be accumulated into a List and then return (still this is async operation).

The ObservableSseEmitter is quite different - this is specialized version of Spring's SseEmitter that emits everything produced from Observable as a server side events. I found the SseEmitter (the only specialized version of RepsonseBodyEmitter the most useful, since it produced the response in structured and standardize manner) that could be conveniently consumed by the clients. Using entirely custom RepsonseBodyEmitter would require to write fully customized client. So I found that not so much useful, still we could add ObservableResponseBodyEmitter at some point in the future.

I think that this approach in general is a bit more flexible, the user can choice one of the above wrapper classes depending on their needs. 

Finally on top of that I've implemented the AsyncHandlerMethodReturnValueHandler for the Observable and Single that behaves exactly the same as *DeferredResult counterparts. So this will provide the "default" behavior whenever Observable or Single would returned from handler method.

jmnarloch:
Still there are some edge cases.
Consider that the DeferredResult let you set as the value ResponseEntity in the result Single<ResponseEntity> would behave correctly, applying to the HTTP response the headers and the status code.

Unfortunately the same does not apply to Observable<ResponseEntity> since those are going to wrapped into a list. As a workaround we could implement the ObservableReturnValueHandler in the way that it would allow to wrap the Observable into the ResponseEntity (like ResponseEntity<Observable<?>>) and then whenever the Observable will complete producing result wrap the output into artificial ResponseEntity with status and headers populated from the original one.

Still this we could end up with confusing structures like Single<ResponseEntity<?>> vs. ResponseEntity<Observable<?>>. So I'm not sure whether that would be a good and clear solution.

I've haven't  yet implemented this behavior, this is something worth considering.

jmnarloch:
CC @spencergibb @dsyer @sdeleuze @rstoyanchev

spencergibb:
@jmnarloch `ResponseEntity<Observable<?>>` seems more natural to me. @dsyer @sdeleuze @rstoyanchev thoughts?

sdeleuze:
Indeed I also think that `ResponseEntity<Observable<?>>` is what we have to support, not the other way around.

About the PR, I am fine with the principle of `SingleDeferredResult` and `ObservableSseEmitter`, but it seems to me than `ObservableDeferredResult` is a different beast because `Observable.toList()` is blocking, and since `Observable` is an async stream of values, users may be quite surprised about how such type is handled by the framework.

Spring Framework 5 will be designed to handle that kind of return value as users could expect it, but for now with Spring Framework 4.x and since RxJava now provides a `Single` type, it seems to me that we have a choice to make between (at least) these 2 alternatives:
 1) Only support `Single` for non SSE use cases, and let users write `observable.toList()` themselves to fully understand the limitation of such operation
 2) Keep `ObservableDeferredResult`  but modify Javadoc and reference documentation with a clear warning  `Observable` return values handling limitation for non SSE use cases

I tend to think than the first solution is less misleading for the user, not complicated (write `observable.toList() is easy and there is also `observable.toSingle()` available) and will make Spring Framework 5 benefits more understandable. Any thoughts?

jmnarloch:
Thanks your comments.

Although I somehow disagree that toList() is blocking operation, to my knowledge and correct me if I am wrong, it isn't a blocking operation (in meaning that it won't block the current thread), though I agree that it won't produce any result unless the whole Observable completes processing. 
Please see: https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorToObservableList.java 

Regarding whether to make the *DeferredResult public or not it's up to you, either way I would keep those, since their nicely encapsulate the subscription to the Observable/Single including the OnSubscribe contract.

Since I have the previous pull request in the branch. If we agree on:
- SingleDefferedResult 
- ObservableDeferredResult

It's not much effort from my side to change the handling for ObservableReturnValueHandler, to produce streaming results.

sdeleuze:
My mistake, I thought `Observable#toList()` was returning a `List<T>`, while it returns an `Observable<List<T>` so you are right indeed.

So my proposal would be to just add a warning in the documentation to warn users dealing with big (or infinite stream), something similar to https://github.com/ReactiveX/RxJava/blob/1.x/src/main/java/rx/internal/operators/OperatorToObservableList.java#L36.

sdeleuze:
Could you please detail what you have in mind with:
> It's not much effort from my side to change the handling for ObservableReturnValueHandler, to produce streaming results.

smaldini:
It seems to me that SingleDeferredResult will need at least a subscribeOn/observeOn or a specific asynchronous handling in a deferred task since its requesting an unbounded amount of data without releasing it. These operators in Rx are usually meant to be used in coordination with splitting, filtering and other reducing operators such as window, groupBy, take etc. I'm not sure what would buy a single handling for collection anyway where the emphasis of Rx or even Reactive Streams is around streaming whereas buffering is encouraged in asynchronous boundaries not as a way to efficiently transform an unbounded sequence into a collection. For this there is toBlockingIterable which will be more efficient.

jmnarloch:
@smaldini Agreed. This why we are introducing ObservableSseEmitter for that purpose, though we want to have still a "native" support for rx.Single and rx.Observable returned from mapped handlers. I came to conclusion that except to rx.Single which is nothing more than the Spring's ListenableFuture / DeferredResult handling Observable in initially proposed streaming manner would end up in very specific implementation with possible narrow use case. What more it might be hard to backup from it. 

Having handler method defined as fallows:
```
@ResponseBody Observable<Entity> entities() {
  ... 
}
```
have some significant limitations. For instance, I wasn't able to conclude how to acquire the MediaType when writing each individual entity? How to define API for user to specify separator for the entries?

Let's say that we would marshall produces POJOs prom Observable into JSON resulting in:
{"attribute1": "value1"}{"attribute2": "value2"}...
How to parse such stream? We could deliberately introduce separator, let's say a new line, that would definitely work for JSON (unless someone decides to enable pretty printing), but what about the other formats?

So in practice this simplified API produces a lot practical problems and considerations. This is way the specialized classes has been introduced as mentioned already:

* SingleDeferredResult - can handle scalar results as well as collections optionally wrapped  in ResponseEntity
* ObservableDeferredResult - Observable can produce multiple results, we want to handle this by aggregating the results, so this will resembles returning collections of objects in Spring MVC.
* ObservableSseEmitter - this is a true streaming solution with standardized contract that can be understand by the clients.

In neither of this cases I would consider using blocking operation, though.

rstoyanchev:
I'm missing something basic and I can't find it in the discussion above (or may be I missed it). 

When I look at examples of controller methods that return Single or Observable, or even ResponseEntity<Observable> (although I see no examples of that) it makes sense instantly. Seeing SingleDeferredResult, or ObservableDeferredResult, or ObservableSseEmitter on the other hand I can't make sense of. Why would I return Single inside a DeferredResult from the controller method?

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/rx/handler/ObservableReturnValueHandlerTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Value("${local.server.port}")
+    private int port = 0;
+
+    private TestRestTemplate restTemplate = new TestRestTemplate();
+
+    @Configuration
+    @EnableAutoConfiguration
+    @RestController
+    protected static class Application {
```
jmnarloch: @rstoyanchev The tests are here.

jmnarloch: @rstoyanchev The tests are here.

jmnarloch:
I've added those classes, since I though that they would add the user the extra freedom, so that he could define for instance the operation timeout per operation or optionally specify the timeout value similary as for the plain DeferredResult.

We can made them package visible, if that's an issue.

jmnarloch:
@spencergibb I've added the javadocs, extended the ObservableReturnValueHandler to support syntax like: ResponseEntity<Observable<?>> that is going to copy the HttpStatus and HttpHeaders from the handler method return value.

rstoyanchev:
@jmnarloch I see now, thanks. In that case there probably shouldn't be a constructor such as `ObservableDeferredResult(Observable<T> observable)` since the only time you want to do this is if you want to customize the timeout. 

Still those type names are pretty long and a bit jarring. Is it actually necessary to declare the controller method like this:
````java
public ObservableDeferredResult<String> single() {
````
or could it be just:
````java
public DeferredResult<String> single() {
````
Also instead of exposing the actual type and constructor:
````java
public ObservableDeferredResult<String> single() {
    // ....
    return new ObservableDeferredResult<>(1000L, myObservable);
}
````
only expose a util method:
````java
public DeferredResult<String> single() {
    // ...
    return DeferredUtils.timeout(100L, myObservable);
````

Another line of thought is that the default timeout (for all requests) is configurable centrally through the MVC config. There is also a timeout operator in Rx which I do see in one of the tests. Between those two there might be enough to deal with timeouts. So you could consider adding less now until there is a clear case that more is needed.

**Comment by smaldini** [comment]
I understand @jmnarloch :+1: I am not personally fan of DeferredResult as it somehow encourage this. But you would need something along the lines of https://github.com/spring-projects/spring-reactive/blob/master/src/main/java/org/springframework/web/reactive/method/annotation/ResponseBodyResultHandler.java if you wanted to stream (+ the bridge to servlet 3.1 or netty via reactor-netty and rxnetty). I'm also very interested in your PR because I could somehow reproduce this support for reactor somehow in particular the rx lite api (http://next.projectreactor.io/core/docs/api/reactor/core/publisher/Flux.html) and (http://next.projectreactor.io/core/docs/api/reactor/core/publisher/Mono.html) in progress.

**Comment by sdeleuze** [comment]
I also think that for the users, dealing just with `Single<T>`, `Observable<T>`, `ResponseEntity<Observable<T>>` and `ResponseEntity<Single<T>>` handler method return values is enough for now (+ `ObservableSseEmitter` for SSE use case).

**Comment by jmnarloch** [comment]
Hi again, sorry for a bit of delay. Thanks for all of thoughts and opinions.
I did fallow them @rstoyanchev @smaldini @sdeleuze and as result came up with the fallowing changes:

* Made all of the SseEmitter/DeferredResult package visible, so we do not expose them to the client.
* Introduced a utility class RxResponse (example usage RxResponse.sse(observable) - if you can suggest a better name I will be happy to change that)
* Finally I trimmed the scope of the PR, removing the ObservableDeferredResult/SingleDeferredResult.

At this point the supported use cases are:

* handler returning rx.Single
* handler returning rx.Observable
* explicitly creating SseEmitter through RxResponse.sse(observable)

I would like to wrap up this PR and optionally made the final polishing. @spencergibb any additional thoughts?

File: spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/rx/RxResponse.java [hunk 3]
```diff
@@ -0,0 +1,58 @@
+package org.springframework.cloud.netflix.rx;
+
+import org.springframework.http.MediaType;
+import org.springframework.web.servlet.mvc.method.annotation.SseEmitter;
+import rx.Observable;
+import rx.Single;
+
+/**
+ * A convenient class allowing to wrap either the {@link Single} or {@link Observable} into a response supported by the
```
spencergibb: I assume this is a copy/paste error.

spencergibb:
@jmnarloch the only thing would be docs, but that can come in another PR.

@rstoyanchev @smaldini @sdeleuze any objections to me merging?

sdeleuze:
I think I am ok with what is proposed for `rx.Single` and `rx.Observable` handler support but I need to have a deeper look about the SSE support especially regarding `RxResponse` and what we are going to support on Spring Reactive for that use case. I will send you a detailed feedback monday or tuesday.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/rx/RxResponse.java
jmnarloch: Yes, thanks, forgot to update the docs on the class after removing the methods.

spencergibb:
Ping @sdeleuze. Hoping to merge this by the end of the week.

jmnarloch:
@spencergibb I can prepare the PR for the documentation once we merge this pull request, plus I can write a blog post with some additional examples.

spencergibb:
nice.

sdeleuze:
Sorry I was not able to spend much time on this before my vacations, but please find bellow my feedback sent from the airport ^^.
 
While it is a good idea to have class like `SingleDeferredResult` package private (since the principle is to return directly `Single` and `Observable`), I am not sure we need this `RxResponse.sse()` thing. Maybe making `ObservableSseEmitter` `public` will be fine and let people write `return new ObservableSseEmitter(...)` is fine.

Please notice that as part of [Spring Reactive](https://github.com/spring-projects/spring-reactive) project and the upcoming Spring Framework 5, I am thinking about a way (annotation?) to enable SSE support that could allow returning directly `Observable` instead of an `SseEmitter` but for now, I think returning an `ObservableSseEmitter` is the best we can provide.

I think the documentation and the blog post should mention the limitation of handler returning `Observable` (regarding memory consumption and latency with big stream of data, and say explicitly that infinite stream are not supported for this use case). That could be nice to explain that also in the blog post, and give some hints to the user about the full Reactive support that is coming in Spring Framework 5 (see with @rstoyanchev and @smaldini for that) in order to explain people why a fully Reactive Spring Framework 5 is need even if a preliminary support for RxJava based on Spring Framework 4.x fondations is provided here.

My last comment is that RxJava1 and RxJava2 will have different packages. That's why in Spring Reactive and Reactor we use and RxJava1 prefix in our class. Maybe you could do the same to anticipate RxJava2 support.

I will not be available since I am on vacation, but it would be nice to check these points with @rstoyanchev before merging and publishing the blog post.

spencergibb:
Thanks @sdeleuze! Hope you have a nice vacation.

spencergibb:
Ping @rstoyanchev

dsyer:
Somebody asked me about this today. I looked at the discussion here and judged it "healthy", so I assumed we would be merging, and possibly porting parts of it to Spring 5?

spencergibb:
Yup, I was waiting to hear back from rossen before I merged. My guess is this would be ported or obseleted. It's nice especially if you have a feign client with Hystrix that returns Observable.

smaldini:
From my perspective it's quite cool stuff ! Even with plain Reactor Core 2.5 support we are talking Observable/Single/Completable out of the box, no extra bridge required, and this is leveraged by Spring Framework 5 Converters since RxJava is a first-class supported Reactive composition library. So I would expect a kind of continuity especially with the NetflixOSS ecosystem depending on it and use cases like Feign.

Maybe a cool next step @jmnarloch after this is getting validated by the crew would be to come have a look at https://github.com/spring-projects/spring-reactive to fill issues and why not joining the effort !

sdeleuze:
Hey guys I am back, I will check with @rstoyanchev that everything is ok and send you a feedback hopefully today.

rstoyanchev:
The more I think about it, and sorry for coming to this realization rather slowly, the more I realize we shouldn't do this here. It went from a simple Observable adapter originally, which was rather okay IMO, to a rather involved Spring MVC / RxJava integration. Lots of things changed in the mean time, like we got Single in RxJava, and we got ResponseBodyEmitter in Spring MVC. 

In any case my point is that I see the case or doing this property in the Spring Framework where we can property support Reactor Flux and Mono as well as RxJava. We might even pull some of the work already from spring-reactive into the Spring Framework master branch such as [these converters](https://github.com/spring-projects/spring-reactive/tree/master/src/main/java/org/springframework/core/convert/support) so that supporting additional reactive libraries becomes a matter of adding another Spring `Converter`.

Beyond that I think `rx.Single` translates to `DeferredResult` and this is correctly represented here, while `rx.Observable` translates to `ResponseBodyEmitter`, not to `DeferredResult` via `Observable.toList()`. 

Assuming we agree on the general point of where this should be solved the next step would creating a ticket in jira.springsource.org/browse/SPR.

spencergibb:
If no one disagree's I'm going to close this in favor of https://jira.spring.io/browse/SPR-14046.

spencergibb:
Reopening as https://jira.spring.io/browse/SPR-14046 was marked as `Won't Fix`. I'll limit this to the `Single` support and `Observable` sse support.

spencergibb:
Spring 5 won't be available until early 2017, likely 'D' spring cloud release or later. Will remove this support when spring cloud supports spring 5.

spencergibb:
See @rstoyanchev's comments here https://github.com/spring-cloud/spring-cloud-netflix/pull/775#issuecomment-196429102

spencergibb:
Closed with some changes via 2bca3504cd103a09a1d0e3ee46d3d8cdbfcf15d2. Most notably `ObservableReturnValueHandler` was removed in favor of `SingleReturnValueHandler` (though the `ResponseEntity` handling was ported), a little polish and some docs.

Thanks @jmnarloch, @rstoyanchev @smaldini & @sdeleuze for all the work and conversation.

PR #784, opened 2016-01-21
Fix Route Locator for both ZuulServlet and DispatcherServlet scenarios

aivans:
Both execution realms must work.

I created a ServletDetectionFilter for early detection for the servlet realm.
Route Locator will use this information for finding the route at runtime.

PR #794, opened 2016-01-26
(Second attempt)Fix Route Locator for both ZuulServlet and DispatcherServlet scenarios

aivans:
Conversation history here: https://github.com/spring-cloud/spring-cloud-netflix/pull/784


aivans:
Created a cleaner pull request without merges, etc

dsyer:
I still get a test failure in `ProxyRequestHelperTests`. It only fails for the whole suite on the command line so I don't think it's related to your change, but we need to fix it.

BTW, can you use UNIX line endings and tabs for whitespace please?

dsyer:
Merged with squash and some whitespace changes.

PR #795, opened 2016-01-26
Add the ability to configure the max connections and max per route on PoolingHttpClientConnectionManager with different values

AndreasKl:
While trying to optimize our throughput we set `zuul.max.host.connections` for SimpleHostRoutingFilter this had the side-effect of increasing the `connectionManager.setDefaultMaxPerRoute` as well.

The PR enables a user to configure both in an independent way by introducing a new config setting without breaking the old (broken?) way.


AndreasKl:
Contributor ID is 34720121008014302

dsyer:
We don't use `System.properties` in Spring Boot projects unless we absolutely have to. Since `SimpleHostRoutingFilter` is a `@Bean` I don't really see why you can't use `ZuulProperties` for this.

AndreasKl:
@dsyer I was surprised as well that the code that I amended used `System.properties`. Therefore I sticked to the method already in use. I'll rework the code to use `ZuulProperties` instead.

PR #797, opened 2016-01-27
Add the ability to configure the max connections and max per route on PoolingHttpClientConnectionManager with different values

AndreasKl:
Followup to https://github.com/spring-cloud/spring-cloud-netflix/pull/795 if you are not happy with the config names:
`zuul.hostRoutingFilter.maxTotalConnections` and `zuul.hostRoutingFilter.maxPerRouteConnections`
let me know.

This breaks backwards compatibility as `zuul.max.host.connections` is not applied any more and `SimpleHostRoutingFilter` ctors has changed.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
@@ -57,6 +57,8 @@ public class ZuulProperties {
 
 	private boolean ignoreLocalService = true;
 
+	private HostRoutingFilterProperties hostRoutingFilter = new HostRoutingFilterProperties();
+
 	@PostConstruct
 	public void init() {
 		for (Entry<String, ZuulRoute> entry : this.routes.entrySet()) {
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -57,6 +57,8 @@
 
 	private boolean ignoreLocalService = true;
 
+	private HostRoutingFilterProperties hostRoutingFilter = new HostRoutingFilterProperties();
```
spencergibb: `hostRoutingFilter` is a mouthful. I'm ok with just `host`. @dsyer?

dsyer: Yeah, traditionally this would be an inner class called `Host` with a field name `host`.

AndreasKl:
Totally agree, sorry for the delay. Renamed the classes accordingly.

spencergibb:
@AndreasKl I need you to rebase rather than merge, otherwise there are extra commits.

AndreasKl:
Done.

spencergibb:
Squashed and merged with some polish via aa2d0eb4425c935218834aa2487dd0036b794583.

Issue #809, opened 2016-02-03
Update servo to 0.10.1


spencergibb:
See https://github.com/Netflix/eureka/issues/704

brharrington:
I'm not sure if it still matters for spring, but 0.10.0 was the last version to support java 6. Everything after that expects java 8. If I recall correctly, that is why spring was using an older version.

spencergibb:
ah, very much so at the moment.

brharrington:
From the eureka ticket, it looks like there is a single small commit that is the main change of interest. We can look at cherry picking that commit and creating a 0.10.1 servo release.

spencergibb:
That would be helpful!

brharrington:
There is now a 0.10.1 servo release with a backport of Netflix/servo#362.

spencergibb:
Thanks!

PR #810, opened 2016-02-03
Upgrade servo to 0.12.3

spencergibb:
fixes gh-809

### DIFF
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/metrics/servo/ServoMetricReaderTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/metrics/servo/ServoMetricReaderTests.java
@@ -36,7 +36,7 @@ public class ServoMetricReaderTests {
 				metricNames.get(0));
 		assertEquals("metricName(statistic=max,type=GAUGE,unit=MILLISECONDS)",
 				metricNames.get(1));
-		assertEquals("metricName(statistic=min,type=GAUGE,unit=MILLISECONDS)",
+		assertEquals("metricName(statistic=totalOfSquares,type=NORMALIZED,unit=MILLISECONDS)",
 				metricNames.get(2));
 		assertEquals("metricName(statistic=totalTime,type=NORMALIZED,unit=MILLISECONDS)",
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/metrics/servo/ServoMetricReaderTests.java
```diff
@@ -36,7 +36,7 @@ public void singleCompositeMonitorYieldsMultipleActuatorMetrics() {
 				metricNames.get(0));
 		assertEquals("metricName(statistic=max,type=GAUGE,unit=MILLISECONDS)",
 				metricNames.get(1));
-		assertEquals("metricName(statistic=min,type=GAUGE,unit=MILLISECONDS)",
+		assertEquals("metricName(statistic=totalOfSquares,type=NORMALIZED,unit=MILLISECONDS)",
```
spencergibb: @tbak @jkschneider I'm a little concerned that I had to change this unit test to pass. The min, GAUGE is replaced by totalOfSquares, NORMALIZED. Is that expected?

jkschneider: @spencergibb This is expected.  Here is the reasoning that Daniel Muino provided when this change was made:

The BasicTimer in Servo is used to measure short-duration events.  To accomplish this, it publishes four different values:

`totalTime` (sum of all durations)
`count` (number of times an event happened)
`max` (maximum duration of an event)
`min` (minimum duration of an event, or 0 if no events occurred)

Three of these values have proven useful over time: the `totalTime` and `count` can be used to get an accurate average across arbitrary groups of metrics, and the `max` statistic is used for detecting outlier behavior in clusters.  The `min` value does not have a good use case: it is not used in any alerts, and it does not aggregate well (if any nodes report zero, then the min is zero).

The next release of Servo will replace the `min` statistic with a different one: `totalOfSquares`.  We hope this new statistic will prove more useful to a majority of our users.  With `totalOfSquares`, we can compute an accurate standard deviation across arbitrary groups of metrics.

PR #828, opened 2016-02-15
Feign support for Spring REST parameter annotation member aliases

mbenson:


dsyer:
Thanks. A bit more information in the commit log would be useful. Plus (and I know it's not your problem, but you can help with a solution if you feel like it) the test class should be `*Tests`.

mbenson:
@dsyer: Updated commit message and renamed test class.

mbenson:
I have a further commit available that should at least provide a discussion point for how to support parameter name discovery for annotated parameters as does `spring-web`. Particularly it would be nice to have this discussion before 4fdb85e is released, in case we need to tweak the API anyplace. I don't want to pollute this PR by pushing to the same branch; should I push to a new branch that includes these commits?

dsyer:
A new PR based on this one is fine

PR #835, opened 2016-02-19
Feign parameter #value() fallback

mbenson:
Builds on #828 to provide parameter name fallback on Java 8 with `-parameters` compile flag (bytecode-based parameter name discovery not applicable for `abstract` methods). This may bear some discussion because the implementation provided assumes that the parameter name is an acceptable fallback for any `String value()` annotation element that is defaulted.

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -199,4 +199,25 @@
 			<optional>true</optional>
 		</dependency>
 	</dependencies>
+	<profiles>
+		<profile>
+			<id>java8plus</id>
+				<activation>
+					<jdk>[1.8,2.0)</jdk>
+				</activation>
+				<build>
+					<plugins>
+						<plugin>
+							<groupId>org.apache.maven.plugins</groupId>
+							<artifactId>maven-compiler-plugin</artifactId>
+							<configuration>
+								<compilerArgs>
+									<arg>-parameters</arg>
+								</compilerArgs>
+						</configuration>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
+	</profiles>
 </project>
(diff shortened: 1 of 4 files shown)


spring-cloud-netflix-core/pom.xml
```diff
@@ -199,4 +199,25 @@
 			<optional>true</optional>
 		</dependency>
 	</dependencies>
+	<profiles>
+		<profile>
+			<id>java8plus</id>
+				<activation>
+					<jdk>[1.8,2.0)</jdk>
```
spencergibb: We build with java8 by default, but have to have java 7 compatibility. I'm guessing this breaks that?

mbenson:
I went to some lengths not to break Java 7 compatibility, hence the reflection-based strategy for enabling the related unit test. That Maven profile is enabled when running on JDK >= 1.8 and in that case compiles with the `-parameters` argument, but this doesn't seem to create problematic bytecode. My assumption is that the Java 7 bytecode structure already included whatever mechanism is used in Java 8 to hold the parameter names.

spring-cloud-netflix-core/pom.xml
mbenson: I went to some lengths not to break Java 7 compatibility, hence the reflection-based strategy for enabling the related unit test. That Maven profile is enabled when running on JDK >= 1.8 and in that case compiles with the -parameters argument, but this doesn't seem to create problematic bytecode. My assumption is that the Java 7 bytecode structure already included whatever mechanism is used in Java 8 to hold the parameter names.

spencergibb: OK, we require a java 8 build anyway (one optional module requires it), so this will always be true.

PR #841, opened 2016-02-22
Feign expanders

mbenson:
Builds on #835 to handle expanders

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,6 +33,8 @@ import org.springframework.cloud.netflix.feign.support.SpringMvcContract;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Scope;
+import org.springframework.format.support.DefaultFormattingConversionService;
+import org.springframework.format.support.FormattingConversionService;
 
 import com.netflix.hystrix.HystrixCommand;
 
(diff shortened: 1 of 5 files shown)


dsyer:
I had to revert that change to give us more time to discuss the change. The thread local is ugly (but might be unavoidable in some form given the lack of DI in Feign).

mbenson:
@dsyer , @spencergibb went ahead and updated my PR to eliminate the `ThreadLocal` and use the API introduced in Feign 8.15.0.

spencergibb:
:+1: for not needing the thread local! LGTM if @dsyer also approves.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -56,6 +58,9 @@
 	@Autowired(required = false)
 	private List<AnnotatedParameterProcessor> parameterProcessors = new ArrayList<>();
 
+	@Autowired(required = false)
+	private List<FeignCustomizer> feignCustomizers = new ArrayList<>();
```
spencergibb: `FeignCustomizer` seems generic like it could be used to customize the builder, maybe `FeignFormatterRegistrar`?

mbenson: I kind of left it open-ended in case we later decided there were other extension points that could also be handled by a custom type, like `WebMvcConfigurer`. But whatever you guys are comfortable with is fine with me. @dsyer ?

spencergibb: I say deal with the future in the future :-)

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignFormatterRegistrar.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.format.FormatterRegistrar;
+import org.springframework.format.support.FormattingConversionService;
+
+/**
+ * Allows an application to customize the Feign {@link FormattingConversionService}.
+ *
+ * @author Matt Benson
+ */
+public interface FeignFormatterRegistrar extends FormatterRegistrar {
+
```
mbenson: renamed as requested

spencergibb:
Closed with some polish via 074b2e6ecb6070f0eac50831c19c88ace0c2074a

mbenson:
thanks!

spencergibb:
NP, thanks for the contribution!

Ed42:
still have to name the @PathVariable, Spring Boot 2.5.5 Spring Cloud 2020.0.4

PR #845, opened 2016-02-22
Move ServletDetectionFilter to ZuulConfiguration

### LINKED ISSUE #844: ServletDetectionFilter should be installed by ZuulConfiguration and not ZuulProxyConfiguration
The `Servlet30WrapperFilter` depends on informations in the `RequestContext` set by the `ServletDetectionFilter`.
The problem is that `Servlet30WrapperFilter` is installed by the `ZuulConfiguration` but the `ServletDetectionFilter` by the `ZuulProxyConfiguration`. So you can end up with the first one only.

If the `ServletDetectionFilter` isn't installed, the request inputStream isn't buffered in zuul (cause it's not wrapped by the `Servlet30WrapperFilter`)  and the `ProxyRequestHelper#debugRequestEntity(..)` method consumes the request body and the body won't be forwarded.

Version was 1.1.0-BUILD-SNAPSHOT from today

joshiste:
The Servlet30WrapperFilter depends on infos set in the ServletDetectionFilter
and since the former one is instatiated in the ZuulConfiguration the latter one
should be too.

fixes #844

CLA is signed


spencergibb:
how about a test?

joshiste:
Which behaviour you want to have tested? Creation of the bean? Or the wrapping of a servlet 3.0 request without a ZuulProxyConfiguration?

spencergibb:
If it failed where it currently is, a test that shows that it no longer fails, so the latter.

dsyer:
I couldn't think of a sensible way to test it (and I didn't notice this PR). Fixed via 79b7480.

Issue #865, opened 2016-03-01
Zuul returns 200 OK if the route does not exist instead of 404


aivans:
My config:
```
zuul:
  servletPath: /
  ignored-services: "*"
  routes:
    files:
      path: /aaa/**
      serviceId: aaa-service
      stripPrefix: false
```

I have a the ZuulServlet mapped on the home path.
Requests to `/aaa/` work fine.

I would expect requests to any other path (ex. /bbb) to return a 404 Not Found.

The actual response status code is 200 OK.

In `PreDecorationFilter` there is this code:
```
else {
			log.warn("No route found for uri: " + requestURI);
			ctx.set("forward.to", requestURI);
		}
```
The request is re-executed and somehow an empty response body is returned with 200OK

spencergibb:
This was added specifically in cd30bea0c8bfd75bd27e74c5ff18f69bd0c0b532. 

> Also properly ignores ignored routes, so they can be handled locally
(rather than giving up and sending 404).

@dsyer will need to comment more.

spencergibb:
Maybe a property like `zuul.forwardUnknownRoutes=false`?

aivans:
I don't understand the uses cases where no route is found for a path, but the request is still handled somehow.

What i would do is this:
```
		else {
			log.warn("No route found for uri: " + requestURI);
//			ctx.set("forward.to", requestURI);
			prepareNoFoundResponse();
		}
		return null;
	}

	private void prepareNoFoundResponse() {
		log.info("Preparing 404 error");
		
		RequestContext ctx = RequestContext.getCurrentContext();
		ctx.set("error.status_code", new Integer(HttpStatus.NOT_FOUND.value()));
		ctx.set("error.message", HttpStatus.NOT_FOUND.getReasonPhrase());
		
	}
```
I would prepare an error for the SendErrorFilter.
The response looks like this:
```
{
    "timestamp": 1456914252667,
    "status": 404,
    "error": "Not Found",
    "message": "Not Found"
}
```

dsyer:
If we forward to a path that doesn't exist don't you get a 404 anyway? I think we need to understand where the 200 is coming from before we do anything else.

aivans:
You get a 200 now.
Conceptually speaking, I would prefer to drive everything from the route config. If the path doesn't match a route, than there should be a 404.
Nothing stops the user from configuring an explicit `forward:` for the intended routes, rather than relying on a fallback.

[...]

PR #869, opened 2016-03-02
fixes gh-865

aivans:
It checks first for matching ignored patterns and in this case it performs the forward as in the current version.
If the ignored patterns are not matched and the route is not found, the context is populated with error fields such that SendErrorFilter kicks in


spencergibb:
@dsyer any objections? I've been through it and think it's ok.

dsyer:
Not 100% sure yet. Do we need `getIgnoredRoutes()` still?

spencergibb:
That's how we disable the routes from `DiscoveryClient`.

Issue #339, opened 2015-05-06
Zuul: Hystrix SEMAPHORE isolation blocks the client way far after the timeout


brenuart:
I noticed ZUUL forces the execution isolation to SEMAPHORE instead of the THREAD default (as recommended by Netflix: https://github.com/Netflix/Hystrix/wiki/Configuration#execution.isolation.strategy).

A comment in `org.springframework.cloud.netflix.zuul.filters.route.RibbonCommand` says:
> we want to default to semaphore-isolation since this wraps
> 2 others commands that are already thread isolated

But still I don't get it :-( What are those two other wrapped commands?

Configured this way, Zuul can only shed load but does not allow for timing out and let the client walk away. In short, even if the Hystrix timeout is set to 1000ms, clients will only be released once the call forwarded to the service down the chain returns (or timeouts because of a ReadTimeout for instance).

I tried to force THREAD isolation by overriding the configuration (per service unfortunately, since the default is forced in the code) and everything seems to work as expected. However, I'm not keen in doing this without proper understanding of the implications - certainly in regards of the comment found in the code and the defaults adopted by the Spring Cloud version of Zuul.

Can someone provide more information?
Thx

spencergibb:
This could easily be refactored to be configurable. `ExecutionIsolationStrategy` is an enum and could be put in `ZuulProperties`.

denyska:
Changes in HttpClientRibbonCommand that are part of 1.1.4 release cause NPE in this line `String contentLengthHeader = context.getRequest().getHeader("Content-Length");` if Hystrix configured with THREAD isolation by default.

I know that RestClientRibbonCommand suggests to use SEMAPHORE strategy, but should it still work with the THREAD isolation and all thread-local variables being resolved even before Command created?

I like that THREAD strategy can interrupt threads and I can keep response time upon agreed SLA for all coming requests.

aivans:
Interested in this one also.

stephlag:
As we have performance limitations using the semaphore approach, we would need this change.
I can report @spencergibb  [commit](https://github.com/spring-cloud/spring-cloud-netflix/commit/d1651a48d62c3f2807239230c65f33224a5b7bf4) on the latest master if nobody else works on this issue.

stephlag:
[...]

PR #871, opened 2016-03-03
Allow the Hystrix `ExecutionIsolationStrategy` to be customized in zuul.

spencergibb:
fixes gh-339

/cc @dsyer @brenuart


stephlag:
This change was never merged into the master branch. Was this intended?

spencergibb:
yeah, code diverged too much. the issue is still open.

stephlag:
OK, as we have performance limitations using the semaphore approach, we would need this change.
I can report this change on the latest master if nobody else works on this issue.

dsyer:
Why not leave a comment on the open issue then?

stephlag:
Done

PR #883, opened 2016-03-08
Fixes false 200 in zuul when using zuul servlet rather than mvc servlet

aivans:
The request is fall-backed to DispatcherServlet in all situations.
The PR does not include the possibility to configure another path for handling the fallback. This is on purpose. It may be added in the future, just in time if needed
`spring.static-path-pattern` needs to be added by app developers to solve the 404/405 issue with GET/POST.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
@@ -88,7 +88,9 @@ public class ZuulProxyConfiguration extends ZuulConfiguration {
 	public PreDecorationFilter preDecorationFilter(RouteLocator routeLocator) {
 		return new PreDecorationFilter(routeLocator,
 				this.zuulProperties.isAddProxyHeaders(),
-				this.zuulProperties.isRemoveSemicolonContent());
+				this.zuulProperties.isRemoveSemicolonContent(),
+				this.server.getServletPrefix(),
+				this.zuulProperties);
 	}
 
 	// route filters
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
@@ -87,8 +87,6 @@ public class ZuulProxyConfiguration extends ZuulConfiguration {
 	@Bean
 	public PreDecorationFilter preDecorationFilter(RouteLocator routeLocator) {
 		return new PreDecorationFilter(routeLocator,
-				this.zuulProperties.isAddProxyHeaders(),
-				this.zuulProperties.isRemoveSemicolonContent(),
 				this.server.getServletPrefix(),
 				this.zuulProperties);
 	}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
@@ -27,7 +27,6 @@ import java.util.Set;
 
 import javax.annotation.PostConstruct;
 
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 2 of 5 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java
```diff
@@ -88,7 +88,9 @@ public DiscoveryClientRouteLocator routeLocator() {
 	public PreDecorationFilter preDecorationFilter(RouteLocator routeLocator) {
 		return new PreDecorationFilter(routeLocator,
 				this.zuulProperties.isAddProxyHeaders(),
-				this.zuulProperties.isRemoveSemicolonContent());
+				this.zuulProperties.isRemoveSemicolonContent(),
+				this.server.getServletPrefix(),
+				this.zuulProperties);
```
dsyer: What's the point in passing `zuulProperties` in here as well as `zuulProperties.isAddProxyHeaders()` etc? It's probably best if `PreDecorationFilter` remains independent of `ZuulProperties`, so please change it that way if possible.

aivans: I struggled with that a bit.
Initially I wanted to externalize the "fallback" path also, so I needed 2 provide 2 more parameters to PreDecorationFilter: the configured fallback path + Zuul.servletPath.
The constructor signature would look ugly, so I decided to go for passing the ZuulProperties.
But I see  your point especially from the testing perspective, where just clear intended params without the generic ZuulProperties would keep things nicer.
On the long run though I think passing ZuulProperties directly will be the way to go in order to keep the constructor signature under control and do not rely on setters.
Maybe a fluent builder interface will help.

Do you think it would be good to just pass the servletPath as a parameter for now and do not pass ZuulProperties and we will see how to cintinue in the future?
There are alot

dsyer: I reckon, pass the `ZuulProperties` if it avoids a tangle. If not create a new DTO and initialize it from `ZuulProperties` before you pass it in.

aivans: Is there an example in Spring Cloud where a DTO is used for this purpose? I would like to follow an already established pattern if possible.

dsyer: I can't think of one.

aivans:
I made the ZuulProperties to be the transfer object.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -27,6 +27,7 @@
 
 import javax.annotation.PostConstruct;
 
+import org.springframework.beans.factory.annotation.Value;
```
spencergibb: unused?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -74,7 +75,7 @@
 
 	private boolean traceRequestBody = true;
 
-	private boolean removeSemicolonContent = true;
+	private boolean removeSemicolonContent = true;
```
spencergibb: revert whitespace

aivans:
Hope it is ok now.

aivans:
Can we include this one for 1.1.0RC1?

spencergibb:
@dsyer any objections? (I'm out Thurs Mar 17 & Fri the 18th btw).

dsyer:
I'm out on a project till Friday. I think we got to a good place here. Might be a bit of polishing to do maybe?

spencergibb:
@aivans how can I recreate the problem? I'd like to see it before and after with this fix.

aivans:
@spencergibb 
you need to curl your application, the request be handled by ZuulServlet, but no route to be matched:
```curl http://localhost:8765/~zuulServletPath~/justaroute```

spencergibb:
Squashed and merged via e87d9432c705e2435e31a5339b93487ce1f825e4

PR #884, opened 2016-03-08
8.15 version of Feign breaks VND content types

marcingrzejszczak:
This test won't pass with the current version of feign (8.15.0). It passes with 8.14.4. The brewery app is broken because of this change too.

Corresponding Feign issue: https://github.com/Netflix/feign/issues/345


adriancole:
fixed in feign 8.15.1 which is out now

spencergibb:
@adriancole sweet!

spencergibb:
Thanks @adriancole and @marcingrzejszczak. Things didn't stay broken very long.

PR #891, opened 2016-03-09
Verify that the stream is more than ZERO bytes before un-gzipping

jecuendet:
When the server behind ZUUL send a GZIPPed stream, but the stream is ZERO bytes in length, ZUUL crashes while decompressing it


dsyer:
Merged after fixing compiler error. Thanks.

PR #885, opened 2016-03-08
Zuul rate limit

viniciusccarvalho:
Adds ratelimit filter to zuul. 

The filter is enabled when `zuul.ratelimit.enabled` is present.

It has an inmemory and a redis based implementation. The redis implementation will work with multiple instances.

There's two policies supported:
```
zuul:
  ratelimit:
    enabled: true
    policies:
        anonymous:
            refreshInterval: 60
            limit: 1000
        authenticated:
            refreshInterval: 60
            limit: 2000
```
The policies control how we create the buckets. Anonymous will use a global bucket counter, the authenticated uses the security principal as the key, so each authenticated user has its own counter.

PR #895, opened 2016-03-09
Rate limit

viniciusccarvalho:
This is a cleaner version of the PR #885 

Adds ratelimit filter to zuul. 

The filter is enabled when `zuul.ratelimit.enabled` is present.

It has an inmemory and a redis based implementation. The redis implementation will work with multiple instances.

There's two policies supported:
```
zuul:
  ratelimit:
    enabled: true
    policies:
        anonymous:
            refreshInterval: 60
            limit: 1000
        authenticated:
            refreshInterval: 60
            limit: 2000
```
The policies control how we create the buckets. Anonymous will use a global bucket counter, the authenticated uses the security principal as the key, so each authenticated user has its own counter.

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -198,6 +198,17 @@
 			<artifactId>jackson-dataformat-smile</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-redis</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>com.github.kstyrc</groupId>
+			<artifactId>embedded-redis</artifactId>
+			<version>0.6</version>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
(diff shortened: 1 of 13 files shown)


shanman190:
Any plans to get this error fixed and merged?

spencergibb:
This isn't an error or bug, but a new feature. We'll consider it for our next release train, Camden.

shanman190:
Sorry for the confusing, I was meaning the build error preventing this from making it into the Brixton. It appeared to have been opened and forgotten about.

kakawait:
take a look at https://github.com/marcosbarbero/spring-cloud-starter-zuul-ratelimit

marcosbarbero:
@kakawait thanks to the reference.
The biggest different between my version to this PR is the `per-service` configuration and the `rate limit` strategy per user, origin or both combined.

I think the sample configuration will talk by itself.

```
zuul:
  ratelimit:
    enabled: true #default false
    policies:
      myServiceId:
        limit: 10
        refresh-interval: 60 #default value (in seconds)
        type: #optional
          - user 
          - origin 
```

spencergibb:
Closing for now. We will revisit for Dalston

PR #930, opened 2016-03-29
Remove need for http basic auth client filter.

spencergibb:
https://github.com/Netflix/eureka/releases/tag/v1.4.5

1.4.5 restores http basic auth.

I setup a server with spring security and basic auth with a client and it worked. (Netflix now tests that basic auth works and shouldn't break in the future). Not sure how to unit/integration test it, ideas @dsyer?


spencergibb:
Closed via c0d767351ac86942c505b58a741dd80a7ff275b7

zyro23:
and this did really work for you? i get again just 401s... i think the upstream fix only applied to `eureka-client-jersey2` while `spring-cloud-netflix` depends on `eureka-client` which seems to be jersey1?

spencergibb:
@zyro23 yes, it works for me. If you can reproduce the problem please open a new issue and provide a sample project.

Issue #648, opened 2015-11-17
Ribbon Retry with RestTemplate?


tkvangorder:
I have been migrating an existing application over to Spring Cloud's service discovery, Ribbon load balancing, and circuit breakers. The application already makes extensive use of the RestTemplate and I have been able to successfully use the load balanced version of the template. However, I have been testing the situation where there are two instances of a service and I drop one of those instances out of operation. I would like the RestTemplate to failover to the next server. From the research I have done, it appears that the fail-over logic exists in the Feign client and when using Zuul. It appears that the LoadBalancedRest template does not have logic for fail-over. In diving into the code, it looks like the RibbonClientHttpRequestFactory is using the netflix RestClient (which appears to have logic for doing retries).

So where do I go from here to get this working? 

I would prefer to not use the Feign client because I would have to sweep A LOT of code.
I had found this [link](http://stackoverflow.com/questions/27468697/spring-cloud-with-resttemplate-ribbon-eureka-retry-when-server-not-available) that suggested using the @Retryable annotation along with @HystrixCommand but this seems like something that should be a part of the load balanced rest template. 

I did some digging into the code for RibbonClientHttpRequestFactory: 

```
		protected ClientHttpResponse executeInternal(HttpHeaders headers)
				throws IOException {
			try {
				addHeaders(headers);
				if (outputStream != null) {
					outputStream.close();
					builder.entity(outputStream.toByteArray());
				}
				HttpRequest request = builder.build();
				HttpResponse response = client.execute(request, config);
				return new RibbonHttpResponse(response);
			} catch (Exception e) {
				throw new IOException(e);
			}

```
It appears that if I override this method and change it to use "client.executeWithLoadBalancer()" that I might be able to leverage the retry logic that is built into the RestClient? This would require me to create my own version of the RibbonClientHttpRequestFactory, I am hoping I am not wandering too far down the rabbit hole. 

Just looking for guidance on the best approach. 

Thanks

tkvangorder:
@spencergibb I was able to get fail-over working in a modified version of RibbonClientHttpRequestFactory. Was there a reason in choosing "client.execute() instead of client.executeWithLoadBalancer? Can we add this capability into the core? I can generate a pull request if you would like.

I had moved this question to stack overflow and I have documented my changes here: 

[Stack OverFlow Link](http://stackoverflow.com/questions/33765049/spring-cloud-getting-retry-working-in-resttemplate/33786574#33786574)

[...]

Issue #357, opened 2015-05-21
Ribbon - RestTemplate Error


aivans:
related to  #350 probably

I get this error:
```
Caused by: java.io.IOException: com.sun.jersey.api.client.ClientHandlerException: org.apache.http.client.ClientProtocolException
        at org.springframework.cloud.netflix.ribbon.RibbonClientHttpRequestFactory$RibbonHttpRequest.executeInternal(RibbonClientHttpRequestFactory.java:125) ~[spring-cloud-netflix-core-1.0.3.BUILD-SNAPSHOT.jar:1.0.3.BUILD-SNAPSHOT]
        at org.springframework.http.client.AbstractClientHttpRequest.execute(AbstractClientHttpRequest.java:53) ~[spring-web-4.1.2.RELEASE.jar:4.1.2.RELEASE]
        at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:568) ~[spring-web-4.1.2.RELEASE.jar:4.1.2.RELEASE]
        ... 122 common frames omitted
Caused by: com.sun.jersey.api.client.ClientHandlerException: org.apache.http.client.ClientProtocolException
        at com.sun.jersey.client.apache4.ApacheHttpClient4Handler.handle(ApacheHttpClient4Handler.java:184) ~[jersey-apache-client4-1.11.jar:1.11]
        at com.sun.jersey.api.client.Client.handle(Client.java:648) ~[jersey-client-1.11.jar:1.11]
        at com.sun.jersey.api.client.WebResource.handle(WebResource.java:670) ~[jersey-client-1.11.jar:1.11]
        at com.sun.jersey.api.client.WebResource.access$200(WebResource.java:74) ~[jersey-client-1.11.jar:1.11]
        at com.sun.jersey.api.client.WebResource$Builder.post(WebResource.java:563) ~[jersey-client-1.11.jar:1.11]
        at com.netflix.niws.client.http.RestClient.execute(RestClient.java:621) ~[ribbon-httpclient-2.0.0.jar:2.0.0]
        at com.netflix.niws.client.http.RestClient.execute(RestClient.java:527) ~[ribbon-httpclient-2.0.0.jar:2.0.0]
[...]

PR #945, opened 2016-04-01
Fixes for #648 and #357

tkvangorder:
This pull request addresses two issues that we encountered as we moved our code base to use Spring Cloud.

https://github.com/spring-cloud/spring-cloud-netflix/issues/648

The primary fix is to add retry capabilities into the RestTemplate such that it will respect the Ribbon Retry settings similar to how the Feign client works.

This fix changes the RibbonHttpRequest to use the 'executeWithLoadBalancer()'. This change leverages Ribbon's LoadBalancedCommand to issue the rest call. This command will correctly select a server instance and update the server stats for that server. Additionally, this command will also retry the request should a connection failure occur. Because this logic is done in the Ribbon layer, it is no longer necessary to choose a server instance in the RibbonClientHttpRequestFactory or to collect the serverstats in the RibbonHttpRequest.

There is a new unit test, `RestTemplateRetryTest` that exercises the logic and insures the server stats are being correctly updated and that retry logic is working as expected. 

Notes: 

I ran into an issue where when a Ribbon client is configured without the use of Eureka, it's 
'DeploymentContextBasedVipAddresses' property was not set. Failure to set this means the ribbon client will not attempt to resolve the client ID to one of the server instances. I found code in 'EurekaRibbonClientConfiguration' that was setting this and I ended up copying this code to RibbonClientConfiguration. This may not be the best way to solve this, @spencergibb or @dsyer is there a better place to do something like this? 

The unit test leverages a simple AvailabilityFilteringRule that just skips servers that have open circuit breakers. It is more difficult to come up with reproducible, easy checks for other types of rules...although I used this test to figure out how other Rules actually work.

This change will affect anyone that is wrapping their rest calls in a high level "retry" mechanism (such as hystrix or using Spring's Retry annotation.

https://github.com/spring-cloud/spring-cloud-netflix/issues/357 

This issue was closed by the original author because he had hacked around the problem. In our case we had a large code base that made this approach challenging. Our fix was again to rely on the Ribbon library to correctly manipulate the Http header. 

See my comment at the end of this issue for details.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
@@ -16,7 +16,11 @@
 
 package org.springframework.cloud.netflix.ribbon;
 
+import static com.netflix.client.config.CommonClientConfigKey.DeploymentContextBasedVipAddresses;
+
 import java.net.URI;
+
+import javax.annotation.PostConstruct;
 
 import org.apache.http.client.params.ClientPNames;
 import org.apache.http.client.params.CookiePolicy;
@@ -32,6 +36,9 @@ import com.netflix.client.DefaultLoadBalancerRetryHandler;
 import com.netflix.client.RetryHandler;
 import com.netflix.client.config.DefaultClientConfigImpl;
 import com.netflix.client.config.IClientConfig;
+import com.netflix.config.ConfigurationManager;
+import com.netflix.config.DynamicPropertyFactory;
+import com.netflix.config.DynamicStringProperty;
 import com.netflix.loadbalancer.ConfigurationBasedServerList;
 import com.netflix.loadbalancer.ILoadBalancer;
 import com.netflix.loadbalancer.IPing;
@@ -56,6 +63,10 @@ import com.sun.jersey.client.apache4.ApacheHttpClient4;
 @EnableConfigurationProperties
 public class RibbonClientConfiguration {
 
+	protected static final String VALUE_NOT_SET = "__not__set__";
+
+	protected static final String DEFAULT_NAMESPACE = "ribbon";
+	
 	@Value("${ribbon.client.name}")
 	private String name = "client";
 
@@ -155,6 +166,28 @@ public class RibbonClientConfiguration {
 		return new DefaultServerIntrospector();
 	}
 
+	@PostConstruct
+	public void preprocess() {
+		setProp(name, DeploymentContextBasedVipAddresses.key(), name);		
+	}
+
+	protected void setProp(String serviceId, String suffix, String value) {
+		// how to set the namespace properly?
+		String key = getKey(serviceId, suffix);
+		DynamicStringProperty property = getProperty(key);
+		if (property.get().equals(VALUE_NOT_SET)) {
+			ConfigurationManager.getConfigInstance().setProperty(key, value);
+		}
+	}
+
+	protected DynamicStringProperty getProperty(String key) {
+		return DynamicPropertyFactory.getInstance().getStringProperty(key, VALUE_NOT_SET);
+	}
+
+	protected String getKey(String serviceId, String suffix) {
+		return serviceId + "." + DEFAULT_NAMESPACE + "." + suffix;
+	}
+	
 	static class OverrideRestClient extends RestClient {
 
 		private ServerIntrospector serverIntrospector;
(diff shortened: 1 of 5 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
```diff
@@ -155,6 +166,28 @@ public ServerIntrospector serverIntrospector() {
 		return new DefaultServerIntrospector();
 	}
 
+	@PostConstruct
+	public void preprocess() {
+		setProp(name, DeploymentContextBasedVipAddresses.key(), name);
```
spencergibb: Why did you need to copy this from `EurekaRibbonClientConfiguration`?

spencergibb: And now I re-read your comments :-) Now I wonder why it's needed in both places.

tkvangorder: Yeah, It is definitely redundant...I dont know enough about the internals of how Eureka is used to create a ribbon client dynamically vs. having a ribbon client explicitly defined with RibbonClientConfiguration.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
```diff
@@ -38,12 +34,8 @@
 
 	private final SpringClientFactory clientFactory;
 
-	private LoadBalancerClient loadBalancer;
```
spencergibb: This really worries me, though I can't put specifics on it yet.

spencergibb: Looking at `RibbonLoadBalancerClient` now.

spencergibb: Probably the biggest thing I can see is automatically detecting `isSecure` vs. having to set the value in properties.

spencergibb: Looks like that might be taken care of with `OverrideRestClient`.

spencergibb:
@dsyer thoughts?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientHttpRequestFactory.java
tkvangorder: yeah, its a little confusing because the LoadBalancerCommand is doing similar things to the logic in the RibbonLoadBalancerClient. Is the "isSecure" flag there so that if some instances are secure and some are not, the URI will have the correct "http" or "https" when using that server?

spencergibb: Yeah. We plug into the `RestClient` here https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java#L170-L177

spencergibb: I'm trying to remember why we did that in the first place. I'm thinking it was to use the `LoadBalancerClient` abstraction, but at that point, we're in the middle of ribbon already.

spencergibb:
@tkvangorder after talking with @dsyer we'd like this feature to be opt-in via a config property. Is that something you'd be willing to change?

spencergibb:
Actually, I'm in the middle of doing #961 which makes `RibbonClientHttpRequestFactory` opt-in. Since that is where your changes are, #961 effectively makes this opt-in. Once I merge, can you rebase?

spencergibb:
Actually, I can rebase.

tkvangorder:
Haha, I dont even time to respond. 8) So, if I am understanding this correctly, after your changes:

Using The RibbonClientHttpRequestFactory (which uses Netflix's RestClient) will be an opt-in. 
And if you opt-in, you will also get the retry logic from my pull request?

spencergibb:
Yes, that is what I'm saying.

spencergibb:
Squashed and merged with some polish via 4479a2b2ee28cba1ea86dcca18805cce369634d8

tkvangorder:
Awesome! It looks like this will get into Brixton?

spencergibb:
Yes, I wanted it before RC2 which should be in the next week.

PR #952, opened 2016-04-05
Set or append to X-Forwarder-For header on Zuul requests

### LINKED ISSUE #352: Zuul and X-Forwarded-For
Should Zuul support X-Forwarded-For out of the box? Just like it sets X-Forwarded-Host

timtebeek:
Fixes #352


dsyer:
Is it worth maybe checking for duplicates (no need to add an address that is already there because of an upstream proxy)?

timtebeek:
Good point, updated accordingly.. Anything else I might add? I saw I'm slightly off-standard with my lack of curly brackets.. That is easily fixed if needed.

dsyer:
I think the braces are covered by a checkstyle rule, but the easiest way to fix them is to use Eclipse and load the coding conventions from the source (https://github.com/spring-cloud/spring-cloud-netflix/tree/master/eclipse).

timtebeek:
Just saw https://github.com/spring-cloud/spring-cloud-netflix/commit/593549551ac93848328203e1ad5bba8a60c20cd3 ; Great! Does that mean I can close this pull request? Or what's the proper flow?

spencergibb:
Closed via 593549551ac93848328203e1ad5bba8a60c20cd3

PR #967, opened 2016-04-12
In SpringEncoder keep headers set by HttpMessageConverters

oleg:
Some `HttpMessageConverter`s call `HttpOutputMessage.getHeaders()`'s and modify returned `HttpHeaders`. `FeignOutputMessage` in `SpringEncoder` always returns new `HttpHeaders` thus making some of HttpMessageConverters unusable. Notable example is `FormHttpMessageConverter` which adds `boundary` param to `Content-Type` header. 

Passing charset to `request.body(...)` makes underlying code behave incorrectly. In case `request.charset()` is not null `feign.httpclient.ApacheHttpClient` tries to create `StringEntity` and could fail with an error calling  `ContentType.create(...)` with parameters. Anyway passing null as charset will lead to creation of `ByteArrayEntity` which seems more desirable.

The main problem this patch solves is sending multipart request.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringEncoder.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringEncoder.java
@@ -22,8 +22,9 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.reflect.Type;
-import java.nio.charset.Charset;
 import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 
 import lombok.extern.apachecommons.CommonsLog;
 
@@ -89,8 +90,13 @@ public class SpringEncoder implements Encoder {
 					catch (IOException ex) {
 						throw new EncodeException("Error converting request body", ex);
 					}
+
+					for (Map.Entry<String, List<String>> entry : outputMessage.getHeaders().entrySet()) {
+						request.header(entry.getKey(), entry.getValue());
+					}
+
 					request.body(outputMessage.getOutputStream().toByteArray(),
-							Charset.forName("UTF-8")); // TODO: set charset
+						null); // TODO: set charset
 					return;
 				}
 			}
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/SpringEncoderTests.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import static org.junit.Assert.assertEquals;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = SpringEncoderTests.Application.class)
+@WebAppConfiguration
+@IntegrationTest({"server.port=0", "spring.application.name=springencodertest",
+    "spring.jmx.enabled=true"})
```
spencergibb: We use tabs instead of spaces.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/SpringEncoderTests.java
```diff
@@ -0,0 +1,102 @@
...
+@DirtiesContext
+public class SpringEncoderTests extends FeignClientFactoryBean {
```
spencergibb: I don't like two tests called `SpringEncoderTests`. Maybe `MultipartFormDataTests`?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringEncoder.java
```diff
@@ -106,11 +112,10 @@ public void encode(Object requestBody, Type bodyType, RequestTemplate request)
 	private class FeignOutputMessage implements HttpOutputMessage {
 
 		private final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
-
-		private final RequestTemplate request;
+		private final HttpHeaders httpHeaders;
 
 		private FeignOutputMessage(RequestTemplate request) {
-			this.request = request;
+			httpHeaders = getHttpHeaders(request.headers());
```
spencergibb: `this.httpHeaders`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringEncoder.java
```diff
@@ -89,8 +90,13 @@ public void encode(Object requestBody, Type bodyType, RequestTemplate request)
...
 						throw new EncodeException("Error converting request body", ex);
 					}
+
+					for (Map.Entry<String, List<String>> entry : outputMessage.getHeaders().entrySet()) {
+						request.header(entry.getKey(), entry.getValue());
+					}
+
 					request.body(outputMessage.getOutputStream().toByteArray(),
-							Charset.forName("UTF-8")); // TODO: set charset
+						null); // TODO: set charset
```
spencergibb: Not sure about this change. What's the idea here?

Issue #973, opened 2016-04-15
Eureka Dashboard shows basic auth data for replicas when using security


mattreyuk:
When we register with Eureka using basic auth, those auth credentials are displayed on the dashboard for Eureka server replicas (registered,unavailable and available replicas and link for DS Replicas). 

I have a PR I should be able to submit shortly to strip this data in EurekaController

dsyer:
Fixed in #974

PR #974, opened 2016-04-15
filter basic auth credentials from Eureka dashboard

mattreyuk:
This PR fixes issue #973 by removing the `user:password@` part of Eureka replica urls with basic auth in the model used for the dashboard.


dsyer:
Thanks. Did you fill in the CLA (for this or any project)? Link in README.

mattreyuk:
I just did.

dsyer:
Thanks.

joshiste:
I wonder if this solution is sufficient. 
Wouldn't it be better to not include the credentials in the url and configure it elsewhere? The url might get shown in logoutput and the `/env` endpoint.

dsyer:
@joshiste yes, but this is a start. The embedded credentials are actually the only way this is supported in the Netflix client library.

PR #975, opened 2016-04-18
Verify request encoding before set the default encoding

marcosbarbero:


marcosbarbero:
I've messed up commits on rebase, ignore it and please consider the follow PR https://github.com/spring-cloud/spring-cloud-netflix/pull/979

PR #979, opened 2016-04-19
Use request.characterEncoding prior to default encoding

marcosbarbero:
Zuul is using ISO-8859-1 as default character encoding making requests fail when it's expected any other encoding. The problem occurs at URI encoding and queryParams encoding.

Related to this [post from stackoverflow](http://stackoverflow.com/questions/36475845/netflix-zuul-query-string-encoding)

### DIFF
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/SampleZuulProxyApplicationTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/SampleZuulProxyApplicationTests.java
@@ -146,7 +146,7 @@ public class SampleZuulProxyApplicationTests extends ZuulProxyTestBase {
 				"http://localhost:" + this.port + "/self/query?foo={foo}", HttpMethod.GET,
 				new HttpEntity<>((Void) null), String.class, "weird#chars");
 		assertEquals(HttpStatus.OK, result.getStatusCode());
-		assertEquals("/query?foo=weird%23chars", result.getBody());
+		assertEquals("/query?foo=weird#chars", result.getBody());
 	}
 
 	@Test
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/SampleZuulProxyApplicationTests.java
```diff
@@ -146,7 +146,7 @@ public void simpleHostRouteWithEncodedQuery() {
 				"http://localhost:" + this.port + "/self/query?foo={foo}", HttpMethod.GET,
 				new HttpEntity<>((Void) null), String.class, "weird#chars");
 		assertEquals(HttpStatus.OK, result.getStatusCode());
-		assertEquals("/query?foo=weird#chars", result.getBody());
+		assertEquals("/query?foo=weird%23chars", result.getBody());
```
spencergibb: I'm worried that this test changed. The `#` char has meaning in a url, if it's changed, I'm not sure I can accept this.

marcosbarbero: @spencergibb you're right the '#' is a gen-delims as specified at RFC, I didn't payed attention on that. 
Anyhow, the URI encoding is a real problem. I'll fix that PR.

marcosbarbero:
@spencergibb the encoding URI is a real problem. I'll give you a scenario.
1. Sending a request to `/zuul/any-sp%C3%A9cial-char/action` with `UTF-8` encoding
2. `ProxyRequestHelper` class will receive it like `/zuul/any-spécial-char/action`
3. `ProxyRequestHelper` class will encode it like `/zuul/any-sp%E9cial-char/action` with  `ISO-8859-1` encoding
4. Zuul routes the request to the final service and it cannot read the request because it is expecting `UTF-8` instead of `ISO-8859-1`

---
My approach here is to check the incoming request encoding and try to use it to encode de URI instead of always use the default value.

spencergibb:
@marcosbarbero I'm not debating that it is not a real problem. Some tests would be helpful.

marcosbarbero:
@spencergibb I'm sorry if it sounded rude, I'd not intention to say you were debating.
I'll try to provide some tests in few days.

spencergibb:
@marcosbarbero NP! Looking forward to the tests.

marcosbarbero:
@spencergibb I've just added few tests. Hope it helps, if need some more scenarios drop me a message :)

spencergibb:
Nice!

spencergibb:
Merged via 1c532d96375485b8d551cbbe766507bacfaafc26

Issue #966, opened 2016-04-12
@FeignClient  -  root context path


tzebrowski:
Hi,

 I have a case where the endpoint has root path defined at  the top level, on the interface, like below

```
@RequestMapping("/foo/")
public interface FooService{

@RequestMapping("/foo1/")
public void foo1();
.....
@RequestMapping("/foo100/")
public void foo100();
}
```
I have client interface which relays on @FeignClient which has similar definition
```
@RequestMapping("/foo/")
@FeignClient(name = "fooService")
public interface FooServiceClient{

@RequestMapping("/foo1/")
public void foo1();
.....
@RequestMapping("/foo100/")
public void foo100();
}
```
Definition also could look like as follows
```
@FeignClient(name = "fooService")
public interface  FooServiceClient extends FooService {}
```

The result of such feign client definition is that spring registers another endpoint on the client side with root  path /foo/foo1,/foo/foo2....

Another issue appears when we define fallback class which is inherited from the FooServiceClient.
Spring also tries to register another endpoint with the same definition as FooServiceClient service has.
```
@FeignClient(name = "fooService", fallback = FooServiceFallback .class)
public interface FooServiceClient extends FooService {
}

@Component
public class FooServiceFallback  implements FooServiceClient {
     
}
```

My question is whenever we can pass  root path to the @FeignClient in the different way, to avoid necessary endpoint registration?

spencergibb:
We don't support putting `@RequestMapping` on the Feign interface. This is a limitation of Spring MVC in that beans that are annotated with `@RequestMapping` on the class are registered as beans even thought they aren't annotated with `@Component`.

tzebrowski:
Ok. My question is rather, how I can pass root-path to the @FeignClient then. Let's assume that I would change the definition of FooService to the form as bellow

```
public interface FooService{

@RequestMapping("/foo1/")
public void foo1();
.....
@RequestMapping("/foo100/")
public void foo100();
}
```

```
@RequestMapping("/foo/")
public interface FooServiceImpl implements FooService{

public void foo1();
.....
public void foo100();
}
```
This change makes that service is available on the same address

/foo/foo1
....
/foo/foo100


When we back to our definition of FooServiceClient 
```
@FeignClient(name = "fooService", fallback = FooServiceFallback .class)
public interface FooServiceClient extends FooService {
}
```
we will see two changes

- /foo/foo1; ... ; /foo/foo100 endpoints will not registers on the client side
[...]

PR #983, opened 2016-04-23
Added path parameter to @FeignClient

venilnoronha:
I've added a `path` parameter to `@FeignClient` which if specified would be prefixed to all method level mappings. Placeholder resolution for this parameter is also provided. See #966 for more info.

Please review and pull. I've signed the ICLA.

Thanks,
Venil Noronha

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dsyer:
Close in favour of #984

venilnoronha:
@dsyer isn't this PR is a little different than #984? Here, we can define a `path` parameter via code (or property placeholder) at the `@FeignClient` level so that all the methods mappings within the client are prefixed with the `path`.

dsyer:
I see, thanks for the clarification.

venilnoronha:
Thanks for considering. :)

cforce:
@dsyer Will this feature be available with 1.1.0 Release? Looks like not merged currently, is it?

dsyer:
It shouldn't break any existing clients right?

spencergibb:
Squashed and merged via 83caebf384f3454b31571a15e630f8689b1c1abe

venilnoronha:
Thanks!

cforce:
superb ;)

Issue #894, opened 2016-03-09
Add support for placeholders to Feign spring-mvc annotations.


PedroAlvarado:
I'd be awesome if spring-cloud-feign supported placeholders in spring-mvc annotations as shown below.

```
@FeignClient("foo")
public interface FooClient{
    @RequestMapping(value = "${my.special.placeholder}",
            method = RequestMethod.GET,
            consumes = MediaType.APPLICATION_JSON_VALUE,
            produces = MediaType.APPLICATION_JSON_VALUE)
    ResponseEntity getSomeData(s);
}
```

PedroAlvarado:
@spencergibb If you can give me a lead, I'd take a stab at implementing this feature.

spencergibb:
I'd start looking at `SpringMvcContract.processAnnotationsOnParameter` and this commit 94d4bdfcc38cb9cc505fe249a0f79f4431b03f8d for how it was done on `@FeignClient`.

sfat:
@PedroAlvarado you still want to work on this? If not, gonna make a PR for this guy

PR #984, opened 2016-04-23
Add support for placeholders to Feign spring-mvc RequestMapping annot…

PedroAlvarado:
Improvement to address: https://github.com/spring-cloud/spring-cloud-netflix/issues/894


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
```diff
@@ -64,6 +69,7 @@
 
 	private final ConversionService conversionService;
 	private final Param.Expander expander;
+	private ResourceLoader resourceLoader = new DefaultResourceLoader();
```
spencergibb: Just autowire an application context.

PedroAlvarado: This class implements ResourceLoaderAware and therefore has a setResourceLoader() method. You are suggesting I remove the interface and instead that I simply use `	@Autowired
	private ApplicationContext applicationContext;` instead correct?

The reason I went this route is because I was bit concerned about this class being used without and application context around and the field initialization was meant to guard against a possible NPE as a result. That said, confirm the change and I'll update the code.

spencergibb: Hold off, let me look again after I get some sleep :-)

dsyer:
Closed via 2b40952

Issue #971, opened 2016-04-14
Netflix Zuul query string encoding 


anoll:
This is a copy of the stackoverflow issue:
http://stackoverflow.com/questions/36475845/netflix-zuul-query-string-encoding

When sending a request via Zuul to a client, Zuul seems to change the query String. More specifically, if the client should receive an url-encoded query String, Zuul decodes the query String once. Here is a concrete example:

If "http://localhost:8080/demo/demo?a=http%3A%2F%2Fsomething/" is sent to the client, the client receives as a query String "a=http://something/".

Looking into Zuul`s code, the function "buildZuulRequestQueryParams" uses "HTTPRequestUtils.getInstance().getQueryParams();" which decodes the query String.

Is this a desired feature or a bug?

dsyer:
What version of Spring Cloud Netflix is this? Did you try from master?

jebeaudet:
I believe query strings should not be decoded when passed to the backend since Zuul is only a proxy. @anoll is right, in master branch, the ProxyRequestHelper still uses `HTTPRequestUtils.getInstance().getQueryParams()` which decode the query strings.

yashanet:
I have the same issue. But when we have slash in parameter name, for example :
http://localhost:8080/demo/?name/withSlash=value".
So please remove URLDecoder.decode() from HTTPRequestUtils.getQueryParams()

dsyer:
`HTTPRequestUtils` is a Netflix utility class. If you want to change it please go to https://github.com/Netflix/zuul. If you want to fix it here, then a pull request is more than welcome.

jebeaudet:
First, there is 2 code path with this problem. 

1. 
If you're using `SimpleHostRoutingFilter`, the query params will be decoded using the `URLDecoder` in `HTTPRequestUtils` which is wrapped inside the `ProxyRequestHelper::buildZuulRequestQueryParams`. These will get reencoded in `ProxyRequestHelper::getQueryString` using `UriTemplate`. 

2. 
If you're using `RibbonRoutingFilter`, the query params will be decoded as in 1 in `ProxyRequestHelper::buildZuulRequestQueryParams` but will not get reencoded and will get send to the backend as is in `HttpClientRibbonCommand` or `RestClientRibbonCommand`. 

[...]

PR #989, opened 2016-04-26
zuul proxy changes order of query params

nbyl:
When forwarding requests using the SimpleHostRoutingFilter query parameters are first parsed to a map and afterwards reconstructed when forwarding. This leads sometimes to re-ordering and the query `?edit&p1=v1` may accidentally be changed to `?p1=v1&edit`.

This should usually not be a problem, but some legacy applications expect the parameters in the query string to have a certain order. The patches re-uses the original query string when forwarding requests to ensure compatibility with these applications.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
@@ -230,14 +230,20 @@ public class ProxyRequestHelper {
 	}
 
 	public Map<String, Object> debug(String verb, String uri,
-			MultiValueMap<String, String> headers, MultiValueMap<String, String> params,
+									 MultiValueMap<String, String> headers, MultiValueMap<String, String> params,
+									 InputStream requestEntity) throws IOException {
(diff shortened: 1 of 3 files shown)


dsyer:
I had to revert this change because it caused some test failures. Can you look into it please?

nbyl:
Hi,

sorry for the long reply time. I have looked into the issue and there is a conflict between the "classic" behaviour (parse and filter the query parameters) and the behaviour required in this use case (pass the query string through as it is). 

I'm currently working on a proposal, which should be done next week.

Greetings,

Nicolas

spencergibb:
ping @nbyl

nbyl:
@spencergibb Sorry for the delay. I have checked the code but the problem is currently not solvable. The parameters are mixed together inside the netflix stack. They are passed back as a map containing no information about the original order. So there is currently no way to solve this in SimpleHostRoutingFilter as I have tried here.

We have solved this using custom zuul filter that bypasses all filtering done insinde the current code and forwards the stuff directly to the target. Are you interested to have this feature in the project?

jebeaudet:
Related to my last comment here https://github.com/spring-cloud/spring-cloud-netflix/issues/971

spencergibb:
Closing 

>  there is currently no way to solve this in SimpleHostRoutingFilter as I have tried here

PR #994, opened 2016-04-27
Fix #993 - Manage X-Forwarded-Prefix

### LINKED ISSUE #993: Zuul overwrites X-Forward-Prefix
A request with a "X-Forwarded-Prefix" header is overwritten by zuul only if route has prefix, so when stripPrefix is true.

Example:
```
route1:
  path: /service/**
  serviceId: service
route2:
  path: /endpoint/**
  serviceId: service
  stripPrefix: false
```

curl on route1:
`curl -H "Host: example.com" -H "X-Forwarded-Prefix: /api" localhost/endpoint`
```
{
  {},
  "_links" : {
    "self" : {
      "href" : "http://example.com/api/endpoint"
    }
  }
}
```                                                                                                                                              
curl on route2:
`curl -H "Host: example.com" -H "X-Forwarded-Prefix: /api" localhost/service/endpoint`
```
{
  {},
  "_links" : {
    "self" : {
      "href" : "http://example.com/service/endpoint"
    }
  }
}    
```

I expected to be `http://example.com/api/service/endpoint`.

Should the "X-Forwarded-Prefix" header be concatened instead of overwritten ?

speralta:
I've managed the incoming X-Forwarded-Prefix.

I've signed the ICLA.


dsyer:
Merged thanks (5613e48 )

Issue #816, opened 2016-02-08
Feign doesn't support regexp in request mapping


asm0dey:
If I use interface 
```java
public interface UserAPI {
  @RequestMapping(method = GET, path = "/user/{username:[a-z@A-Z\\d\\.-_]+}"
  UserDTO getUserByUsername();
}
```
And want to create client as
```java
@FeignClient("service")
public interface UserClient extends UserAPI{}
```
it won't work because feign thinks that pattern is part of path. It's even more impressive if you want to use pattern like `.+`

jmgilmour:
Any timeframe on this being included in an upcoming release?

spencergibb:
@jmgilmour no. PRs welcome.

jmgilmour:
Thanks for the quick response.  I was under the impression that the referenced commit by @venilnoronha  above would fix this issue.

spencergibb:
That commit merely didn't error if there was a regex, it didn't support them.

jmgilmour:
Good to know :smile:.  I guess I'll take a crack at a fix and PR when I get the chance.  Again, thanks for the quick response.

PR #1008, opened 2016-05-06
Support regex in feign client @RequestMapping

venilnoronha:
I've added `@RequestMapping` regex parsing to `SpringMvcContract` to now accept `@FeignClient` mappings such as `@RequestMapping("/{login:[_'.@a-z0-9-]+}")`. See #816 for more info.

**Sample Usage**
```
@FeignClient("service") // Can be a regular URL
public interface ServiceClient {

    @RequestMapping("/api/{symbolicName:[a-z-]+}-{version:\\d\\.\\d\\.\\d}{extension:\\.[a-z]+}")
    ResponseEntity<Model> fetch(@PathVariable("symbolicName") String symbolicName,
        @PathVariable("version") String version, @PathVariable("extension") String extension);

}
```

Please review and pull. I've signed the CLA.

Thanks,
Venil Noronha

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
@@ -19,7 +19,6 @@ package org.springframework.cloud.netflix.feign.support;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import java.lang.reflect.Parameter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -36,10 +35,10 @@ import org.springframework.cloud.netflix.feign.annotation.RequestParamParameterP
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.DefaultParameterNameDiscoverer;
+import org.springframework.core.MethodParameter;
 import org.springframework.core.ParameterNameDiscoverer;
-import org.springframework.core.annotation.AnnotatedElementUtils;
-import org.springframework.core.annotation.AnnotationAttributes;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.core.annotation.SynthesizingMethodParameter;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
 import org.springframework.core.io.DefaultResourceLoader;
@@ -81,6 +80,7 @@ public class SpringMvcContract extends Contract.BaseContract
 	private final ConversionService conversionService;
 	private final Param.Expander expander;
 	private ResourceLoader resourceLoader = new DefaultResourceLoader();
+	private final ParameterNameDiscoverer parameterNameDiscoverer = new DefaultParameterNameDiscoverer();
 
 	public SpringMvcContract() {
 		this(Collections.<AnnotatedParameterProcessor> emptyList());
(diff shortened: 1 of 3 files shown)


venilnoronha:
Build is failing now due to some heap space error.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
```diff
@@ -144,6 +154,25 @@ public MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method metho
 		return md;
 	}
 
+	protected void parseRegexAndUpdatePath(MethodMetadata md, Method method) {
```
spencergibb: Can you walk me through what this method is doing and why?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
```diff
@@ -144,6 +154,25 @@ public MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method metho
...
+			for (int i = 0; i < paramCount; i++) {
+				MethodParameter param = new SynthesizingMethodParameter(method, i);
+				param.initParameterNameDiscovery(parameterNameDiscoverer);
+				if (param.hasParameterAnnotation(PathVariable.class)) {
+					PathVariable pathVariable = param.getParameterAnnotation(PathVariable.class);
+					String varName = pathVariable.value();
+					url = url.replaceAll("\\{" + varName + ":.+?\\}", "{" + varName + "}");
+				}
+			}
+			Field urlField = ReflectionUtils.findField(RequestTemplate.class, "url");
```
spencergibb: This worries me a bit. Without having looked myself, is there a way to  do this without reflection?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
```diff
@@ -144,6 +154,25 @@ public MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method metho
...
 
+	protected void parseRegexAndUpdatePath(MethodMetadata md, Method method) {
+		if (isAnnotated(method, RequestMapping.class)) {
+			String url = md.template().url();
+			int paramCount = method.getParameterTypes().length;
+			for (int i = 0; i < paramCount; i++) {
+				MethodParameter param = new SynthesizingMethodParameter(method, i);
+				param.initParameterNameDiscovery(parameterNameDiscoverer);
+				if (param.hasParameterAnnotation(PathVariable.class)) {
+					PathVariable pathVariable = param.getParameterAnnotation(PathVariable.class);
```
spencergibb: We went through steps to remove reference to `PathVariable` from this class. See `PathVariableParameterProcessor`.

spencergibb:
Tell me if I'm wrong, but this doesn't 'support' regex's in path as ignores them. Am I correct?

venilnoronha:
Yes. I thought the use case was to use a common `interface` for the `@Controller` and the `@FeignClient`, which one wasn't able to do `@FeignClient` with regex in path would throw a `IllegalArgumentException`. That's the reason I thought of ignoring the regex part of the path.

If this is not what we're looking at I can rework this.

dsyer:
One of the goals of Spring Cloud Netflix is *not* sharing an interface between a `@Controller` and a `@FeignClient`. So unless we can think of a way to actually support regex on the feign side, I don't see much advantage in adding this feature.

spencergibb:
Closing per our comments. @venilnoronha please feel free to open a new PR if you want to support regex's, not ignore them.

Issue #1023, opened 2016-05-12
SpringMvcContract does not support path parameters defined in interface's @RequestMapping annotation


jebeaudet:
In `SpringMvcContract`, the path defined in the interface's class `@RequestMapping` annotation is prepended at the end of the `parseAndValidateMetadata(Class<?> targetType, Method method)` method. While this works fine for most cases, it does not support path parameters defined there that apply to all methods. This is because the `PathVariableParameterProcessor` expect it to be in the `MethodMetadata` at the time or it adds it as a form param as a fallback.

Example of a non working interface at the moment: 
```
	@RequestMapping("/prepend/{anotherId}")
	public interface Example {
		@RequestMapping(value = "/test/{id}", method = RequestMethod.GET)
		ResponseEntity<TestObject> getTest(@PathVariable String anotherId, 
                                           @PathVariable String id);

		@RequestMapping(method = RequestMethod.POST)
		TestObject postTest(@PathVariable String anotherId, 
                            @RequestBody TestObject object);
	}
```
I'll submit a PR shortly! Thanks again for the great lib.

spencergibb:
Using @RequestMapping on an interface level is not supported because of the way spring MVC works. It assumes they are real controllers. There is a path attribute on @FeignClient.

jebeaudet:
Hmm, we're not using `@FeignClient` here because we're generating clients with the Feign.Builder() to use in some functional tests. 

We're using the pattern defined in the documentation with one interface defining the resource and a controller implementing it and a client extending the interface in separate projects. AFAIK, using annotation on the interface worked properly when implementing it in our controller but I'll have to verify that with my colleague tomorrow. You're saying that this shouldn't work in the first place is that it?

jebeaudet:
We're implementing the pattern defined here http://cloud.spring.io/spring-cloud-static/spring-cloud.html#spring-cloud-feign-inheritance

It's working well having all the @RequestMapping annotations on the shared interface with the client extending the interface and the resource implementing the interface. The only thing not working is the support for query param on the class lever @RequestMapping annotation which this PR provides a fix.

PR #1024, opened 2016-05-12
SpringMvcContract does not support path parameters defined in interface's @RequestMapping

jebeaudet:
Proposed fix for #1023.

I've signed the contributor's agreement. Thanks!

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
@@ -107,6 +107,26 @@ public class SpringMvcContract extends Contract.BaseContract
 	}
 
 	@Override
+	protected void processAnnotationOnClass(MethodMetadata data, Class<?> clz) {
+		if(clz.getInterfaces().length == 0) {
+			RequestMapping classAnnotation = findMergedAnnotation(clz,
+					RequestMapping.class);
+			if (classAnnotation != null) {
+				// Prepend path from class annotation if specified
+				if (classAnnotation.value().length > 0) {
+					String pathValue = emptyToNull(classAnnotation.value()[0]);
+					pathValue = resolve(pathValue);
+					if (!pathValue.startsWith("/")) {
+						pathValue = "/" + pathValue;
+					}
+					data.template().insert(0, pathValue);
+				}
+			}
+		}
+    }
+	
+	
+	@Override
 	public MethodMetadata parseAndValidateMetadata(Class<?> targetType, Method method) {
 		this.processedMethods.put(Feign.configKey(targetType, method), method);
 		MethodMetadata md = super.parseAndValidateMetadata(targetType, method);
@@ -114,19 +134,6 @@ public class SpringMvcContract extends Contract.BaseContract
 		RequestMapping classAnnotation = findMergedAnnotation(targetType,
 				RequestMapping.class);
 		if (classAnnotation != null) {
-			// Prepend path from class annotation if specified
-			if (classAnnotation.value().length > 0) {
-				String pathValue = emptyToNull(classAnnotation.value()[0]);
-				checkState(pathValue != null,
-						"RequestMapping.value() was empty on type %s",
-						method.getDeclaringClass().getName());
-				pathValue = resolve(pathValue);
-				if (!pathValue.startsWith("/")) {
-					pathValue = "/" + pathValue;
-				}
-				md.template().insert(0, pathValue);
-			}
-
 			// produces - use from class annotation only if method has not specified this
 			if (!md.template().headers().containsKey(ACCEPT)) {
 				parseProduces(md, method, classAnnotation);
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/support/SpringMvcContractTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/support/SpringMvcContractTests.java
@@ -93,6 +93,56 @@ public class SpringMvcContractTests {
 
 		assertEquals("id", data.indexToName().get(0).iterator().next());
 	}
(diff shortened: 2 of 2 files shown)


dsyer:
I had to revert this because there were test failures. If you can rebase and fix them, that would be great.

spencergibb:
My fear (see https://github.com/spring-cloud/spring-cloud-netflix/issues/1023#issuecomment-218629178) is that this is just not compatible with spring cloud netflix. I just haven't had time to look.

jebeaudet:
I'll take a look at it, sorry about that, I was sure I ran the tests.

jebeaudet:
Hmm, `mvn clean install` completes succesfully with maven 3.3.9 and openjdk 1.8_91 on ubuntu 14.04 LTS. Same with `./mvnw clean install`. Do you have any input on this? I see that the travis build fails with an OOM error so I'm not sure where to look.
Thanks

dsyer:
Hmm. I was getting test failures in ribbon bits.

Travis is failing because of some weird javadoc thing. I would ignore it for now.

jebeaudet:
Well, I had to disable the `Build Automatically` in eclipse while building but that's pretty much it.

jebeaudet:
Did you have the chance to take a second look at it? Thx :+1:

spencergibb:
Wondering if you can rework the tests to leave the `_Simple` tests as they are and add a new set with `@RequestMapping` on the interface. That way we know you haven't broken existing functionality.

jebeaudet:
Definitely, that's a good suggestion. I'll work on that next week.

jebeaudet:
PR updated. Do you prefer squashed commits or you'd rather do it on your end? Thx again

dsyer:
So @spencergibb, you're not worried about compatibility any more?

@jebeaudet if you squash them it will save us time. They get squashed anyway usually, but most people don't seem to know how.

jebeaudet:
Squashed.

spencergibb:
@dsyer yes I am, see #1079

jebeaudet:
OK I see, does extending the interface for your client as mentioned in http://cloud.spring.io/spring-cloud-static/spring-cloud.html#spring-cloud-feign-inheritance would help?

spencergibb:
I don't think so. Spring MVC automatically creates controllers if a bean has `@RequestMapping` at the class level.

jebeaudet:
I'll give it a shot with @FeignClient, as I said in the issue, we're not using it. Thanks for the hint, it might not be possible after all.

jebeaudet:
I may be missing something but I've made a sample in https://github.com/jebeaudet/feign-client-sample reproducing exactly the pattern defined in http://cloud.spring.io/spring-cloud-static/spring-cloud.html#spring-cloud-feign-inheritance with a class level `@RequestMapping` with a path parameters. The only thing I've added is that the resource implementing the api also needs the `@PathVariable` annotations.

The repo should be self contained, I've used the Brixton release and I've created the `SpringMvcContract` class with the code from this PR in the same package as the dependency so my class has precedence.

To test, spawn the eureka server, run `UsersClientApplication` and `UsersResourceApplication`. The resource expose an endpoint on `http://localhost:8081/organizations/{organizationId}/users/{username}` which simply prints out the path parameters. The client expose an endpoint on `http://localhost:8080/` that simply proxy the resource endpoint with the path parameters `hello` and `from feign client proxy`.

I tried reproducing the issue mentioned in the link you gave @spencergibb without success, even when using `@RequestMapping` over `@FeignClient` (though I don't see why someone would do that), it didn't work but it didn't throw. I've tried with both my modified `SpringMvcContract` and the current one. 

Thanks again!

dsyer:
I think the problem @spencergibb was worried about occurred when people put `@RestController` on their feign interfaces (which is clearly wrong but some people want us to support it). I think a `@RequestMapping` at the class level is fine. Isn't that what this PR is partly about?

jebeaudet:
Yes, this PR is about supporting `@PathVariable` defined in the class' `@RequestMapping` (I just noticed that I've used query string all along when I meant path parameters, I'll update the commit message to be clearer).

I'm struggling to reproduce the problem mentioned in the linked issue. Also, since `@RequestMapping` is already supported at the interface level to prepend a constant path, I don't see how this PR breaks anything (see my explanation in the first comment of #1023), it just needs to be in the `MethodMetadata` before the `parseAndValidateMetadata` is called.

I'd be happy to dig around a little more but I'm not sure where to look next.

dsyer:
I think we can probably merge this as long as the tests are fine again. Spencer can weigh in when he gets a chance.

spencergibb:
This is what we've had problems with before https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/RequestMappingHandlerMapping.java#L169-L172

If you share an interface annotated with a top level `@RequestMapping` between feign and a real `@Controller` in the same project, spring will try and create a controller out of it even if it is not annotated with `@Controller`.

jebeaudet:
OK so the client and the resource have to be in the same project which doesn't make any sense to me, why would you use a client to access a resource in the same application? Even if you have multiple application instances, it's most likely always better to directly use the service than to go through an http request...

I just want to point out that : 
- class level `@RequestMapping` is already supported in `SpringMvcContract`, only the path params defined in it are not, are we planning in deprecating the support for it?
- Separating the interface+client+resource in different projects with correct dependency management (as shown in my example) works fine.

I don't really mind if we drop this PR, we already use our custom `SpringMvcContract` to build our feign clients. It's not blocking for us but I think this improvement, if the class level `@RequestMapping` is to be still supported in future releases, should make its way into the codebase to support path params.

Once again, thanks for your inputs :+1:

spencergibb:
I agree it doesn't make sense :-). I think this is ok to go.

jebeaudet:
Great news, happy to contribute!

spencergibb:
Closed with a little polish (formatting) via b3ef770819e121dfbbbda371b3e7fe6ecf587b96

Issue #1037, opened 2016-05-20
POST request without Content-Type is broken when using SimpleHostRoutingFilter


shanman190:
I've found that commit 053296d introduces a bug in how Content-Type is handled when using the SimpleHostRoutingFilter of a Zuul proxy. The line of code in question doesn't take into account that the Content-Type can be null and accidentally creates a Content-Type value of the string null (eg "null"). This when sent to another spring-based service causes an InvalidMediaTypeException to be thrown due to the lack of it being null or having a '/' character within the string value. This occurs in the latest Brixton release and the latest snapshot release.

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java#L260

PR #1038, opened 2016-05-20
Update SimpleHostRoutingFilter.java

shanman190:
Check if request Content-Type header is null before creating a ContentType object.

Fixes gh-1037

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -184,6 +184,16 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-security</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.security.oauth</groupId>
+			<artifactId>spring-security-oauth2</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
 			<scope>test</scope>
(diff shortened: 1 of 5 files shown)


spencergibb:
how about a test case?

spencergibb:
Have you signed the [Contributor License Agreement](https://support.springsource.com/spring_committer_signup)?

shanman190:
Yeah, I can write up a quick test case for it. I've signed one of the CLAs before, I think it was in relation to Spring Boot, but I can do it again if need be.

spencergibb:
Nope, once is enough.

shanman190:
Alright, so looking closer at this change I'm beginning to wonder how valuable such a test would be as I'd have to be getting into the internals of the class just to cause the reported behaviour. (eg Reflection to overlay the http client that's hidden within the class itself and building up almost all of an HTTP Request) Unless you guys have some other suggestions for testing this guy I'm not sure how much I can manage here...

spencergibb:
I'm unsure why you would use reflection for the test case. We do many tests where we send real requests to zuul.

shanman190:
Yeah. I was trying to write the wrong kind of test. I was thinking a unit test rather than an integration test. I think that I can create a test that causes the behavior described in the issue. But it will take me some time to do that. The error is a combination of the addition of the header with the string value of null in combination with using an OAuth protected resource and having an expired OAuth token. Both services are running Zuul in my case. The first one is running a UI application and the second one is an API application proxying downstream services. The UI application adds the Content-Type header with an invalid value from the perspective of the MediaType class. The DefaultOAuth2ExceptionRenderer class makes a call to the getHeaders method of the ServletServerHttpRequest class which then tries to get the Content-Type cascading eventually down to where the exception is being thrown.

shanman190:
Alright. I've gotten two test cases written around the error. The first is a successful request, get an OAuth token then call the protected endpoint, and the second is the failure test, send an invalid OAuth token and make sure that a 401 is returned instead of a 500.

shanman190:
It appears that @dsyer fixed the issue and with a much simpler test case, so I'm not sure if this PR is still beneficial anymore. His solution is probably better because it avoids the need for bringing spring security into the test context.

Issue #1030, opened 2016-05-15
Zuul Proxy: Custom RouteLocator must inherit from DiscoveryClientRouteLocator


droy314:
When trying to inject a custom RouteLocator, the ZuulProxyConfiguration requires the instance to be a DiscoveryClientRouteLocator.

The logged error message is:
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'zuulEndpoint' defined in class path resource [org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration$RoutesEndpointConfiguration.class]: Unsatisfied dependency expressed through constructor argument with index 0 of type [org.springframework.cloud.netflix.zuul.filters.discovery.DiscoveryClientRouteLocator]: No qualifying bean of type [org.springframework.cloud.netflix.zuul.filters.discovery.DiscoveryClientRouteLocator] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type [org.springframework.cloud.netflix.zuul.filters.discovery.DiscoveryClientRouteLocator] found for dependency: expected at least 1 bean which qualifies as autowire candidate for this dependency. Dependency annotations: {}
```
```
ZuulProxyConfiguration.RoutesEndpointConfiguration expects a DiscoveryClientRouteLocator to inject the zuul RoutesEndpoint. 
```
Changing the hierarchy to use DiscoveryClientRouteLocator works. Do we really need a DiscoveryClientRouteLocator instead of a RouteLocator?

dsyer:
I guess we do (or did you change it and it still worked?). Is that a problem?

droy314:
Okay... I realize now I should have given more information.The request here is to change the argument type in <code>RoutesEndpointConfiguration</code>, where the <code>RoutesEndpoint</code> is setup, from <code>DiscoveryClientRouteLocator</code> to <code>RouteLocator</code>. [source code](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyConfiguration.java#L135-L142)

[...]

PR #1040, opened 2016-05-21
Fixed gh-1030

droy314:
#1030


droy314:
The change here works..however, I needed to check with my employer on the contributor agreement. I created the pull request before I read that section.

Issue #1053, opened 2016-05-26
Expose security context to any Hystrix command.


daniellavoie:
This issue is a follow up from Spring Cloud Security issue [#87](https://github.com/spring-cloud/spring-cloud-security/issues/87#issuecomment-220549541). I came to the conclusion that the request feature was Hystrix focused rather than Spring Cloud Security. Spring Cloud Security is aimed towards OAuth2 support.

Hystrix commands are actually unable to access the security context if executed with an isolation strategy set to `THREAD`. Only workaround available consists in setup the execution isolation to SEMAPHORE. 

Using Hystrix command hooks, we could build an infrastructure to transfert the Security Context from the main thread to the one executing the Hystrix command.

spencergibb:
Closed via 36fec07a17de7e54894bb343ff97c009b48a5320

PR #1054, opened 2016-05-26
Expose security context of a feign client to an hystrix command.

daniellavoie:
PR for issue #1053 

Currently leverages on Command Hook registration of Hystrix. With this current implementation, users won't be able to register a custom Hook. Needs to be addressed. Waiting for project leads feedback.

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -36,6 +36,16 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-security</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
 			<optional>true</optional>
 		</dependency>
@@ -172,11 +182,6 @@
 			<!-- Only needed at compile time -->
 			<scope>compile</scope>
 			<optional>true</optional>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-starter-test</artifactId>
-			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/CustomCommandHook.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/CustomCommandHook.java
@@ -1,0 +1,7 @@
+package org.springframework.cloud.netflix.feign.hystrix;
+
+import com.netflix.hystrix.HystrixCommand;
+
+public interface CustomCommandHook {
+	<T> void onRunStart(HystrixCommand<T> commandInstance);
+}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/HystrixRequestContextEnablerFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/HystrixRequestContextEnablerFilter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.feign.hystrix;
+
+import java.io.IOException;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netflix.hystrix.strategy.concurrency.HystrixRequestContext;
+
+/**
+ * @author Daniel Lavoie
+ */
+public class HystrixRequestContextEnablerFilter implements Filter {
+	private static Logger LOGGER = LoggerFactory
+			.getLogger(HystrixRequestContextEnablerFilter.class);
+
+	@Override
+	public void doFilter(ServletRequest request, ServletResponse response,
+			FilterChain chain) throws IOException, ServletException {
+		if (LOGGER.isTraceEnabled())
+			LOGGER.trace("Initializing Hystrix Request Context.");
+
+		HystrixRequestContext context = HystrixRequestContext.initializeContext();
+		try {
+			chain.doFilter(request, response);
+		}
+		finally {
+			context.shutdown();
+		}
+	}
+
+	@Override
+	public void init(FilterConfig filterConfig) throws ServletException {
+	}
+
+	@Override
+	public void destroy() {
+	}
+}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/FeignHystrixSecurityAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/FeignHystrixSecurityAutoConfiguration.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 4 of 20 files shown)


daniellavoie:
I'm actually trying to figure out why the RestTemplateRetryTest test only crashes during maven build. I'm not encountering any error when running the full JUnit suite. Any guess ?

spencergibb:
I get the same error on master.

daniellavoie:
Yeap, same here. I'm also surprised to see my new test on error. Going to dig on it but I don't have the same behavior than travis when running maven install.

spring-cloud-netflix-core/pom.xml
```diff
@@ -180,6 +185,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-security</artifactId>
```
dsyer: You shouldn't need both this and spring-security-core

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/HystrixRequestContextEnablerFilter.java
```diff
@@ -0,0 +1,62 @@
...
+
+	@Override
+	public void doFilter(ServletRequest request, ServletResponse response,
+			FilterChain chain) throws IOException, ServletException {
+		if (LOGGER.isTraceEnabled())
+			LOGGER.trace("Initializing Hystrix Request Context.");
+
+		HystrixRequestContext context = HystrixRequestContext.initializeContext();
+		try {
+			chain.doFilter(request, response);
```
dsyer: This doesn't look like it is security related. Can you explain what is happening here and why?

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/hystrix/TestInterceptor.java
```diff
@@ -0,0 +1,48 @@
...
+	private String password;
+
+	@Override
+	public void apply(RequestTemplate template) {
+		if (SecurityContextHolder.getContext().getAuthentication() != null)
+			template.header("username",
+					SecurityContextHolder.getContext().getAuthentication().getName());
+
+		template.header("Authorization", "Basic " + new String(
+				Base64.getEncoder().encode((username + ":" + password).getBytes())));
```
dsyer: I'm not sure what this is doing yet. There are 2 usernames in play here, one from the security context and one that was injected as a `@Value`. What is the header for?

dsyer:
If hooks are singular (as stated in https://github.com/spring-cloud/spring-cloud-security/issues/87), we would need an extension mechanism, so users can add their own hooks. Would they work as a composite (and the user can add one to the composite by configuring a bean)?

I'm not sure I understand the use case for this one yet though. What is there in the `SecurityContext` that is useful in a Feign interceptor? The tests show a header being sent with the current user name, but it seems somewhat artificial.

daniellavoie:
I'll give you that the use case is not quite obvious. Let me explain it to you.

I am interested in having access to the `SecurityContext` inside a Feign Interceptor so I can automatically inject custom headers or parameters required a backend service. In my case, the backend targeted by Feign requires a username inside the header for audit purpose for every endpoints. Rather than having to define the username as a parameter to every Feign endpoints, I prefer having a centralized interceptor for this.  This backend is secured with generic credential which is unrelated to the user logged inside my `Security Context`. 

The test I contributed tries to reproduce this use case inside a single Web Service. 

- `TestController` represent a backend service which requires the username as part of the header.
- `ProxyTestController` represent my service secured endpoint who uses Feign to call the backend service. It uses the Feign `TestInterceptor` to automatically inject the username header from the `SecurityContext`. 
- `TestInterceptor` injects the `SecurityContext` username. At the moment, it also inject a Basic Auth Header because the backend url is secured by the same filter than `ProxyTestController`. I'll change that. For my demonstration, the backend service does not need to be secured.

I'll go through your reviews comments and reply to them.

Hope I could be helpful with those explanations.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/hystrix/TestInterceptor.java
daniellavoie: There is indeed 2 username in play. In my test purpose, the targeted Feign url uses the same security filter than the one who makes the Feign call. I realized that for this test, it doesn't need to be secured as I am only interested in showing that the username header gets injected properly.

In a real life scenario, this interceptor could inject an Auth header that is specific for the targeted backend. For exemple, a legacy backend requiring an generic username password.

I'l rework the test case to remove the security on the backend service.

spring-cloud-netflix-core/pom.xml
daniellavoie: You are right, I didn't realized `spring-cloud-commons`was already referencing the spring security libs. Thanks for pointing.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/HystrixRequestContextEnablerFilter.java
daniellavoie: I'll just quote the excellent Juan Naverrete [post](https://jfconavarrete.wordpress.com/2014/09/15/make-spring-security-context-available-inside-a-hystrix-command/). Basically, it initialize the holder which allow the SecurityContext to be request scoped rather than thread scoped. It's the little box where you'll store the SecurityContext.

> The HystrixRequestContext must be initialized at the beginning of each request before RequestVariable functionality can be used. This enables the HystrixRequestContext to contain the state and manage the lifecycle of HystrixRequestVariableDefault objects that provide request scoped (rather than only thread scoped) variables so that multiple threads within a single request can share state.

daniellavoie:
In the mean time, I am going to think about the cleanest extension mechanism. 

Difficulty resides in the fact that users might expect to directly registers their hook with Hystrix API without knowing about our pre registered one. A runtime exception with explanations could be triggered if we detect that our Spring Cloud hook was removed. Our hook could then make use of an optional bean providing extension capabilities. What do you think about it ?

dsyer:
I think if people want to use the raw Netflix API then they will forgo the opportunity to benefit from our customizations (i.e. don't worry about it).

daniellavoie:
Alright then ! I'll just at least provide provide an extension bean. By the way, how do you feel about this feature ? Have I made myself more clear on the use case ? If you are interested into integrating that feature, I'll prepare the doc into `spring-cloud-netflix-docs`.

dsyer:
I'm not sure yet. Probably something like this makes sense. I'm a little wary of adding 2 new filters to the stack, but maybe we can find a different way, or make them optional somehow, once we have the feature clearly defined.

daniellavoie:
I understand your concerns with the new filters. I'll think about other alternatives. I'm open to suggestion if you have any idea :) I'm guessing aop could be another way around but I'm not a big fan of overusing aspects. In the meantime. I could deactivate the filters by default and make them optional with a property. 

Since this feature answer an edge case, deactivating it by default would make sense.

daniellavoie:
By the way, I am currently investigating the issue with the RestTemplateRestryTest. The problem comes from the retry handler of the ribbon configuration who gets overwritten when running the full maven build. the tests sets up a next server retry count to 10 but gets overwritten by the default handler. I'm close to finding out why it happens. I'll push try to push a PR once I identified to source. This will help to fix the travis build.

Regards

daniellavoie:
I am getting desperate with that PR :( I'll setup a travis build on my personal repo so I can stop polluting this PR. 

**INSERT GENERIC WORKS ON MY PC COMMENT**

spencergibb:
@daniellavoie have you rebased with master?

daniellavoie:
@spencergibb Yeap, I double and triple checked. Since you fixed master I am now able to successfully build locally. I'll play with travis and see what's wrong. My unit test is very elaborate. I'm making some web service-ception so that I can trigger the a feign client with a security context filed by Spring Security Filter.

daniellavoie:
I'm kinda lost. I have a travis build on my personal repo. My commit [be60b08](https://github.com/daniellavoie/spring-cloud-netflix/commit/be60b08c01fddd2c95b14282ddec563215048713) was built successfully in my [repo](https://travis-ci.org/daniellavoie/spring-cloud-netflix/builds/133679890). The force push of this commit to my PR branch triggered the spring cloud netflix travis build [1301](https://travis-ci.org/spring-cloud/spring-cloud-netflix/builds/133679894) who resulted in a merge with master (even tough the commit was already rebased on master). My personal travis builds without problem but the Spring Cloud travis instance crashes. I'm speechless :/ It feels like there is something I am not understand with Travis. Any idea ? Many thanks. Sorry for disturbing :/

dsyer:
Travis is a bit mysterious sometimes. Don't worry about it.

daniellavoie:
Victory ! I've been encountering some weird behavior on properties defined with `@WebIntegration`. So I moved them to `application.yml`. Can't tell why but Travis gave me different behavior than locally. 

I implemented the extendable hook mechanism. Users only need to declare a bean implementing `CustomCommandHook`.

daniellavoie:
I'll rework that PR with the proper documentation within asciidoc sources.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/hystrix/CustomCommandHook.java
```diff
@@ -0,0 +1,7 @@
+package org.springframework.cloud.netflix.feign.hystrix;
+
+import com.netflix.hystrix.HystrixCommand;
+
+public interface CustomCommandHook {
```
shuraa: I think you should add javadoc here.

daniellavoie: I agree, I'll add the javadoc on my next PR which will also inclode the asciidoc update.

daniellavoie:
I actually realized that this feature does not limit itseft to a feign hystrix command. It allows you to access the `SecurityContext` from any Hystrix Command. I'll rework PR description and package naming. Many people are complaning about not being able to access `SecurityContext` without using `SEMAPHORE` isolation.

daniellavoie:
I'm closing this PR since I had to renamed the branch. Replaced by PR #1093 .

PR #1058, opened 2016-05-27
Fix RestTemplateRetryTest Unit test

daniellavoie:
The metrics tests caused a concurrency issue with the RestTemplateRetryTest. The metrics interceptor was overwriting the retry handler defined in the ribbon RestTemplateRetryTest. I assume that the metrics interceptor detected the `badClient` and redefinned with the ribbon properties available his scope. 

I moved the maxRetry properties to the application.yml so those properties would be common for the interceptor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ before_install:
   - git config credential.helper "store --file=.git/credentials"
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
   - gem install asciidoctor
+  - export MAVEN_OPTS="-Xmx1g"
 install:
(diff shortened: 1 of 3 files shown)


daniellavoie:
The Netflix Core is fixed. Build is now crashing further at eureka server.

spencergibb:
Merged with a little polish via 16e827f3ba2d831a45148da4064308c9ae6e08e3

spencergibb:
thanks!

spencergibb:
FYI [this](https://issues.apache.org/jira/browse/MSOURCES-94) is the cause of the build error (which only happens on travis btw).

daniellavoie:
Thanks for the heads up ! I've been scratching my head on this one. Any other issue you guys need some help ?

PR #1059, opened 2016-05-28
Update SpringMvcContract to support placeholders in headers

acennan:
When using property placeholders in the `@RequestMapping` annotation the ones placed in the `headers` attribute do not get replaced. This appears to be because the other attributes, such as the `url`, use the `resolve()` function to perform the substitution whereas the header values are parsed as-is. This change adds a call to resolve for each side of the equals in the header string so that the properties are substituted correctly.


spencergibb:
Have you signed the [Contributor License Agreement](https://support.springsource.com/spring_committer_signup)?

acennan:
Hi Spencer,

Sorry, forgot to do that earlier so have just done it now.

Damian.

spencergibb:
Merged with a test via 711f095ebc2093bce4e8f3f25ab0e71c4dcf544c

Issue #1045, opened 2016-05-24
Upgrade to ribbon v2.2.0


spencergibb:
https://github.com/Netflix/ribbon/releases/tag/v2.2.0

aivans:
@spencergibb do you have some reference for using this new feature? 
```
7eef72f33e3585ba052c35429a96539afaffe66b: (take 2 of) Adding eureka aware Loadbalancers that make use of new notification mechanism from eurekClient
```
Does it help in faster propagation of Eureka information from Eureka Client to Ribbon? My understanding now is that Ribbon will poll Eureka client every 30s.

spencergibb:
@aivans no I don't

PR #1061, opened 2016-05-29
Upgrade ribbon to 2.2.0

daniellavoie:
Fix gh-1045

2.2.0 seems to provide a new mechanism to handle eureka notifications. I don't think Spring Cloud need to wrap this feature since users can directly use the ribbon API.


spencergibb:
Closed via 1e52fe274940d61854953549c95e1eb10fcf7e8f

PR #1071, opened 2016-06-01
Fix the path encoding when character encoding is set on HttpServletRequest

### LINKED ISSUE #647: Zuul proxy is not working with utf-8 urls
I have a spring cloud app with Zuul proxy. My zuul proxy doesn't work with UTF-8 urls. How should I configure it?

This will work in my microservice, without proxy: ```http://localhost:8080/rest/item/гандан_тикет``` However when its behind a proxy ```http://localhost:9000/core/rest/item/гандан_тикет```
the гандан\_тикет will come as ```??????_?????``` into the microservice.

jebeaudet:
This fixes #647. I've already signed the contributor's agreement. Thanks!


spencergibb:
Merged via fc08b55eaca983d865281d85cb0bc73a75bd1710. @jebeaudet similar fix as #979, but I wanted to include your tests. Thanks again!

Issue #927, opened 2016-03-25
Ribbon load balancer configurable in yml/properties file, and also support spring cloud config


iamzhout:
Currently only [part of the ribbon configurations](https://github.com/Netflix/ribbon/wiki/Getting-Started) can take effect in configuration file, while not supporting [load balancer rules](https://github.com/Netflix/ribbon/wiki/Working-with-load-balancers), is it possible to also make these lb rules configurable and also support Spring Cloud Config? or is it possible to tell the way how to make the change, and I can work on this.

the items supposed to be supported including:
> NFLoadBalancerClassName
> NFLoadBalancerRuleClassName
> NFLoadBalancerPingClassName
> NIWSServerListClassName
> NIWSServerListFilterClassName

OrangeDog:
+1
I'm hoping this just requires configuring archaius to delegate to the spring application environment.

OrangeDog:
While we're there, it would be very useful if sping-cloud would support a default property prefix to be shared by all ribbon clients.

OrangeDog:
Documentation claims that this already works:

> Spring Cloud has a Spring Environment Bridge so Archaius can read properties from the Spring Environment. This allows Spring Boot projects to use the normal configuration toolchain, while allowing them to configure the Netflix tools, for the most part, as documented.

Must be some other reason it's not working.

spencergibb:
"for the most part" ;-) @jkschneider asked for this yesterday as well.

iamzhout:
@spencergibb so does this mean you will have plan to support this recently?

spencergibb:
@iamzhout it means we plan to support it sometime in the future.

OrangeDog:
Is it because the `NFFooClassName` properties are only used if they are omitted from the builder, but `RibbonClientConfiguration#ribbonRestClient` ensures that they exist and are set?

iamzhout:
@spencergibb Is there any update on this feature implementation plan? 
@OrangeDog I have checked, and seems it has no relation to  `RibbonClientConfiguration#ribbonRestClient`? as I am busy on other things recently, I will check this later.

OrangeDog:
As an extreme, if you declare this:

    @Bean 
    public ILoadBalancer ribbonLoadBalancer(IClientConfig config) {
        return LoadBalanceBuilder.newBuilder()
            .withClientConfig(config)
            .buildLoadBalancerFromConfigWithReflection();
    }

[...]

PR #1090, opened 2016-06-09
Load ribbon classes if defined via properties.

spencergibb:
Allows ribbon class name properties to be used to define beans.

This seems like boilerplate, but works, though not easily extensible.

fixes gh-927

/cc @dsyer @jkschneider @twicksell


spencergibb:
Closed via 7dc8b86001ea6cb32b7e70b46ea716b0b6dfd0a8

PR #1093, opened 2016-06-10
Expose security context to any Hystrix command.

daniellavoie:
Reworked PR from #1054. This feature let Hystrix command access Spring SecurityContext even when using `THREAD` isolation strategy.

Fix issue gh-1054

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -36,6 +36,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-security</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
 			<optional>true</optional>
 		</dependency>
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.hystrix.security;
+
+import javax.annotation.PostConstruct;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.cloud.netflix.hystrix.security.HystrixSecurityAutoConfiguration.HystrixSecurityCondition;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.core.context.SecurityContext;
+
+import com.netflix.hystrix.Hystrix;
+import com.netflix.hystrix.strategy.HystrixPlugins;
+import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy;
+import com.netflix.hystrix.strategy.eventnotifier.HystrixEventNotifier;
+import com.netflix.hystrix.strategy.executionhook.HystrixCommandExecutionHook;
+import com.netflix.hystrix.strategy.metrics.HystrixMetricsPublisher;
+import com.netflix.hystrix.strategy.properties.HystrixPropertiesStrategy;
+
+/**
+ * @author Daniel Lavoie
+ */
+@Configuration
+@Conditional(HystrixSecurityCondition.class)
+@ConditionalOnClass({ Hystrix.class, SecurityContext.class })
+public class HystrixSecurityAutoConfiguration {
+	@Autowired(required = false)
+	private HystrixConcurrencyStrategy existingConcurrencyStrategy;
+
+	@PostConstruct
+	public void init() {
+		// Keeps references of existing Hystrix plugins.
+		HystrixEventNotifier eventNotifier = HystrixPlugins.getInstance()
+				.getEventNotifier();
+		HystrixMetricsPublisher metricsPublisher = HystrixPlugins.getInstance()
+				.getMetricsPublisher();
+		HystrixPropertiesStrategy propertiesStrategy = HystrixPlugins.getInstance()
+				.getPropertiesStrategy();
+		HystrixCommandExecutionHook commandExecutionHook = HystrixPlugins.getInstance()
+				.getCommandExecutionHook();
+
+		HystrixPlugins.reset();
+
+		// Registers existing plugins excepts the Concurrent Strategy plugin.
+		HystrixPlugins.getInstance().registerConcurrencyStrategy(
+				new SecurityContextConcurrencyStrategy(existingConcurrencyStrategy));
+		HystrixPlugins.getInstance().registerEventNotifier(eventNotifier);
+		HystrixPlugins.getInstance().registerMetricsPublisher(metricsPublisher);
+		HystrixPlugins.getInstance().registerPropertiesStrategy(propertiesStrategy);
+		HystrixPlugins.getInstance().registerCommandExecutionHook(commandExecutionHook);
+	}
+
+	static class HystrixSecurityCondition extends AllNestedConditions {
+
+		public HystrixSecurityCondition() {
+			super(ConfigurationPhase.REGISTER_BEAN);
+		}
+
+		@ConditionalOnProperty(name = "feign.hystrix.enabled", matchIfMissing = true)
+		static class HystrixEnabled {
+
+		}
+
+		@ConditionalOnProperty(name = "hystrix.shareSecurityContext")
+		static class ShareSecurityContext {
+
+		}
+	}
+}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
@@ -48,7 +48,7 @@ public class ZuulProperties {
 	 * duplicated if the proxy and the backend are secured with Spring. By default they
 	 * are added to the ignored headers if Spring Security is present.
 	 */
-	private static final List<String> SECURITY_HEADERS = Arrays.asList("Pragma",
+	public static final List<String> SECURITY_HEADERS = Arrays.asList("Pragma",
 			"Cache-Control", "X-Frame-Options", "X-Content-Type-Options",
 			"X-XSS-Protection", "Expires");
 
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ZuulPropertiesTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ZuulPropertiesTests.java
@@ -46,7 +46,8 @@ public class ZuulPropertiesTests {
 
 	@Test
 	public void defaultIgnoredHeaders() {
-		assertTrue(this.zuul.getIgnoredHeaders().isEmpty());
+		assertTrue(this.zuul.getIgnoredHeaders()
+				.containsAll(ZuulProperties.SECURITY_HEADERS));
 	}
 
 	@Test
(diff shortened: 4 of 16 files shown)


dsyer:
LGTM. If we are going to schedule Camden this summer I think we should put this on master (1.2.0).

daniellavoie:
Glad I could help !

spring-cloud-netflix-core/pom.xml
```diff
@@ -36,6 +36,16 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-security</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<optional>true</optional>
```
spencergibb: Why did you change this from `scope:test` to `optional:true`?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixRequestContextEnablerFilter.java
```diff
@@ -0,0 +1,62 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netflix.hystrix.strategy.concurrency.HystrixRequestContext;
+
+/**
+ * @author Daniel Lavoie
+ */
+public class HystrixRequestContextEnablerFilter implements Filter {
+	private static Logger LOGGER = LoggerFactory
```
spencergibb: Should be commons logging.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,77 @@
...
+@Configuration
+@Conditional(HystrixSecurityCondition.class)
+@ConditionalOnClass({ Hystrix.class, SecurityContext.class })
+public class HystrixSecurityAutoConfiguration {
+	@Autowired(required = false)
+	private CustomCommandExecutionHook customCommandExecutionHook;
+
+	@PostConstruct
+	public void init() {
+		HystrixPlugins.reset();
```
spencergibb: This worries me a bit. If the user has set other plugins, they will be unceremoniously reset.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixRequestContextEnablerFilter.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.netflix.hystrix.strategy.concurrency.HystrixRequestContext;
+
+/**
+ * @author Daniel Lavoie
+ */
+public class HystrixRequestContextEnablerFilter implements Filter {
```
spencergibb: `HystrixRequestContextFilter`? I don't think `Enabler` helps.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/SecurityContextRegistratorCommandHook.java
```diff
@@ -0,0 +1,65 @@
...
+			.getLogger(SecurityContextRegistratorCommandHook.class);
+
+	private CustomCommandExecutionHook customCommandHook;
+
+	public SecurityContextRegistratorCommandHook(CustomCommandExecutionHook customCommandHook) {
+		this.customCommandHook = customCommandHook;
+	}
+
+	@Override
+	public <T> void onRunStart(HystrixCommand<T> commandInstance) {
```
spencergibb: `onRunStart` is deprecated, use `onExecutionStart`.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -48,7 +48,7 @@
 	 * duplicated if the proxy and the backend are secured with Spring. By default they
 	 * are added to the ignored headers if Spring Security is present.
 	 */
-	private static final List<String> SECURITY_HEADERS = Arrays.asList("Pragma",
+	public static final List<String> SECURITY_HEADERS = Arrays.asList("Pragma",
```
spencergibb: Why did you change this?

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ZuulPropertiesTests.java
```diff
@@ -46,7 +46,8 @@ public void teardown() {
 
 	@Test
 	public void defaultIgnoredHeaders() {
-		assertTrue(this.zuul.getIgnoredHeaders().isEmpty());
+		assertTrue(this.zuul.getIgnoredHeaders()
+				.containsAll(ZuulProperties.SECURITY_HEADERS));
```
spencergibb: Again, why was this changed?

spring-cloud-netflix-core/src/test/resources/application.yml
```diff
@@ -30,6 +30,9 @@ badClients:
   ribbon:
     MaxAutoRetriesNextServer: 10
     ReadTimeout: 200
+username:
+  ribbon:
+    listOfServers: localhost:${local.server.port}
```
spencergibb: Does this work? If so, it surprises me because I didn't think it was available in time.

spring-cloud-netflix-core/pom.xml
daniellavoie: Because the dependencies are required by the AutoConfiguration on compilation.

spencergibb:
How about a rebase from master as well.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixRequestContextEnablerFilter.java
daniellavoie: Sure, I'll fix it.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
daniellavoie: Adding Spring Security changes the headers injected by Zuul. `ZuulPropertiesTests.defaultIngoredHeaders` started to fail for this reason. I updated the tests to expect the headers declared with `ZuulProperties.SECURITY_HEADERS`.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/filters/ZuulPropertiesTests.java
daniellavoie: See my remark on `ZuulProperties.java`.

daniellavoie:
Thanks for the review ! I'll rebase and work on those remarks this weekend.

spring-cloud-netflix-core/src/test/resources/application.yml
daniellavoie: I got surprised too. But according to my unit tests it does. I'll double check when I'll rework the PR.

spring-cloud-netflix-core/pom.xml
spencergibb: That shouldn't be the case. If it is, something is broken.

spencergibb: I pulled your branch and tried reverting this particular change and it worked fine, so please revert.

spencergibb:
Thanks for all your work @daniellavoie!

spring-cloud-netflix-core/pom.xml
daniellavoie: Ah ! I miss read your review. I though you were asking why `spring-boot-starter-security` changed to optional:true. Changing `spring-boot-starter-test` to optional is indeed an error. I'll fix it.

daniellavoie:
Hi @spencergibb ! What is the target date for 1.2.0.M1 ? I'm currently in vacation so I finally have some free time to finish my work on this PR.

spencergibb:
After spring one, so August unless @dsyer says otherwise. I'm away from work as well.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
daniellavoie: Ok, I came around it. I'm calling reset because the test suite registers a dummy command execution hooks and blocks my tests since multiple hooks can't be registered. Since only one hook may be registered, I'm forcing the reset so my test can pass.

I am going to move this reset to my integration test. Users will simply have to understand that they won't be able to register a hook as they have the custom command hook mechanism. How do you feel with that @dsyer ?

spencergibb: Makes sense. You'll have to note that in documentation.

dsyer: We have a custom hook in sleuth. Probably need to look at that and make sure there is a way to add both.

daniellavoie: Thanks for the heads up @dsyer . 

Sleuth is actually calling `HystrixPlugins.reset` through `SleuthHystrixConcurrencyStrategy`, but @marcingrzejszczak is doing it properly. He keeps a reference of any existing custom plugin and re registers them after the reset. Sleuth is only interested in overwriting the concurrency strategy. I'll do the same in my own class before calling reset and throw an explanatory runtime exception in case the Command Hook is already defined. Both Hyxtris and Sleuth won't be impacted by any race condition.

pivotal-issuemaster:
@daniellavoie Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1093)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1093) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@daniellavoie Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1093)!

daniellavoie:
Got hit by travis mysteries once again. Build is fine now !

In the end, I decided to not implement some fancy runtime detection of command hook overwrite. I just stated clearly in the doc that any command hook would be overwritten and that the custom hook mechanism is available.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/SecurityContextRegistratorCommandHook.java
```diff
@@ -0,0 +1,65 @@
...
+			if (LOGGER.isTraceEnabled())
+				LOGGER.trace("Injecting security context.");
+
+			SecurityContextHolder.setContext(context);
+		}
+
+		if (customCommandHook != null)
+			customCommandHook.onExecutionStart(commandInstance);
+	}
+}
```
pukkaone: I just started using Hystrix, so this question may be a result of my ignorance. I'm curious why you implemented a command execution hook instead of a concurrency strategy that overrides the [wrapCallable](https://github.com/Netflix/Hystrix/wiki/Plugins#concurrencystrategy) method. For example:
```java
import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy;
import java.util.concurrent.Callable;
import org.springframework.security.concurrent.DelegatingSecurityContextCallable;

public class SecurityContextConcurrencyStrategy extends HystrixConcurrencyStrategy {
    @Override
    public <T> Callable<T> wrapCallable(Callable<T> callable) {
        return super.wrapCallable(new DelegatingSecurityContextCallable(callable));
    }
}
```
After the executing the callable, `DelegatingSecurityContextCallable` clears the security context from the thread.

daniellavoie: Your approach sounds interresting. I lost access to internet. I'll check out if the DelegatingSecurityContextCallable is more elegant than the command hook as soon the router recovers. Seems like he is on a hangover.

daniellavoie:
Ok, I've turned this PR around by using the `HystrixConcurrencyStrategy` plugin rather than the Command Hook. Result is less code and a better handling of the security context by using `DelegatingSecurityContextCallable`. Extension is still available in a friendlier manner simply by declaring a `HystrixConcurrencyStrategy` bean. Spring Cloud will detect it within the Spring Context and wrapping it inside it's own plugin. Doc has been adapted.

Feel free to share your thoughts on that rewrite. Thanks @pukkaone.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
```diff
@@ -0,0 +1,89 @@
...
+		HystrixMetricsPublisher metricsPublisher = HystrixPlugins.getInstance()
+				.getMetricsPublisher();
+		HystrixPropertiesStrategy propertiesStrategy = HystrixPlugins.getInstance()
+				.getPropertiesStrategy();
+		HystrixCommandExecutionHook commandExecutionHook = HystrixPlugins.getInstance()
+				.getCommandExecutionHook();
+
+		HystrixPlugins.reset();
+
+		// Registers existing plugins excepts the Command Hook.
```
spencergibb: Comment now out of date?

spencergibb:
I like the update. @dsyer or @ryanjbaxter any bandwidth to get this in before M1?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
daniellavoie: Yeap ! Will fix it right away :)

daniellavoie:
I'll rebase this PR in the evening.

daniellavoie:
@ryanjbaxter Wouhooo, thanks for merging. Finally I can stop maintaining that branch :) Let me know if you need any help on any other 1.2.0.M1 issue.

Issue #1091, opened 2016-06-10
Upgrade zuul to v1.2.2


spencergibb:
https://github.com/Netflix/zuul/releases/tag/v1.2.2
https://github.com/Netflix/zuul/releases/tag/v1.2.1
https://github.com/Netflix/zuul/releases/tag/v1.2.0

spencergibb:
The following are updates to groovy filters that aren't part of the library
https://github.com/Netflix/zuul/commit/f9a462b81946e288b8dce776702bb90e44a648b0
https://github.com/Netflix/zuul/commit/11ae37f5928a7c81e374bb0ba64e6ead91f665df
https://github.com/Netflix/zuul/commit/262271981a6001d2afaca69d7559b848d88bc5ef
https://github.com/Netflix/zuul/commit/1a24fc6d25b15400b378bb636fb723576aa025b6

PR #1106, opened 2016-06-15
Upgrade to zuul 1.2.2

spencergibb:
/cc @dsyer @marcingrzejszczak 

fixes gh-1091
closes gh-958


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendErrorFilter.java
```diff
@@ -76,6 +77,8 @@ public Object run() {
 				request.setAttribute("javax.servlet.error.message", message);
 			}
 
+			//unwrap request if needed, otherwise tomcat hangs in endless loop
```
spencergibb: This is what was causing the endless while loop.

spencergibb: Even after removing the send 404 to `/error` I still need to unwrap. I'm a bit concerned and wonder if anyone can comment.

The problem occurs in [`ApplicationDispatcher`]( https://github.com/apache/tomcat/blob/TOMCAT_8_0_0/java/org/apache/catalina/core/ApplicationDispatcher.java#L907-L908).

```
if (!(current instanceof ServletRequestWrapper))
                break;
```
The request is always a `ServletRequestWrapper` unless I unwrap it.

@markt-asf @wilkinsona @philwebb @snicoll any thoughts?

markt-asf: @spencergibb  Source code for the `ServletRequestWrapper` being used at that point?

spencergibb: @markt-asf https://github.com/Netflix/zuul/blob/1.x/zuul-core/src/main/java/com/netflix/zuul/http/HttpServletRequestWrapper.java

markt-asf: And there is the problem. getRequest() doesn't return the request that is wrapped by the current wrapper (i.e. req) it returns a new, wrapped version of req. That is a bug in zuul.

spencergibb: Thanks @markt-asf!

Issue #1126, opened 2016-06-23
RibbonLoadBalancerClient reconstructs scheme-less URIs into http://my-app.example.com:443 


william-tran:
I want to use a `@LoadBalanced` RestTemplate to access `MY-APP` without needing to know how `MY-APP` was registered, eg, insecure only, secure only, or both. To achieve this, I want to leave out the scheme in the URI, eg 
```
restTemplate.getForObject("//MY_APP/api", Object.class);
```
I would expect this URI to resolve to the secure route if available, and if not, the insecure route. But this URI only works when the registration is insecure only.

If there is a secure route, that port is used as the ServiceInstance's port:

https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.1.2.RELEASE/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/ribbon/eureka/DomainExtractingServerList.java#L66-L67
https://github.com/Netflix/ribbon/blob/v2.2.0/ribbon-eureka/src/main/java/com/netflix/niws/loadbalancer/DiscoveryEnabledServer.java#L43-L44

But, if there is no specific ribbon client configuration for that service, the scheme defaults to http:
https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.1.2.RELEASE/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancerClient.java#L56-L60

This is because the config object is never null, even when I don't have any explicit client configuration. And `config.get(CommonClientConfigKey.IsSecure, false)` defaults to false. 
https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.1.2.RELEASE/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancerClient.java#L117

So, without any explicit client config, we have a scheme that defaults to http but a port that defaults to the secure port. 

This can be fixed by using the InstanceInfo to determine `IsSecure` if there is no explicit client configuration of `IsSecure`:
```
	private boolean isSecure(Server server, String serviceId) {
		IClientConfig config = this.clientFactory.getClientConfig(serviceId);
		if (config != null) {
			Boolean isSecure = config.get(CommonClientConfigKey.IsSecure);
			if (isSecure != null) {
				return isSecure;
			} 
		}
		return serverIntrospector(serviceId).isSecure(server);
	}
```

[...]

PR #1127, opened 2016-06-23
Without explicit client config of IsSecure, lookup from registration

william-tran:
instead of defaulting to false.

Fixes gh-1126


spencergibb:
And when using eureka it will be based on if the secure port is enabled, not the port ending in 443.

spencergibb:
Closed with a little formatting polish on test via 99beb9d04573b96a53006a0bbef8b63747b96169

Issue #1123, opened 2016-06-22
Zuul error filter


Aloren:
In Zuul if exception happened during filtering, routing, etc -- this exception can be logged by error filter. (For more details see `com.netflix.zuul.http.ZuulServlet`). Currently if exception happens, for example, in preroute part it is not logged anywhere. I propose to add simple error filter that is able to log such exceptions. Thanks.

```
import com.netflix.zuul.ZuulFilter;
import com.netflix.zuul.context.RequestContext;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class ZuulErrorLoggingFilter extends ZuulFilter {

    private final Logger logger = LoggerFactory.getLogger(getClass());

    @Override
    public String filterType() {
        return "error";
    }

    @Override
    public int filterOrder() {
        return 0;
    }

    @Override
    public boolean shouldFilter() {
        return RequestContext.getCurrentContext().getThrowable() != null;
    }

    @Override
    public Object run() {
        Throwable throwable = RequestContext.getCurrentContext().getThrowable();
        logger.error("Exception was thrown in filters: ", throwable);
        return null;
    }
}
```

spencergibb:
Pull Requests welcome!

Aloren:
@spencergibb Please review the following [PR](https://github.com/spring-cloud/spring-cloud-netflix/pull/1129)
Thanks!

spencergibb:
@Aloren looks like I ended up with something similar to your PR, sorry it took me so long to understand.

Aloren:
@spencergibb Looks even better than my initial pull request. Thanks!

umamaheswarreddy:
thanks :)

AlekseyVasilenko89:
Анастасия здравствуйте. Нужна помощь по вашему решению, оно у меня не работает(. Что ещё нужно добавить. Может этот фильтр надо зарегистрировать или добавить конфигурацию.

PR #1129, opened 2016-06-24
Zuul exception handling

Aloren:


Aloren:
Looks like PR #1090 Load ribbon classes if defined via properties. broke the build in master.

spencergibb:
Yeah, currently those tests only fail in travis. It worked on our own CI server: https://jenkins.spring.io/view/Spring%20Cloud/view/CI/job/spring-cloud-netflix-ci/409/

spencergibb:
build fixed.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/error/ErrorLoggingFilter.java
```diff
@@ -0,0 +1,33 @@
+package org.springframework.cloud.netflix.zuul.filters.error;
+
+import com.netflix.zuul.ZuulFilter;
+import com.netflix.zuul.context.RequestContext;
+import org.slf4j.Logger;
```
spencergibb: This should be commons logging.

spencergibb:
@dsyer thoughts?

dsyer:
I'd expect errors to be bubbled up to error controller. Maybe we need a bit more analysis here of when this filter should run and when not?

spencergibb:
@Aloren what kinds of exceptions are you hoping to log?

Aloren:
@spencergibb I was doing some testing with custom filters and had NPE inside one of the filters and it was not neither logged nor rethrown anywhere up. I suppose logging is not the best solution for such case but at least for debugging purposes it is quite a useful feature.

Aloren:
@spencergibb I just reviewed error handling in Zuul and trying to understand the flow.
Let's assume that we have a pre filter which fails with exception. Exception will be handled in error and postroute filters.
```
ZuulServlet.java

        try {
                preRoute();
            } catch (ZuulException e) {
                error(e);
                postRoute();
                return;
            }
```
Currently there are no error filters in spring-cloud-zuul and 2 postroute filters: SendErrorFilter and SendResponseFilter.
`SendErrorFilter` checks for error.status_code in `RequestContext`, `SendResponseFilter` checks for response body or response body stream existence in `shouldFilter` method. For our assumption none of the filters will be run, so client will not receive error response. And I suppose any unhandled exception occured during processing of request (filtering in zuul as well) should be returned to client with correct error response. 
My thoughts about SendErrorFilter:
1. `SendErrorFilter` seems to be error filter as it handles only exceptional cases.
2. `SendErrorFilter` should check for existence of exception in `RequestContext` in `shouldFilter` method. As I understand current implementation works only for `RibbonRoutingFilter`.
What do you think? Thanks

spencergibb:
`SendErrorFilter` is triggered by `RibbonRoutingFilter` and `SimpleHostRoutingFilter`.
We don't have any `error` type filters and the condition of `RequestContext.getCurrentContext().getThrowable() != null` seems correct.

The routing filters catch `Exception` so it seems like this would only be run if there are exceptions like a NPE in a condition or any exception in a `pre` filter.

Aloren:
@spencergibb 
1. It looks like that in case of any exception in prefilter client will get response with status 200 and empty body, routing will not be executed, and exception will not be handled. 

2. Current exception handling and error response creation is quite confusing. 
> SendErrorFilter is triggered by RibbonRoutingFilter and SimpleHostRoutingFilter.
Shouldn't `SendErrorFilter` be triggered by any exception in filtering chain?

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/zuul/SimpleZuulServerApplicationTests.java
```diff
@@ -83,13 +90,26 @@ public void getOnSelf() {
 	}
 
 	@Test
-	public void getOnSelfViaFilter() {
+	public void getOnSelfViaFilterShouldSucceed() {
+		getOnSelfViaFilter();
+	}
+
+	@Test
+	public void exceptionOccurredInFilterShouldBeLogged() throws Exception {
```
Aloren: @spencergibb As you can see from this test: exception was thrown in prefilter, but response 200 with empty body was returned to the client. Is it correct behaviour? I suppose no.

Aloren:
Moreover if you look into zuul-netflix-webapp they alredy have such error filter: [`ErrorResponse.groovy`](https://github.com/Netflix/zuul/blob/1.x/zuul-netflix-webapp/src/main/groovy/filters/pre/ErrorResponse.groovy)
I think that I will remove my `ErrorLoggingFilter` and will change `SendErrorFilter` logic to handle any exception.

aivans:
I am encountering the same issue with "pre" filters.

@Aloren Conceptually it makes a lot of sense to have `SendErrorFilter` as an `error` filter type.

One thing that should be documented somewhere is that the `post` filters are also invoked all the time.
```
            try {
                preRoute();
            } catch (ZuulException e) {
                error(e);
                postRoute();
                return;
            }
```
Given this, I have some thoughts.
Maybe it would be a good thing to prepare the response contents in the `error` filter and handle the actual response handling in `SendResponseFilter` only, either success, either error?
I have the feeling this would work better on the long run.

WDYT?

Aloren:
> Maybe it would be a good thing to prepare the response contents in the error filter and handle the actual response handling in SendResponseFilter only, either success, either error?

@aivans I like your idea. This is how it is done in netflix-zuul-webapp in [ErrorResponse.groovy](https://github.com/Netflix/zuul/blob/1.x/zuul-netflix-webapp/src/main/groovy/filters/pre/ErrorResponse.groovy). I would also add that `SendErrorFilter` should find appropriate `ExceptionHandler`, execute it, fill response body with error response and after that `SendResponseFilter ` should be executed. As for me this flow looks good.
But I've faced an interesting issue with exception handling during implementing my PR. Let's take for example `SimpleHostRoutingFilter`. Its `run` method implementation throws `IOException` which then wrapped via following code into `UndeclaredThrowableException`:
```
} catch (Exception ex) {
			ReflectionUtils.rethrowRuntimeException(ex);
		}
```
Next `run` method is called inside `runFilter` method of `ZuulFilter`, which in its turn is called by `FilterProcessor`'s `processZuulFilter`. The whole method is wrapped with:
```
} catch (Throwable e) {
// ommited code
            if (e instanceof ZuulException) {
                throw (ZuulException) e;
            } else {
                ZuulException ex = new ZuulException(e, "Filter threw Exception", 500, filter.filterType() + ":" + filterName);
                ctx.addFilterExecutionSummary(filterName, ExecutionStatus.FAILED.name(), execTime);
                throw ex;
            }
        }
```
SO when IOException actually happens I will get a huge stacktrace:
- ZuulException
- UndeclaredThrowableException
- IOException
It would be better to omit UndeclaredThrowableException from stacktrace, but currently this is not possible because ZuulException is checked exception and `ZuulFilter`'s method `run` is declared without any checked exceptions :(
Due to this stacktrace resolving of rootcause looks really awful. I would like to hear your opinions on this issue as well. Thanks.

aivans:
@Aloren 
` would also add that SendErrorFilter should find appropriate ExceptionHandler, execute it, fill response body with error response and after that SendResponseFilter should be executed.`
I think that as @dsyer mentioned, the error response handling and sending should be delegated to Spring mechanism as it is done now from various places where a forward to `error.path` is triggered. The simply put it, which I think your PR already takes care of, you just need to make sure an error filter is invoked, the filter finds the exception in the context and makes sure the error.path forward happens.

`
} catch (Exception ex) {
            ReflectionUtils.rethrowRuntimeException(ex);
        }
`
I cannot find this code in `SimpleHostRoutingFilter`.
The practice seems that RuntimeException is thrown, which seems a good practice.
```
if (e instanceof ZuulException) {
                throw (ZuulException) e;
            } else {
                ZuulException ex = new ZuulException(e, "Filter threw Exception", 500, filter.filterType() + ":" + filterName);
```
Zuul will wrap this directly in a ZuulException.

aivans:
By the way, the NPEs in `pre` filters can give a lot of headaches for those writing custom filters.

aivans:
@Aloren Regarding the error response preparation. Seems that the mechanism is something like this:

Steps:

-  `pre` filter prepares the error response details
Example:
```
    private void prepareRateLimitErrorResponse(final RequestContext ctx) {
        LOGGER.debug("Prepare rate limit error response");       

        ctx.set("error.status_code", HttpStatus.TOO_MANY_REQUESTS.value());
        ctx.set("error.exception", 
                new ZuulException("Quota exceeded", HttpStatus.TOO_MANY_REQUESTS.value(), ""));
        ctx.set("error.message", "Quota exceeded");

    }
```
-  `pre` filter throw a RuntimeException
```
throw new RuntimeException();
```
- `route` filters are skipped now
- `SendErrorFilter` prepares the request attributes and forwards to error path

It works for me

spencergibb:
Closing as this PR is too broad. If you want to rework for just for #1123 go ahead and reopen or open a new PR.

Issue #1143, opened 2016-06-28
Feign - Apache HttpClient not being used for non loadbalanced clients


singram:
I'm attempting to get Feign to use the Apache HttpClient library.
This is supported by the Netflix Feign project, support for which appears to have been merged into spring-cloud-netflix via the following issue
https://github.com/spring-cloud/spring-cloud-netflix/issues/415
and code
https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.1.0.RELEASE/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java

My expectation was by including 
    `compile('com.netflix.feign:feign-httpclient')`
in my gradle.build this would be autowired in and in fact looks to be the case by examining the output of the \autoconfig actuator endpoint.

My issue is that I do not believe it's actually using the Apache HttpClient.
I see zero change in behavior at the log level (Feign logging set to FULL) and the expected Apache HttpClient logging is also absent (set by `logging.level.org.apache.http: DEBUG` in application.properties).  In the same application I have a test call using Apache HttpClient directly and see the expected log behavior.

I have found no documentation on this outside of the issues/code but would be happy to contribute.  I've also tried the `feign-okhttp` library with identical results.  In addition I am using Pinpoint APM which supports the instrumentation of ok-http and http-client, it does not recognize any Feign call at the moment but does the direct usage of http-client adding evidence to the theory that Feign is not using either specified library and I'm doing something incorrect.

Any thoughts would be greatly appreciated.

        springBootVersion = '1.3.5.RELEASE'
        springCloudVersion = 'Brixton.SR1'

Thanks

spencergibb:
Are you using a direct url? ie `@FeignClient(url="http://example.com", name="example")`?

spencergibb:
If so, the following is a workaround.
```
@EnableFeignClients
@RestController
@SpringBootApplication
@RibbonClient(name = "example", configuration = RibbonConfig.class)
public class DemofeignapachehttpgradleApplication {

	@FeignClient("example")
	public interface Client {

		@RequestMapping(path = "/get", method = RequestMethod.GET)
		String get();
	}

	@Autowired
	Client client;

	@RequestMapping(path = "/", produces = "application/json")
	public String get() {
		return client.get();
	}

	public static void main(String[] args) {
		SpringApplication.run(DemofeignapachehttpgradleApplication.class, args);
	}

}

[...]

PR #1144, opened 2016-06-29
Allow custom feign clients when not using ribbon.

spencergibb:
/cc @dsyer @marcingrzejszczak @adriancole 

Wondering if the scope of this is ok for 1.1.x or should it go in 1.2.0?

fixes gh-1143

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/DefaultTargeter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/DefaultTargeter.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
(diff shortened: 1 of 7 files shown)


adriancole:
only suggestion might be to make the types extracted package private, since as far as I can tell, they don't yet need to be public

spencergibb:
And I made some public because the test is in another package :-( I should make a new one.

adriancole:
or move the tests to the same package?

dsyer:
It seems like a small thing, but since you decided it wasn't a bug, I would say lets put it in 1.2.0. We want to release that quite soon anyway, and there is a workaround in the meantime. In 1.2.0 we want the apache client to be the default anyway, don't we?

wujiming:
赶上直播了吗，正好我也遇到了这个问题

marcingrzejszczak:
I agree with @dsyer - let's put it into 1.2.0.

PR #1147, opened 2016-06-29
Improved proxy header support

### LINKED ISSUE #1108: Redirects from backend service to internal IP
The headers `X-Forwarded-*` headers set in `PreDecorationFilter` are ignored by default spring-boot services built on Tomcat. As a consequence, redirects to absolute URLs (due to `context.setUseRelativeRedirects(false)` in `TomcatEmbeddedServletContainerFactory`) will go to the value in `Host` rather than `X-Forwarded-Host`. 

For instance, this happens with `server.contextPath: /foo` and a request to `curl -H 'X-Forwarded-Host: api.example.com' http://127.0.0.1/foo` with a redirect to `http://127.0.0.1/foo/` (due to `context.setMapperContextRootRedirectEnabled(true)` in `TomcatEmbeddedServletContainerFactory`)

I think there are 2 workaround.

1) proxy: use a  `ZuulFilter` to set `Host` additionally to `X-Forwarded-Host`

2) client: add a `Valve` to override the server name:

```java
	@Bean
	public EmbeddedServletContainerCustomizer customizer() {
		return (final ConfigurableEmbeddedServletContainer container) -> {

			final TomcatEmbeddedServletContainerFactory tomcat = (TomcatEmbeddedServletContainerFactory) container;

			tomcat.addContextCustomizers((final Context context) -> {
				context.setMapperContextRootRedirectEnabled(false);
			});

			tomcat.addContextValves(new ValveBase() {

				@Override
				public void invoke(final Request request, final Response response) throws IOException, ServletException {

					final MessageBytes serverNameMB = request.getCoyoteRequest().serverName();
					String originalServerName = null;
					final String forwardedHost = request.getHeader("X-Forwarded-Host");
					if (forwardedHost != null) {
						originalServerName = serverNameMB.getString();
						serverNameMB.setString(forwardedHost);
					}

					int originalPort = -1;
					final String forwardedPort = request.getHeader("X-Forwarded-Port");
					if (forwardedPort != null) {
						try {
							originalPort = request.getServerPort();
							request.setServerPort(Integer.valueOf(forwardedPort));
						} catch (final NumberFormatException e) {
							log.debug("ignoring forwarded port {}", forwardedPort);
						}
					}

					try {
						getNext().invoke(request, response);
					} finally {
						if (forwardedHost != null) {
							serverNameMB.setString(originalServerName);
						}
						if (forwardedPort != null) {
							request.setServerPort(originalPort);
						}
					}
				}
			});
		};
	}
```

Note: requires disabling `mapperContextRootRedirectEnabled` for `Valve` processing to kick in in order to work with the example above

sfussenegger:
- add `zuul.add-host-header` property to add `Host` header
- add port to `X-Forwarded-Host` as defined in [RFC 7239](https://tools.ietf.org/html/rfc7239)
- extract `PreDecorationFilter.filterOrder()` into a public constant

fixes #1108

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
@@ -19,6 +19,8 @@ package org.springframework.cloud.netflix.zuul.filters.pre;
 import java.net.MalformedURLException;
 import java.net.URL;
 
+import javax.servlet.http.HttpServletRequest;
+
 import org.springframework.cloud.netflix.zuul.filters.ProxyRequestHelper;
 import org.springframework.cloud.netflix.zuul.filters.Route;
 import org.springframework.cloud.netflix.zuul.filters.RouteLocator;
@@ -35,6 +37,8 @@ import lombok.extern.apachecommons.CommonsLog;
 
 @CommonsLog
 public class PreDecorationFilter extends ZuulFilter {
+
+	public static final int FILTER_ORDER = 5;
 
 	private RouteLocator routeLocator;
 
@@ -58,7 +62,7 @@ public class PreDecorationFilter extends ZuulFilter {
 
 	@Override
 	public int filterOrder() {
-		return 5;
+		return FILTER_ORDER;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
```diff
@@ -149,6 +152,9 @@ else if (!xforwardedfor.contains(remoteAddr)) { // Prevent duplicates
 					}
 					ctx.addZuulRequestHeader("X-Forwarded-For", xforwardedfor);
 				}
+				if (this.properties.isAddHostHeader()) {
```
dsyer: So maybe instead of a boolean this could be the header name (defaults to "X-Forwarded-Host")? If empty do nothing.

sfussenegger: That's an of course an option. 

For the current default tomcat configuration of spring-boot a combination of `X-Forwarded-Proto` and `Host` is required. A buggy implementation could ignore the `*-Proto` header if the `*-Host` header is missing so it might be better to have two identical headers - `Host` and `X-Forwarded-Host` - for best results.

But since HTTP proxy headers are not standardized one could probably make a point to make all headers configurable. But in this case, a customizer interface would probably the better call.

spencergibb:
Closed with tests added via beb84dbe07e1be3bd6cb5db9df4241af55c702fb

Issue #1155, opened 2016-07-01
Servo: DefaultMetricsTagProvider makes use of Guava


brenuart:
`DefaultMetricsTagProvider` makes use of Guava's ImmutableMap class.
Guava is "forbidden" in the codebase as per https://github.com/spring-cloud/spring-cloud-netflix/issues/154

PR #1156, opened 2016-07-01
GH1155: remove Guava ImmutableMap in favour of Collections.unmodifiab…

brenuart:
https://github.com/spring-cloud/spring-cloud-netflix/issues/1155:
- Remove usage of Guava's ImmutableMap in favor of Colelctions.unmodifiableMap()
- Make sanitizeUrlTemplate() protected for easy sub-classing


brenuart:
`clientHttprequestTag()` returns an ImmutableMap, but `httpRequestTags()` doesn't...
Ideally, they should both have the same behavior. What do you think ?

Issue #1223, opened 2016-08-03
Content-length not properly set when using the FormBodyWrapperFilter


jebeaudet:
When using the `FormBodyWrapperFilter`, the form parameters gets encoded in the `FormHttpMessageConverter`. The problem arises when you get a form param that was encoded with the javascript `encodeURIComponent` method or `curl` which do not have the same rules as `URLEncoder` (for example, '(' doesn't get encoded in `encodeURIComponent` but will be in `URLEncoder`). 

This causes the body to be larger than the original body. While this is not necessarily a problem, the content-length of the request is not properly set in `HttpClientRibbonCommand` on Brixton and in `AbstractRibbonCommand` on the master branch. Result : the form param get stripped on the backend or in the case where the body is shorter than the original, the backend server will wait for the missing bytes and will most likely trigger a timeout.

I'll submit a PR shortly for discussion.

jebeaudet:
Thx for merging the PR. Would it be possible to merge the PR for the brixton branch as well for the 1.1.6 release?

ryanjbaxter:
#1224?

jebeaudet:
Yes exactly

ryanjbaxter:
Done

jebeaudet:
Thanks a bunch!

PR #1225, opened 2016-08-03
Content-length not properly set when using the FormBodyWrapperFilter

jebeaudet:
Fix https://github.com/spring-cloud/spring-cloud-netflix/issues/1223 in master branch.

The PR https://github.com/spring-cloud/spring-cloud-netflix/issues/1223 fixes the Brixton branch.

There was also a bug related to https://github.com/spring-cloud/spring-cloud-netflix/commit/aa23660d0c906d795b2f7088196fa049cf2dbb24 where the `FormBodyWrapperFilter` didn't set both `request` object in the corresponding wrapper.

I'm open to suggestion here as I wanted to add tests and I struggled to get a request sent with `TestRestTemplate` *without* using the `FormHttpMessageConverter` otherwise the form params would not get encoded like in javascript or curl. Also, for some reason `RestClient` doesn't have the same behavior so I had to override the test in the corresponding integration test class.


ryanjbaxter:
@jebeaudet thanks for the pull request, can you take a look at the merge conflicts?

jebeaudet:
Done!

PR #1260, opened 2016-08-15
Update ribbon client to generate a proper URI endpoint even if there …

mikexliu:
…is no information about the request

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
@@ -200,7 +200,11 @@ public class RibbonClientConfiguration {
 		public URI reconstructURIWithServer(Server server, URI original) {
 			String scheme = original.getScheme();
 			if (!"https".equals(scheme) && this.serverIntrospector.isSecure(server)) {
-				original = UriComponentsBuilder.fromUri(original).scheme("https").build(true)
+				original = UriComponentsBuilder.fromUri(original)
+						.scheme("https")
+						.host(server.getHost())
+						.port(server.getPort())
+						.build(true)
 						.toUri();
 			}
 			return super.reconstructURIWithServer(server, original);


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
```diff
@@ -200,7 +200,11 @@ protected OverrideRestClient(IClientConfig ncc,
...
 			if (!"https".equals(scheme) && this.serverIntrospector.isSecure(server)) {
-				original = UriComponentsBuilder.fromUri(original).scheme("https").build(true)
+				original = UriComponentsBuilder.fromUri(original)
+						.scheme("https")
+						.host(server.getHost())
+						.port(server.getPort())
+						.build(true)
 						.toUri();
 			}
 			return super.reconstructURIWithServer(server, original);
```
spencergibb: So `super.reconstructURIWithServer` adds the host and port from the server. Was there an error in the original problem?

mikexliu: Yeah, the problem was toUri() call tried to create an URI without the host and port and threw an invalid URI exception so it could never reach the super call.

spencergibb: I ask because if host and port of the `server` match original, none of the logic in super is executed, which is what this change does.

spencergibb: Where were you getting an empty url (ie `https:///`)?

ryanjbaxter:
Wouldnt it be better to let the developer know the URI is invalid instead of trying to construct a URI without a host and port?

spencergibb:
@ryanjbaxter right, I'd rather fix the problem of why there is no host and port.

mikexliu:
I received this error when accessing a valid URL via zuul.

I can try creating a unit test again but had trouble initially with this particular case. Essentially, it works fine if it's http but does not seem to work with https. The passed in URI is empty (except for the scheme field).

What is the expected value for `original` if I am accessing `https://zuulhost:2345/application-service` ?

ryanjbaxter:
@mikexliu a unit test that demonstrates the problem or a simple sample app would help us out

mikexliu:
@ryanjbaxter Yeah. I'll create a simple app if I can't get the unit test working. No worries.

dsyer:
FWIW it works for me, and I can't see any failures, irrespective of the path and server. URI allows you to construct one with no host, so I don't think that was the problem.

spencergibb:

Issue #1259, opened 2016-08-15
Actuator/s-c-bus endpoint to change eureka instance status


spencergibb:
```
       UP, // Ready to receive traffic
        DOWN, // Do not send traffic- healthcheck callback failed
        STARTING, // Just about starting- initializations to be done - do not
        // send traffic
        OUT_OF_SERVICE, // Intentionally shutdown for traffic
        UNKNOWN;
```

This could be useful to start instances with `OUT_OF_SERVICE` then changing it to `UP` after testing.

heowc:
I think this feature would be useful.
Is there anything in progress regarding this?

(It was done a few years ago and then canceled, but I don't think it's been done since.)

Or is it something that already exists but I just can't find it? 🤣

PR #1261, opened 2016-08-15
Initial implementation for /eureka actuator endpoint.

spencergibb:
Allows you to override the status and delete the override.

fixes gh-1259

/cc @dsyer @ryanjbaxter 

The initial thought was to use `ApplicationInfoManager.setInstanceStatus()` but in practice that doesn't work. Making direct api calls to eureka works without a hitch.

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/CloudEurekaClient.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/CloudEurekaClient.java
@@ -16,8 +16,12 @@
 
 package org.springframework.cloud.netflix.eureka;
 
+import java.lang.reflect.Field;
 import java.util.concurrent.atomic.AtomicLong;
 
+import com.netflix.appinfo.InstanceInfo;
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@spencergibb Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1261)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1261) to manually synchronize the status of this Pull Request.

dsyer:
Could this be generic (i.e. in commons) with a eureka implementation?

spencergibb:
@dsyer I could. It could use spring-boot `Status` and let the impl translate.

spencergibb:
@dsyer started thinking about where to put this in commons. Seems like it would go with service registry, not discovery client, but that's not going to be merged until Dalston. Should we wait on this as well?

dsyer:
I think it might be best to wait, but I'm open to persuasion. It's not like users can't write code to do this if they need it with eureka. If we need somethings now at least let's give it a generic name.

Issue #1270, opened 2016-08-17
Non-secure FeignClient assigned https URI on non-secure port


rhutzel:
When a server registers with Eureka indicating both `nonSecurePortEnabled` and `securePortEnabled` to be true, FeignClients that want to use the non-secure port end up with an unusable https URL pointing to the non-secure port.

The FeignClient builds server instances based upon the insecure port because it has `ribbon.IsSecure` set to false. See `DiscoveryEnabledNIWSServerList::initWithNiwsConfig`.

However, the original URI, `http:///server/endpoint`, received by `FeignLoadBalancer::reconstructURIWithServer` is converted to `https:///server/endpoint` only because the service has its secure port enabled.

This results in an unusable URL that points to `https://hostname:insecurePort/endpoint`

Expected behavior would be for `reconstructURIWithServer` to also take into account whether the non-secure port is also enabled and if the FeignClient has `Ribbon.IsSecure` set to false.

PR #1275, opened 2016-08-18
#1270: Feign load balancer respects Ribbon.IsSecure as false for securable service

rhutzel:
Related to issue #1270  Feign load balancer respects Ribbon.IsSecure being set to false even when a discovered service has a secure port enabled in addition to a non-secure port enabled.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
@@ -26,9 +26,14 @@ import com.netflix.loadbalancer.Server;
  */
 public class DefaultServerIntrospector implements ServerIntrospector {
 	@Override
-	public boolean isSecure(Server server) {
+	public boolean secureAvailable(Server server) {
 		// Can we do better?
 		return (""+server.getPort()).endsWith("443");
+	}
+
+	@Override
+	public boolean insecureAvailable(Server server) {
+		return (""+server.getPort()).endsWith("80");
 	}
 
 	@Override
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
@@ -26,13 +26,13 @@ import com.netflix.loadbalancer.Server;
  */
 public class DefaultServerIntrospector implements ServerIntrospector {
 	@Override
-	public boolean secureAvailable(Server server) {
+	public boolean isSecure(Server server) {
 		// Can we do better?
 		return (""+server.getPort()).endsWith("443");
 	}
 
 	@Override
-	public boolean insecureAvailable(Server server) {
+	public boolean isInsecure(Server server) {
 		return (""+server.getPort()).endsWith("80");
 	}
 
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
@@ -107,8 +107,8 @@ public class FeignLoadBalancer extends
 	}
 
 	private boolean onlySecureAvailable(Server server) {
-		return this.serverIntrospector.secureAvailable(server)
-				&& !this.serverIntrospector.insecureAvailable(server);
+		return this.serverIntrospector.isSecure(server)
+				&& !this.serverIntrospector.isInsecure(server);
 	}
 
 	static class RibbonRequest extends ClientRequest implements Cloneable {
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
@@ -98,13 +98,17 @@ public class FeignLoadBalancer extends
 
 	@Override
 	public URI reconstructURIWithServer(Server server, URI original) {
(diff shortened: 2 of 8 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
```diff
@@ -26,12 +26,17 @@
  */
 public class DefaultServerIntrospector implements ServerIntrospector {
 	@Override
-	public boolean isSecure(Server server) {
+	public boolean secureAvailable(Server server) {
```
spencergibb: Changing the signature of a public API is probably not a good idea.

rhutzel: @spencergibb I was hesitant about that, too. My thinking was that isSecure, when it represents securePortEnabled in the case of the EurekaServerIntrospector, is an actively misleading security-related method. Knowing that the interface is used in only two places and that I'd need to modify it anyway with the additional insecure method, I erred on the side of code clarity. That said, should we stick with this or refactor for consistency to isSecure and isInsecure?

spencergibb:
So I copied `FeignLoadBalancerTest` over to s-c-n-eureka-client and modified `testInsecureUriFromInsecureClientConfigToOptionallySecureServer` to use the `EurekaServerIntrospector` without your changes and it passed. If `DiscoveryEnabledNIWSServerList` sets `isSecure` correctly on `DiscoveryEnabledServer` this should work. I'd like to see either a test that fails or a project that reproduces the problem.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/DefaultServerIntrospector.java
spencergibb: Downstream projects may be using `isSecure`.

rhutzel:
@spencergibb Are you having that `EurekaServerIntrospector` indicate that its secure port is enabled? If so, the test should fail without this pull request.

This fails from s-c-n-eureka-client on the master branch:

    @Test
    @SneakyThrows
    public void testInsecureUriFromInsecureClientConfigToOptionallySecureServer() {
        ServerIntrospector eurekaIntrospector = Mockito.mock(EurekaServerIntrospector.class);
        when(eurekaIntrospector.isSecure(Mockito.any(Server.class))).thenReturn(true);

        when(this.config.get(IsSecure, false)).thenReturn(false);
        this.feignLoadBalancer = new FeignLoadBalancer(this.lb, this.config, eurekaIntrospector);
        Server server = new Server("foo", 7777);
        URI uri = this.feignLoadBalancer.reconstructURIWithServer(server, new URI("http://foo/"));
        assertThat(uri, is(new URI("http://foo:7777/")));
    }

The scenario is having a service with both a secure port enabled and an insecure port enabled. Then, having a FeignClient with ribbon.IsSecure=false, wanting to use the service's insecure port. So, ribbon.IsSecure will cause the server list to be populated with "service:80". But later on, only because the service offers a secure port, and regardless of IsSecure=false and nonSecurePortEnabled=true, the URL becomes "https://service:80"

spencergibb:
But `ribbon.IsSecure=false` is used in `DiscoveryEnabledNIWSServerList`. This is what I'm trying to reconcile. If that is not working we need to fix that.

rhutzel:
`DiscoveryEnabledNIWSServerList` works as I'd expect. `ribbon.IsSecure=false` will cause it to add "service:insecurePort" to the server list and `ribbon.IsSecure=true` will cause it to add "service:securePort" to the server list.

Personally, I wish both ports were carried along with the `Server` so that I could dynamically set `@FeignClient("${dynamicProtocol}:/myservice")` and only have `ribbon.IsSecure` override the URI protocol and port if I didn't explicitly specify a protocol. But that's a separate discussion and I wanted to take baby steps.

Anyway, the URI rewrite logic that comes later in the `FeignLoadBalancer` will overwrite the URI with https if `ribbon.IsSecure=true`. So, if `DiscoveryEnabledNIWSServerList` is only going to provide one port when both secure and non-secure are available, it is correctly providing a port that reflects `ribbon.IsSecure`.

spencergibb:
Ok, so if `ribbon.IsSecure=false` works in `DiscoveryEnabledNIWSServerList` why would it be different when it gets to `FeignLoadBalancer`?

rhutzel:
Because when support for `ServerIntrospector` was added to `FeignLoadBalancer` last September, the following code appeared, which overwrites the protocol with https *either* when `ribbon.IsSecure=true` *or* when the discovered server makes a secure port available:

        if (!"https".equals(scheme) && (this.serverIntrospector.isSecure(server)
                || this.clientConfig.get(CommonClientConfigKey.IsSecure, false))) {

spencergibb:
`ServerIntrospector.isSecure` is used in other places as well. This would make the use inconsistent. Marking for discussion.

spencergibb:
Can you provide a project that reproduces the problem?

rhutzel:
This demo project reproduces the problem:
https://github.com/rhutzel/bug-demo-spring-secured-feign

spencergibb:
Thanks for the demo project. That helped me understand the case where it was happening. I decided to do things a bit differently. If `ribbon.IsSecure` is set, that value will be used, otherwise, the `ServerIntrospector` will be called. All three places that have similar functionality all use the same code now.

Closing in favor of 43adc84dfdfc33f3be0268e09b193afc82f9bf9a.

spencergibb:
@rhutzel will you try with snapshots when they are available? This is in master and cherry picked to 1.1.x.

rhutzel:
@spencergibb Of course. I like the approach that you took. Thank you for being so responsive.

spencergibb:
And thanks to you @rhutzel for being patient with me as I was trying to understand. I did run your sample with the fix, though I broke the 1.1.x build.

rhutzel:
@spencergibb Verified against `spring-cloud-netflix:1.1.6.BUILD-SNAPSHOT`  It resolves the issue in my sample project. Also, with your new logic, the client still responds to `ribbon.IsSecure: true` as expected.

PR #1318, opened 2016-09-02
Remove dep on commons-lang in DimensionalServoMetricNaming (#1268)

### LINKED ISSUE #1268: Servo: commons-lang:commons-lang required but not listed in pom.xml dependencies
`org.springframework.cloud.netflix.metrics.servo.DimensionalServoMetricNaming` requires `commons-lang:commons-lang` but that dependency is not listed in pom.xml

This is causing issue when using the Servo integration features only.

jkschneider:
`commons-lang` was only on the classpath as a result of an optional dependency on `commons-configuration` and so this class is a potential runtime problem. Fixes #1268.


spencergibb:
@jkschneider you could use `org.springframework.util.StringUtils.collectionToCommaDelimitedString`

jkschneider:
@spencergibb Ah, who knew?! Updated the PR.

PR #1325, opened 2016-09-08
Fix lastFetch status calculation.

herder:
Was previously milliseconds/seconds when calculating whether to display the message about reporting failures.


spencergibb:
@herder or @ryanjbaxter can one of you explain why this changed?

herder:
`lastFetch` is calculated using milliseconds, but `clientConfig.getRegistryFetchIntervalSeconds` is specified in seconds. So `lastFetch` was always on the order of a 1000 times too large, which made the "Eureka discovery client is reporting failures to connect to a Eureka server" always show up.
I'd be happy to write a test for this, but I'm not sure how... :)

ryanjbaxter:
@spencergibb exactly what @herder said.  At least that was my interpretation of the code when I looked at it.

spencergibb:
Thanks guys for the explanation. This snuck in to Brixton as well.

Issue #1348, opened 2016-09-15
eureka displays incorrect URL for healthCheck and statusPage


mattcj:
When `management.port` is customized to a value other than `server.port` `eureka.instance.statusPageUrlPath` and `eureka.instance.healthCheckUrlPath` are ignored.

This occurs because when a custom `management.port` is detected then `eureka.instance.statusPageUrl` and `eureka.instanceHealthCheckUrl`  values get set, and those take precedence over any custom value in the equivalent `*Path` values.

PR #1349, opened 2016-09-15
Use statusPage/healthCheckUrlPath w/ custom management.port

mattcj:
Fixes gh-1348

Take into account the `statusPageUrlPath` and `healthCheckUrlPath` values when building `statusPageUrl` and `healthCheckUrl`.

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -62,6 +62,7 @@ import static org.springframework.cloud.commons.util.IdUtils.getDefaultInstanceI
  * @author Dave Syer
  * @author Spencer Gibb
  * @author Jon Schneider
+ * @author Matt Jenkins
  */
 @Configuration
 @EnableConfigurationProperties
@@ -80,6 +81,12 @@ public class EurekaClientAutoConfiguration {
 
 	@Value("${eureka.instance.hostname:${EUREKA_INSTANCE_HOSTNAME:}}")
 	String hostname;
+
+	@Value("${eureka.instance.statusPageUrlPath:}")
+	String statusPageUrlPath;
+
+	@Value("${eureka.instance.healthCheckUrlPath:}")
+	String healthCheckUrlPath;
 
 	@Autowired
 	ConfigurableEnvironment env;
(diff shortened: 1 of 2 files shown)


spencergibb:
@Value doesn't take relaxed binding into account. I'll take a look in a bit

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
```diff
@@ -81,6 +82,12 @@
 	@Value("${eureka.instance.hostname:${EUREKA_INSTANCE_HOSTNAME:}}")
 	String hostname;
 
+	@Value("${eureka.instance.statusPageUrlPath:}")
```
spencergibb: Because the other values don't have camel case (`eureka.instance.hostname`, `management.port`, etc...) there are only two relaxed values and we account for those, this won't.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
```diff
@@ -88,6 +89,30 @@ public void managementPort() {
 	}
 
 	@Test
+	public void statusPageUrlPathAndManagementPort() {
+		EnvironmentTestUtils.addEnvironment(this.context, "server.port=8989",
+				"management.port=9999",
+				"eureka.instance.statusPageUrlPath=/myStatusPage");
```
spencergibb: `eureka.instance.status-page-url-path` is valid and would fail here.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
```diff
@@ -111,6 +118,12 @@ public EurekaInstanceConfigBean eurekaInstanceConfigBean(InetUtils inetUtils) {
 			if (StringUtils.hasText(this.hostname)) {
 				instance.setHostname(this.hostname);
 			}
+			if (StringUtils.hasText(statusPageUrlPath)) {
```
spencergibb: If you use a 
```
resolver = new RelaxedPropertyResolver(env);
resolver.getProperty("eureka.instance.statusPageUrlPath");
// etc...
```

I'd be more comfortable.

spencergibb:
@mattcj thanks for the quick turnaround and tests!

mattcj:
Thank you.

PR #1369, opened 2016-10-02
Close inputStream on errors writing to the outputStream

joshiste:
Currently all errors during writing to the outputStream are ignored. So the inputStream is read to the end. 

This is a problem when the inputStream is infite or very long running (`hystrix.stream` for example) and results in the inputStream never beingClosed and eating up threads (on zuul and the proxied backend). With this commit the errors during writing are not ignored resulting in closing the inputStream.

Would it be possible to integrate this bugfix into Camden.SR1?

### DIFF
--- a/spring-cloud-netflix-core/.gitignore
+++ b/spring-cloud-netflix-core/.gitignore
@@ -1,0 +1,1 @@
+/.apt_generated/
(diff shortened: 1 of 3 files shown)


spencergibb:
Zuul isn't recommended for use with long running requests.  I'm hesitant to change something like this in a service release. I'd like to see if others have similar problems. I wouldn't recommend running the hystrix stream through zuul.

joshiste:
Hmm I just figured out that this fix is rather useless since the underlying `org.apache.http.impl.io.ChunkedInputStream` never calls `close()` on the underlying socket.

So I revoke this PR and have a look at the proxy-stream solution of the hystrix-dashboard....

joshiste:
FYI: The correct way would be to call `abort()` on the underlying HttpGet but this would lead to bigger changes in `SimpleHostRoutingFilter` and `SendResponseFilter`.

PR #1372, opened 2016-10-03
Close response on errors writing to the outputStream

joshiste:
This is a follow up to #1369. 
I couldn't reopen the old PR so I needed to create a new one....

Currently all errors during writing to the outputStream are ignored. So the
response is read to the end. This is a problem when the response is
infite or very long running (`hystrix.stream` for example) and results in the
socket never being closed and eating up threads (on zuul and the proxied
backend). With this commit the errors during writing are not ignored resulting
in closing the response.

Now the underlying socket gets closed. The changes to `SimpleHostRoutingFilter` and `SendResponseFilter` weren't that big as expected.
Just instead of closing the inputStream (which ended up in trying to reach the end of the inputStream) the http response get's closed, so the underlying socket gets closed appropriate.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.netflix.zuul.filters.post;
 
 import java.io.ByteArrayInputStream;
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -184,28 +185,33 @@ public class SendResponseFilter extends ZuulFilter {
 			}
 		}
 		finally {
+			/**
+			* Closing the wrapping InputStream itself has no effect on closing the underlying tcp connection since it's a wrapped stream. I guess for http
+			* keep-alive. When closing the wrapping stream it tries to reach the end of the current request, which is impossible for infinite http streams. So
+			* instead of closing the InputStream we close the HTTP response.
+			*
+			* @author Johannes Edmeier
+			*/
 			try {
-				if (is != null) {
-					is.close();
+				Object zuulResponse = RequestContext.getCurrentContext()
+						.get("zuulResponse");
+				if (zuulResponse instanceof Closeable) {
+					((Closeable) zuulResponse).close();
 				}
 				outStream.flush();
 				// The container will close the stream for us
 			}
 			catch (IOException ex) {
+			log.warn("Error while sending response to client: " + ex.getMessage());
 			}
 		}
 	}
 
 	private void writeResponse(InputStream zin, OutputStream out) throws Exception {
-		try {
-			byte[] bytes = buffers.get();
-			int bytesRead = -1;
-			while ((bytesRead = zin.read(bytes)) != -1) {
-				out.write(bytes, 0, bytesRead);
-			}
-		}
-		catch(IOException ioe) {
-			log.warn("Error while sending response to client: "+ioe.getMessage());
+		byte[] bytes = buffers.get();
+		int bytesRead = -1;
+		while ((bytesRead = zin.read(bytes)) != -1) {
+			out.write(bytes, 0, bytesRead);
 		}
 	}
 
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -42,10 +42,10 @@ import org.apache.http.HttpHost;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
 import org.apache.http.ProtocolException;
-import org.apache.http.client.HttpClient;
 import org.apache.http.client.RedirectStrategy;
 import org.apache.http.client.config.CookieSpecs;
 import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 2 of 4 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -170,13 +173,8 @@ private void writeResponse(InputStream zin, OutputStream out) throws Exception {
 		byte[] bytes = new byte[INITIAL_STREAM_BUFFER_SIZE.get()];
 		int bytesRead = -1;
 		while ((bytesRead = zin.read(bytes)) != -1) {
-			try {
-				out.write(bytes, 0, bytesRead);
-				out.flush();
-			}
-			catch (IOException ex) {
```
spencergibb: Again, I'm hesitant to change something like this on a point release. We don't recommend running `/hystrix.stream` through zuul.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -176,8 +176,9 @@ public Object run() {
 		this.helper.addIgnoredHeaders();
 
 		try {
-			HttpResponse response = forward(this.httpClient, verb, uri, request, headers,
-					params, requestEntity);
+			CloseableHttpResponse response = forward(this.httpClient, verb, uri, request,
+					headers, params, requestEntity);
+			RequestContext.getCurrentContext().set("zuulResponse", response);
```
spencergibb: Move to `setResponse`

joshiste:
I did the requested changes.

> Again, I'm hesitant to change something like this on a point release. We don't recommend running `/hystrix.stream` through zuul. 

 So will you target this PR for 1.3?

spencergibb:
Yes, we'll target for 1.3

spencergibb:
@joshiste there are merge conflicts in `SendResponseFilter` will you rebase?

joshiste:
I'll have a look at it tomorrow.

spencergibb:
Thanks!

joshiste:
@spencergibb PR updated.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -172,27 +173,25 @@ else if (context.getResponseGZipped() && isGzipRequested) {
 		}
 		finally {
 			try {
-				if (is != null) {
```
ryanjbaxter: Should we not be closing the `InputStream` as well?  Or does doing that cause the issue you were trying to solve?  I just want to make sure we aren't introducing a memory leak of any kind...

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -172,27 +173,25 @@ else if (context.getResponseGZipped() && isGzipRequested) {
...
+				Object zuulResponse = RequestContext.getCurrentContext()
+						.get("zuulResponse");
+				if (zuulResponse instanceof Closeable) {
+					((Closeable) zuulResponse).close();
 				}
 				outStream.flush();
 				// The container will close the stream for us
 			}
 			catch (IOException ex) {
+				log.warn("Error while sending response to client: "+ex.getMessage());
```
ryanjbaxter: small nit, missing a space after `+`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
joshiste: IIRC correctly closing the InputStream simply had no effect. The InputStream instance was some kind of wrapped stream, wich tried to reach the end of the stream on closing... which resulted in waiting forever on infinite chunked http response.
IMHO the InputStream shloudn't be necessary... If you insist on closing the stream it must be done after closing the response (but will result in an IOException in the mentioned edge case on trying to reach the end, since the underlying HttpResponse is closed).

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
joshiste: fixed.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
ryanjbaxter: Maybe it would be worth adding this explanation in some comments in the code?

joshiste: I'll go for it tommorrow.

joshiste: Updated

spencergibb:
Unfortunately, this broke a test.

spencergibb:
That failure might not be your fault.

joshiste:
Will this make it into the Dalston release?

spencergibb:
Yes, can you rebase?

joshiste:
> Yes, can you rebase?

done.

asarkar:
Did this make into the Dalston release?

spencergibb:
Yes 3e7b196950c5f1aec6899fc35089b7c7be7bc72d is in the 1.3.0.RELEASE tag.

Issue #1124, opened 2016-06-23
Hystrix/Ribbon got error when using OAuth2


tonny1983:
I use spring-cloud Brixton.SR1, and I've got some microservices as follows:
 discovery  -> Eureka Server
some -> service which registers to discovery
ui  -> api gateway using Zuul with OAuth2 enabled

In "ui", I've got the following code which uses hystrix/Ribbon

```
@Service
public class SomeHystrixService {   
    @Autowired
    private OAuth2RestOperations restTemplate;

    @HystrixCommand(fallbackMethod = "fallbackSome")
    public String getSome() {
        return restTemplate.getForObject("http://some/getsome", String.class);
    }

    public String fallbackSome(){ 
        return "some service error";
    }
}
```

If without OAuth2, everything is fine. But with OAuth2 enabled, when I tried to access the service, I got the error of "org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'scopedTarget.oauth2ClientContext': Scope 'session' is not active for the current thread; consider defining a scoped proxy for this bean if you intend to refer to it from a singleton; nested exception is java.lang.IllegalStateException: No thread-bound request found: Are you referring to request attributes outside of an actual web request, or processing a request outside of the originally receiving thread? If you are actually operating within a web request and still receive this message, your code is probably running outside of DispatcherServlet/DispatcherPortlet: In this case, use RequestContextListener or RequestContextFilter to expose the current request."

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'scopedTarget.oauth2ClientContext': Scope 'session' is not active for the current thread; consider defining a scoped proxy for this bean if you intend to refer to it from a singleton; nested exception is java.lang.IllegalStateException: No thread-bound request found: Are you referring to request attributes outside of an actual web request, or processing a request outside of the originally receiving thread? If you are actually operating within a web request and still receive this message, your code is probably running outside of DispatcherServlet/DispatcherPortlet: In this case, use RequestContextListener or RequestContextFilter to expose the current request.
at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:355) ~[spring-beans-4.2.6.RELEASE.jar:4.2.6.RELEASE]
at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197) ~[spring-beans-4.2.6.RELEASE.jar:4.2.6.RELEASE]
[...]

PR #1379, opened 2016-10-05
Propagate RequestAttributes across Hystrix Thread boundary

taxone:
This code fixes [this open issue](https://github.com/spring-cloud/spring-cloud-netflix/issues/1124).
The problem with OAuth2 is that Hystrix threads can't access the request/session scoped beans stored by Spring in the main thread, so it's necessary to copy the attributes managed by the Spring **RequestContextHolder** from the main thread to the threads managed by Hystrix thread pool.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfiguration.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
+import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategy;
 import org.apache.catalina.core.ApplicationContext;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -68,6 +69,11 @@ public class HystrixCircuitBreakerConfiguration {
 	@Bean
 	public HasFeatures hystrixFeature() {
 		return HasFeatures.namedFeatures(new NamedFeature("Hystrix", HystrixCommandAspect.class));
+	}
+
+	@Bean
+	public HystrixConcurrencyStrategy hystrixConcurrencyStrategy(){
+		return new HystrixRequestScopeConcurrencyStrategy();
 	}
 
 	@Configuration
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixRequestScopeConcurrencyStrategy.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public <T> Callable<T> wrapCallable(Callable<T> callable) {
+
+        //retrieves current ThreadLocal request attributes
+        RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
+
+        return new WrapperCallable<T>(requestAttributes,callable);
+    }
+
+    private static class WrapperCallable<T> implements Callable<T>{
```
ryanjbaxter: Add space before `{`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfiguration.java
```diff
@@ -70,6 +71,11 @@ public HasFeatures hystrixFeature() {
 		return HasFeatures.namedFeatures(new NamedFeature("Hystrix", HystrixCommandAspect.class));
 	}
 
+	@Bean
+	public HystrixConcurrencyStrategy hystrixConcurrencyStrategy(){
```
spencergibb: Should this be opt-in (`@ConditionalOnProperty`)? Since there can only be one `HystrixConcurrencyStrategy`?

Writtscher:
Exactly what I need. But @spencergibb is right. There is always one `HystrixConcurrencyStrategy`. This means as soon as you need to decorate the `Callable` you need to provide your own `HystrixConcurrencyStrategy` which is not doing much but decorates the `Callable` with another `Callable`.

IMO it would be nice if there would be a `DecoratingHystrixConcurrencyStrategy` that has a list of `CallableDecoration`-Components (just a name, maybe an interface where the call method provides the previous decorated `Callable` to continue chain). As soon as the `wrapCallable` method is called the strategy creates a `DecoratedCallable` that decorates the callable with all existing `CallableDecoration`s.

This way it would be easier to hook into it.

I hope this was understandable :expressionless:. What do you guys think?

spencergibb:
ping @taxone on requested changes

spencergibb:

PR #1384, opened 2016-10-06
#1376 Register and cancel event sending fix

### LINKED ISSUE #1376: EurekaInstanceRegisteredEvent not dispatching as intended
I went through the source code and figured out that eureka dispatches the following 3 events from springs `ApplicationContext`

1. `EurekaInstanceRenewedEvent`
1. `EurekaInstanceRegisteredEvent`
1. `EurekaInstanceCancelledEvent`  
  

Of these only the EurekaInstanceRenewedEvent works and the listener is triggered when the server receives a heartbeat. Other events are not dispatched when a new instance is registered or canceled from eureka.

I think `InstanceRegistry` is overriding the wrong `register` method.  

The `ApplicationResource` controller calls the `register` method in the  `InstanceRegistry` when a new instance is registered. But the call goes to the super class of `InstanceRegistry`, bypassing the event dispatch. Not sure if this is intended and this is how it is supposed to work. If it is then what is the use of the event and how can it be used?

bslota:
#1376 This commit fixes application event publishing when an instance is being registered/canceled.

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/InstanceRegistry.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/InstanceRegistry.java
@@ -89,7 +89,7 @@ public class InstanceRegistry extends PeerAwareInstanceRegistryImpl
 		final int instanceLeaseDuration = resolveInstanceLeaseDuration(info);
 		logRegistration(info, isReplication, instanceLeaseDuration);
 		publishEurekaInstanceRegisteredEvent(info, instanceLeaseDuration, isReplication);
-		superRegister(info, isReplication);
+		super.register(info, isReplication);
 	}
 
 	@Override
@@ -128,16 +128,7 @@ public class InstanceRegistry extends PeerAwareInstanceRegistryImpl
 	protected boolean internalCancel(String appName, String id, boolean isReplication) {
 		logCancelation(appName, id, isReplication);
 		publishEurekaInstanceCanceledEvent(appName, id, isReplication);
-		return superInternalCancel(appName, id, isReplication);
-	}
-
-	protected boolean superInternalCancel(String appName, String id,
-			boolean isReplication) {
 		return super.internalCancel(appName, id, isReplication);
-	}
-
-	protected void superRegister(InstanceInfo info, boolean isReplication) {
-		super.register(info, isReplication);
 	}
 
 	private void logRegistration(InstanceInfo info, boolean isReplication,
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/InstanceRegistry.java
```diff
@@ -126,4 +123,41 @@ public boolean renew(final String appName, final String serverId,
...
+	@Override
+	protected boolean internalCancel(String appName, String id, boolean isReplication) {
+		logCancelation(appName, id, isReplication);
+		publishEurekaInstanceCanceledEvent(appName, id, isReplication);
+		return super.internalCancel(appName, id, isReplication);
+	}
+
+	private void logRegistration(InstanceInfo info, boolean isReplication, int instanceLeaseDuration) {
+		if (log.isDebugEnabled()) {
+      log.debug("register " + info.getAppName() + ", vip " + info.getVIPAddress()
```
ryanjbaxter: spacing looks a little off here

bslota: Thanks for your feedback, I will fix formatting, and add tests

bslota:
Tests of overriden methods end with the calls of netflix's superclass methods (`internalCancel `and `register`). In order to mock them, I wrapped them in `super*` methods. In tests, I'm checking if event is properly produced, and if mentioned wrappers are being invoked.

ryanjbaxter:
@bslota Instead of creating methods you mock, have you considered using mockito and mocking the listener so you can verify which methods are called?  I am not really fond of adding code to a class for testing purposes.

bslota:
@ryanjbaxter Are you referring to the test class? Or are you talking about `superRegister` and `superInternalCancel` methods? I must say I don't like the idea of wrapping super class methods at all - if you have any idea of how to mock superclass methods, I will appreciate any advice..

ryanjbaxter:
I dont think u need to worry about what happens in the super calls from a unit test perspective.  You can assume that the super class is being tested properly.  You just need to worry about testing the code you added, which is essentially making sure the correct events get fired.

bslota:
Have no idea why spring-cloud-netflix-core module test failed...

bslota:
I tried to remove as many duplications as possible. I also used used created private methods in `renew`  method which initially wasn't changed, but I wanted the class to be consistent. Of course I added a test to check if `renew` also creates a proper event.

PR #1388, opened 2016-10-10
Upgrade hystrix 1.5.6

### LINKED ISSUE #1374: Upgrade Hystrix to 1.5.6
https://github.com/Netflix/Hystrix/releases/tag/v1.5.6

ryanjbaxter:
Fixes #1374
Not sure what in Hystrix 1.5.6 removes the Apache Commons Land dep but something does.  Regardless it seems like we should not be depending on a transitive dependency like that so I switched over to using Spring's `StringUtils` class.  I had to deal with some differences in functionality between the two.  With Spring, if the delimiter does not exist in the string it returns `null` where as with Apache it returns and array with one item being the string.  I replicated that functionality with some additional code.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
@@ -62,7 +62,7 @@ public class FormBodyWrapperFilter extends ZuulFilter {
 		this.requestField = ReflectionUtils.findField(HttpServletRequestWrapper.class,
 				"req", HttpServletRequest.class);
 		this.servletRequestField = ReflectionUtils.findField(ServletRequestWrapper.class,
-				"request", ServletRequest.class);		
+				"request", ServletRequest.class);
 		Assert.notNull(this.requestField,
 				"HttpServletRequestWrapper.req field not found");
 		Assert.notNull(this.servletRequestField,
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
@spencergibb usually no reason to have anyone look at these types of changes but since there is some code changes to go along with it can you take a look?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
```diff
@@ -16,22 +16,10 @@
...
-import java.util.Set;
-
-import javax.servlet.ServletInputStream;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletRequestWrapper;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.Part;
-
-import org.apache.commons.lang3.StringUtils;
+import com.netflix.zuul.ZuulFilter;
```
spencergibb: java, javax, org, then com :-)

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
```diff
@@ -232,8 +229,12 @@ private synchronized void buildContentData() {
 		private Set<String> findQueryParams() {
 			Set<String> result = new HashSet<>();
 			String query = this.request.getQueryString();
-			if (query != null) {
-				for (String value : StringUtils.split(query, "&")) {
+			String[] splitQuery = StringUtils.split(query, "&");
```
spencergibb: I recently learned about `tokenizeToStringArray`, may help clean this up a bit.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
spencergibb: yeah

ryanjbaxter:
Thanks @spencergibb!  Fixed ordering and tried `tokenizeToStringArray` and it worked!

PR #1409, opened 2016-10-19
Fix potential decoding and reencoding issues in query params when using SimpleHostRoutingFilter

jebeaudet:
Some webserver are picky with the encoding (uWSGI for example) and the current code behavior of decoding and reencoding the query string can lead to valid requests rejected by the backend after the modifications in Zuul. 

The test SimpleZuulProxyApplicationTests::getOnSelfWithComplexQueryParam will fail with the current codebase because it doesn't reencode the `/`, `:` and some other characters that are encoded in the javascript `encodeURIComponent` method.

This fixes https://github.com/spring-cloud/spring-cloud-netflix/issues/971. 

As you can see in the PR, I had to remove a UT that had a custom filter that injected overriden query string, this can't work anymore with what I have done. I've traced back the commit : https://github.com/spring-cloud/spring-cloud-netflix/commit/18dff8527d1555566b75ee3a6081ed89973b39ac but it doesn't mention anything about this "feature". I'm of course open to suggestions here, but unless we introduce a specific key to put in the `RequestContext` to override the query params, I don't see a graceful way to handle this.

Thanks for looking!


jebeaudet:
@spencergibb I don't see a clean way of doing this without adding an extra configuration option in `ZuulProperties` or `ZuulRoute`that would only affect `SimpleHostRoutingFilter` since requests handled by the `RibbonRoutingFilter` don't use the `ProxyRequestHelper` class. Is this something you'd be willing to accept?

I guess that another option would be making `ProxyRequestHelper` or `SimpleHostRoutingFilter` conditional on missing bean for easy customization? I'm currently using a `BeanPostProcessor` to inject my own `ProxyRequestHelper`, it's not cleanest way IMO.

Thanks

spencergibb:
`RibbonRoutingFilter` certainly uses [`ProxyRequestHelper`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonRoutingFilter.java#L45) though it doesn't use `getQueryString()`.

spencergibb:
You will need to rework this. An option is a possibility in `ZuulProperties`. The existing behaviour would need to be the default.

jebeaudet:
that's what I meant by not using `ProxyRequestHelper`, the apache client handles the encoding of the params as opposed to the `SimpleHostRoutingFilter` where we pass them already encoded. 

Adding a `ZuulProperties` would be something possible however it would only apply to `SimpleHostRoutingFilter` for the reason I just mentioned, I just don't find it clean having a property that doesn't apply everywhere, unless we name it specifically  like `useOriginalQueryStringEncodingInSimpleHostRoutingFilter`?

spencergibb:
Use a short name and put it in the extended javadocs.

jebeaudet:
Alright, I'll rework this one out then, thx for the feedback :)

jebeaudet:
@spencergibb I've updated the PR with an additional flag in `ZuulProperties`. Thanks!

jebeaudet:
I've messed up bad while trying to squash the merge commit, I'll close this one and open another.

PR #1410, opened 2016-10-19
Zuul Route Hystrix Fallback

### LINKED ISSUE #250: Zuul - specify fallback per route
Since it is possible to specify Hystrix and Ribbon parameters for each route in Zuul, it would be nice to specify the Hystrix fallback method per route also.

The solution  I have in mind right now:
- programmatic configuration of Hystrix by extending RibbonCommand, where ```getFallback()``` would be overridden
- application.yml: specify the custom class in the ```zuul.<<route>>.hystrix``` configuration
- use the default RibbonCommand if nothing configured

What do you think?

ryanjbaxter:
Fixes #250

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/apache/HttpClientRibbonCommandFactory.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/apache/HttpClientRibbonCommandFactory.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.netflix.zuul.filters.route.apache;
 
+import java.util.Collections;
 import java.util.Set;
 
 import org.springframework.cloud.netflix.ribbon.SpringClientFactory;
@@ -34,6 +35,10 @@ public class HttpClientRibbonCommandFactory extends AbstractRibbonCommandFactory
 	private final SpringClientFactory clientFactory;
 	
 	private final ZuulProperties zuulProperties;
+
+	public HttpClientRibbonCommandFactory(SpringClientFactory clientFactory, ZuulProperties zuulProperties) {
+		this(clientFactory, zuulProperties, Collections.<ZuulFallbackProvider>emptySet());
+	}
 
 	public HttpClientRibbonCommandFactory(SpringClientFactory clientFactory, ZuulProperties zuulProperties,
 										  Set<ZuulFallbackProvider> fallbackProviders) {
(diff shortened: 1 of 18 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommandFactory.java
```diff
@@ -0,0 +1,54 @@
...
+public abstract class AbstractRibbonCommandFactory implements RibbonCommandFactory {
+
+	private final Set<ZuulFallbackProvider> fallbackProviders;
+
+	private Map<String, ZuulFallbackProvider> fallbackProviderCache;
+
+	public AbstractRibbonCommandFactory(Set<ZuulFallbackProvider> fallbackProviders){
+		this.fallbackProviders = fallbackProviders;
+		this.fallbackProviderCache = new HashMap<String, ZuulFallbackProvider>();
+	}
```
spencergibb: Why not eagerly create the cache here?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/apache/HttpClientRibbonCommandFactory.java
```diff
@@ -16,33 +16,40 @@
 
 package org.springframework.cloud.netflix.zuul.filters.route.apache;
 
+import java.util.Set;
```
spencergibb: should have another newline.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommandFactory.java
ryanjbaxter: You mean populate the cache so its just a lookup when `getFallbackProvider` is called?

spencergibb: yes

ryanjbaxter:
@spencergibb made the changes you suggested.

When I first started making the changes it felt simple and it could go in 1.2.2.  Looking at what I ended up doing it seemed like a lot.  However, I don't think any of the changes effect backward compatibility so I dont see why we couldnt include in 1.2.2.

spencergibb:
What do our versions mean then?

Issue #1363, opened 2016-09-27
Provide ability to get the full interface type in a custom logger


rfalke:
FeignClientFactoryBean has a special treatment for the default Slf4jLogger:

```
	protected Feign.Builder feign(FeignContext context) {
		Logger logger = getOptional(context, Logger.class);

		if (logger == null) {
			logger = new Slf4jLogger(this.type);
		}
...
```

While other loggers are apparently constructed without the information about the full name of the feign interface (`this.type`).

Is is possible to get `this.type` also in a custom logger? If yes, how?

spencergibb:
Not currently.

rfalke:
What is a good idea to support this? I can think of:

1. adding a method Logger.setType(Class<?> type) which is called after the creation of the Logger instance. Problem: will always be a new bean created?
2. adding "Object ...args" to org.springframework.cloud.netflix.feign.FeignClientFactoryBean#getOptional and org.springframework.cloud.context.named.NamedContextFactory#getInstance. Problem: NamedContextFactory is in another git repo/maven project
3. adding a new parameter "Class<?> type" feign.Logger#logRequest, logRetry, logAndRebufferResponse and logIOException. The information is available in the only caller (SynchronousMethodHandler). Problem 1: one logger for all feign clients and therefore no log level filtering. Problem 2: feels unclean.

Any recommendations?

spencergibb:
Hmm... I'd think an interface like `FeignLoggerFactory` with a `create(Class)` method would be better.

spencergibb:
See https://twitter.com/starbuxman/status/785194631956828160

rfalke:
@venilnoronha: Thanks for the patch.

PR #1411, opened 2016-10-19
Adds Feign Logger factory interface

venilnoronha:
This PR adds the `FeignLoggerFactory` interface, which is used for creation of Feign `Logger` instances. As a fallback, `Slf4jLogger` is still supported.

Refer #1363 for more information.

I've signed the CLA. Please review and pull.

Thanks,
Venil


spencergibb:
Yes it should go into `FeignClientsConfiguration`.

venilnoronha:
Looks OK?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
```diff
@@ -128,6 +124,15 @@ public void setApplicationContext(ApplicationContext context) throws BeansExcept
 		return builder;
 	}
 
+	private Logger getLogger(FeignContext context) {
+		Logger logger = getOptional(context, Logger.class);
```
ryanjbaxter: Why are you doing this?  Why not just get the `FeignLoggerFactory` and create the logger?

venilnoronha:
Okay. I'll add some tests.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
venilnoronha: Well, we could do that, but, I'm thinking if that would really be a "default" implementation then. I was thinking of the default implementation to be more of a fallback mechanism than the core logic (of identifying the logger implementation). What do you think?

ryanjbaxter: I think if we are going to provide a factory to create the logger, let the factory create the logger.  How that logger is created can be an implementation detail.  (IE if a logger bean is present use it, if not than create a new one)

venilnoronha: Okay. That makes sense. I'll update the implementation. Thanks.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -116,9 +116,12 @@ public FeignLoggerFactory feignLoggerFactory() {
 
 	static class DefaultFeignLoggerFactory implements FeignLoggerFactory {
 
+		@Autowired(required = false)
```
ryanjbaxter: Use a constructor instead of auto wiring

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -108,4 +108,22 @@ public FormattingConversionService feignConversionService() {
...
 
+	@Bean
+	@ConditionalOnMissingBean
+	public FeignLoggerFactory feignLoggerFactory() {
+		return new DefaultFeignLoggerFactory();
+	}
+
+	static class DefaultFeignLoggerFactory implements FeignLoggerFactory {
+
+		@Autowired(required = false)
```
ryanjbaxter: Inject this via a constructor instead of auto-wiring

venilnoronha: You mean do a constructor-injection instead?

venilnoronha:
I've added a few tests.

venilnoronha:
I'm quite not understanding why those tests are failing. @ryanjbaxter do you have a hint on that?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignLoggerAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@Configuration
+public class FeignLoggerAutoConfiguration {
+
+	@Autowired(required = false)
+	private Logger logger;
+
+	@Bean
+	@ConditionalOnMissingBean(FeignLoggerFactory.class)
+	public FeignLoggerFactory feignLoggerFactory() {
```
spencergibb: This should go in `FeignClientsConfiguration`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignLoggerAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+		return new DefaultFeignLoggerFactory(logger);
+	}
+
+	static class DefaultFeignLoggerFactory implements FeignLoggerFactory {
```
spencergibb: Move to top level `org.springframework.cloud.netflix.feign`.

spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
```diff
@@ -2,6 +2,7 @@ org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
 org.springframework.cloud.netflix.archaius.ArchaiusAutoConfiguration,\
 org.springframework.cloud.netflix.feign.ribbon.FeignRibbonClientAutoConfiguration,\
 org.springframework.cloud.netflix.feign.FeignAutoConfiguration,\
+org.springframework.cloud.netflix.feign.FeignLoggerAutoConfiguration,\
```
spencergibb: Remove

spencergibb:
The errors are because of your auto-configuration, which in my recent review, you should remove.

spring-cloud-netflix-core/src/main/resources/META-INF/spring.factories
ryanjbaxter: This is causing the test failures

venilnoronha:
@spencergibb @ryanjbaxter thank you! I've made the necessary changes.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/DefaultFeignLoggerFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.springframework.cloud.netflix.feign;
+
+import feign.Logger;
+import feign.slf4j.Slf4jLogger;
+
+/**
+ * @author Venil Noronha
+ */
+public class DefaultFeignLoggerFactory  implements FeignLoggerFactory {
```
ryanjbaxter: Looks like there are 2 spaces before `implements`

venilnoronha:
Fixed that.

venilnoronha:
@spencergibb thank you!

PR #1412, opened 2016-10-20
Document Manual Feign Client Creation

### LINKED ISSUE #1211: Support different custom configurations for multiple feign clients with the same name
using `@FeignClient` in spring currently has one fatal issue, which is 
demonstrated with [this repository][]. If I got two feign clients, each having it's own configuration (excluded from component scan), and one of them has an `RequestInterceptor`, both (and all others, if present) get this interceptor also applied. The repository shows the following experiment to show this:

## experiment

We assume some foreign service, serving "foo" entities. You have to be authenticated
to access this the foo resource and have a role "USER". If you have "ADMIN", you
will see different data, as when "USER".

This behavior is implemented using plain spring boot with spring security in
the "producer" service. Additionally it registers to an eureka instance, so some
"consumer" service is able to fetch this.

So the consumer service should fetch data from consumer, automatically passing
basic auth of "users he know". While this may not clearly makes sense for real world,
something very similar is happening consuming OAuth2 secured resource servers with
client credentials grant!
So to get the data correctly, we define 2 clients, with individually set up
basic auth credentials using feign client configuration.

Since the main application
does **not** use `@SpringBootApplication`, but `@EnableAutoConfiguration`,
and an exclude filter on `@ComponentScan`, the two different configuration
do not declare the internal beans.

So the expected result of these 2 clients is: the `UserFooClient` retrieving
at least a foo entity with value "user1", while the `AdminFooClient`s result
should contain a "admin1" foo.

To verify, which result is intended, the same test is done using plain old
`RestTemplate`.

## consequences

This issue leads towards we can't really use multiple feign clients in spring boot
using the `@FeignClient` annotation, because as soon one client is using
some authorization flow (basic, oauth2), this is applied to all other 
clients.

[this repository]: https://github.com/xetys/feign-bug-demo

ryanjbaxter:
Fixes #1211

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -989,18 +989,17 @@ class FooController {
 
     @Autowired
 	public FooController(
-			ResponseEntityDecoder decoder, SpringEncoder encoder, EurekaClient discoveryClient) {
-        InstanceInfo prodSvcInfo =  discoveryClient.getNextServerFromEureka("PROD-SVC", false);
-		this.fooClient = Feign.builder()
+			ResponseEntityDecoder decoder, SpringEncoder encoder, Client client) {
+		this.fooClient = Feign.builder().client(client)
 				.encoder(encoder)
-                .decoder(decoder)
-	            .requestInterceptor(new BasicAuthRequestInterceptor("user", "user"))
-	            .target(FooClient.class, prodSvcInfo.getHomePageUrl());
-		this.adminClient = Feign.builder()
+				.decoder(decoder)
+				.requestInterceptor(new BasicAuthRequestInterceptor("user", "user"))
+				.target(FooClient.class, "http://PROD-SVC");
+		this.adminClient = Feign.builder().client(client)
 				.encoder(encoder)
-                .decoder(decoder)
-	            .requestInterceptor(new BasicAuthRequestInterceptor("admin", "admin"))
-	            .target(FooClient.class, prodSvcInfo.getHomePageUrl());
+				.decoder(decoder)
+				.requestInterceptor(new BasicAuthRequestInterceptor("admin", "admin"))
+				.target(FooClient.class, "http://PROD-SVC");
     }
 }
 ----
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
@spencergibb @alexandrenavarro please take a look, thanks!

alexandrenavarro:
Seems ok, just add a mention of possibility to use LoadBalancerFeignClient instead of EurekaClient. Thanks.

ryanjbaxter:
@alexandrenavarro not sure I understand how you could use `LoadBalancerFeignClient` as a replacement for `EurekaClient`.  How would you get the service URL for the client?

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -970,6 +970,46 @@ This replaces the `SpringMvcContract` with `feign.Contract.Default` and adds a `
...
+class FooController {
+
+	private FooClient fooClient;
+
+	private FooClient adminClient;
+
+    @Autowired
+	public FooController(
+			ResponseEntityDecoder decoder, SpringEncoder encoder, EurekaClient discoveryClient) {
+        InstanceInfo prodSvcInfo =  discoveryClient.getNextServerFromEureka("PROD-SVC", false);
```
spencergibb: Instead of this line inject `feign.Client` which if there is eureka will be a `LoadBalancerFeignClient`. Then below do `builder.client(client);`.

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -970,6 +970,45 @@ This replaces the `SpringMvcContract` with `feign.Contract.Default` and adds a `
...
+				.requestInterceptor(new BasicAuthRequestInterceptor("admin", "admin"))
+				.target(FooClient.class, "http://PROD-SVC");
+    }
+}
+----
+
+NOTE: In the above example `FeignClientsConfiguration.class` is the default configuration
+provided by Spring Cloud Netflix.
+
+NOTE: `prodSvcInfo` is the service the the Clients will be making requests to.
```
spencergibb: Should be just `PROD-SVC` now?

alexandrenavarro:
Perfect with the modification @spencergibb . Thanks for your work and reactivity.

neildcruz:
It would be helpful if this can be merged to the main spring-cloud documentation at http://projects.spring.io/spring-cloud/spring-cloud.html.

Took me a while to find this.

spencergibb:
@neildcruz http://cloud.spring.io/spring-cloud-static/Camden.SR6/#_creating_feign_clients_manually

PR #1448, opened 2016-11-07
Add composed route locators

joshiste:
Until now it's not easy to configure additional `RouteLocator`s to the existing one's configured by spring cloud.

This PR contains two commits to allow composing of `RouteLocator`s:
 
The **first commit** adds a `CompositeRouteLocator` implementation and adds it by default as primary `RouteLocator`.

The **second commit** removes the behaviour for simple routes from the `DiscoveryClientRouteLocator`.  The `SimpleRouteLocator` (for simple routes) and the `DiscoveryClientRouteLocator` (for discovered routes) are composed to preserve the current behaviour and to align with the "favor composition over inheritance" principle. 

If you think the second commit with quite a lot changes is to risky, I would appreciate if at least the first commit is accepted.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.netflix.zuul;
 
+import java.util.Collection;
 import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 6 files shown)


ryanjbaxter:
@joshiste appreciate the PR, but it is pretty daunting.  Do you mind separating it out into to PRs?

joshiste:
Ok then I will revert this PR to the first commit. Depending on your decision I'll file the second one.

ryanjbaxter:
Thanks!

spencergibb:
@joshiste sorry for the delay. I'm looking to clear the queue this week in time for the milestone or RC

PR #1450, opened 2016-11-07
Make Netflix Servlet 3.0.1 compatible

### LINKED ISSUE #1447: Spring Cloud Netflix Not Compatible for Tomcat 7.0.X (Servlet API 3.0.1)
Dear Team, 

I have found that the RibbonRoutingFilter.java, FormBodyWrapperFilter.java etc. are using the **getContentLengthLong ()** from **servlet 3.1 API**.  So we are not able to deploy our application to Tomcat 7.0.65 which has Servlet API 3.0.1

We are using the Spring cloud netflix **1.2.0.RELEASE** version in our application

It would be great if you make it compatible for Servler 3.0.1 API too.

ryanjbaxter:
Fixes #1447

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
@@ -53,6 +53,17 @@ public class SendResponseFilter extends ZuulFilter {
 	private static DynamicBooleanProperty SET_CONTENT_LENGTH = DynamicPropertyFactory
 			.getInstance()
 			.getBooleanProperty(ZuulConstants.ZUUL_SET_CONTENT_LENGTH, false);
+	private boolean useServlet31 = true;
+
+	public SendResponseFilter() {
+		super();
+		// To support Servlet API 3.0.1 we need to check if setcontentLengthLong exists
+		try {
+			HttpServletResponse.class.getMethod("setContentLengthLong");
+		} catch(NoSuchMethodException e) {
+			useServlet31 = false;
+		}
+	}
 
 	@Override
 	public String filterType() {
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
I added some tests in spring-cloud-samples/tests#9

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -210,10 +207,23 @@ private void addResponseHeaders() {
 		// Only inserts Content-Length if origin provides it and origin response is not
 		// gzipped
 		if (SET_CONTENT_LENGTH.get()) {
-			if (contentLength != null && !ctx.getResponseGZipped()) {
-				servletResponse.setContentLengthLong(contentLength);
+			if ( contentLength != null && !ctx.getResponseGZipped()) {
+				// To support Servlet API 3.0.1 we need to check if setcontentLengthLong exists
+				try {
+					HttpServletResponse.class.getMethod("setContentLengthLong");
```
spencergibb: Any potential performance problems using reflection here?

ryanjbaxter: This will only occur if `zuul.set-content-length` is set to true, but I don't see another good way of doing this.

ryanjbaxter:
@spencergibb I moved the reflection calls to the constructors.  We only need to figure out which version of the API is available once and then we can make the appropriate API call later on.  The reflection call now only happens once.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -56,6 +53,17 @@
...
 			.getBooleanProperty(ZuulConstants.ZUUL_SET_CONTENT_LENGTH, false);
+	private boolean useServlet31 = true;
+
+	public SendResponseFilter() {
+		super();
+		// To support Servlet API 3.0.1 we need to check if setcontentLengthLong exists
+		try {
+			HttpServletResponse.class.getMethod("setContentLengthLong");
+		} catch(NoSuchMethodException e) {
+			useServlet31 = false;
```
cbos: We face an issue that useServlet31 is always false. 
In that case if the content-lenght is bigger that max int there is not content-length value in the response headers. 
HttpServletResponse.class.getMethod("setContentLengthLong"); gives always a NoSuchMethodException, as there is not setContentLengthLong() method, as it always requires a long as input.

I think the check should be: 
HttpServletResponse.class.getMethod("setContentLengthLong", long.class)

spencergibb: please open an issue, rather than commenting on a merged pull request.

cbos: Thanks, I have done that, see #2196 
I added a comment here, as the issue started here, I was wondering if that had to be confirmed here.

Issue #1290, opened 2016-08-23
Best practices for using Eureka for 0-downtime Blue/Green deployments


william-tran:
This issue is more of a discussion that I want to open up to the community so we can figure this out. The goal is to use Eureka to facilitate blue-green deployments with 0 downtime. 
# Example 1: From application code

Say you have 2 instances of an app deployed using two distinct uris:

```
spring.application.name: api

---
spring.profiles: blue
eureka.instance.hostname: api-blue.example.com

---
spring.profiles: green
eureka.instance.hostname: api-green.example.com

```

Let's say both are up and healthy. You have a client app my-app that needs to make requests to api:

```
@RestController
@SpringBootApplication
@EnableDiscoveryClient
public class MyApp {

    @Bean
    @LoadBalanced
    public RestTemplate restTemplate() {
        return new RestTemplate();
    }

    @RequestMapping("/hit-api")
    public Object hitApi() {
        return restTemplate().getForObject("https://api/something", Object.class);
    }

```

When green-api.example.com goes down for upgrade, it gracefully shuts down and deletes its registration from Eureka. my-app won't know this happened until it fetches the registry again, which happens every 30 seconds by default. So for up to 30 seconds, the `@LoadBalanced` RestTemplate will send requests to `api-green.example.com` even though its down. 

There are a few things we can do:
## Use Ribbon's support for retry

This doesn't work out of the box because the `@LoadBalanced` RestTemplate doesn't use Netflix's RestClient and related classes, which provide the retry functionality. One of the reasons for not using it is because it's deprecated, and doesn't support PATCH.
## Use Spring Retry

By adding the following to my project:

```
    compile("org.springframework.boot:spring-boot-starter-aop")
    compile("org.springframework.retry:spring-retry")
```

```
// ...
@EnableRetry
public class MyApp {
// ...
    @Retryable
    @RequestMapping("/hit-api")
    public Object hitApi() {
// ...
```

A request made to `api-green.example.com` will error out, the RestTemplate with throw an exception, that exception breaks out of the `@Retryable` method and the method is executed again. Because the `@LoadBalanced` RestTemplate round robins, the next method execution tries a different server, api-blue.example.com which is still up. This works great. 

[...]

PR #1457, opened 2016-11-10
Use Spring Retry For Failed Requests Using Feign

ryanjbaxter:
Part of our continued effort to unify retry logic in Spring Cloud around Spring Retry.  Ties back to #1290.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
@@ -16,16 +16,16 @@
 
 package org.springframework.cloud.netflix.feign;
 
-import com.netflix.hystrix.HystrixCommand;
 import feign.Contract;
 import feign.Feign;
+import feign.Logger;
 import feign.Retryer;
 import feign.codec.Decoder;
 import feign.codec.Encoder;
 import feign.hystrix.HystrixFeign;
+
 import java.util.ArrayList;
 import java.util.List;
-
 import org.springframework.beans.factory.ObjectFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -42,17 +42,7 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.format.support.DefaultFormattingConversionService;
 import org.springframework.format.support.FormattingConversionService;
-
-import java.util.ArrayList;
-import java.util.List;
 import com.netflix.hystrix.HystrixCommand;
-
-import feign.Contract;
-import feign.Feign;
-import feign.Logger;
-import feign.codec.Decoder;
-import feign.codec.Encoder;
-import feign.hystrix.HystrixFeign;
 
 /**
  * @author Dave Syer
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
@@ -103,10 +103,16 @@ public class FeignClientsConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean
+	public Retryer feignRetryer() {
+		return Retryer.NEVER_RETRY;
+	}
+
+	@Bean
 	@Scope("prototype")
 	@ConditionalOnMissingBean
-	public Feign.Builder feignBuilder() {
-		return Feign.builder().retryer(Retryer.NEVER_RETRY);
+	public Feign.Builder feignBuilder(Retryer retryer) {
+		return Feign.builder().retryer(retryer);
 	}
 
 	@Bean
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
@@ -16,10 +16,13 @@
 
 package org.springframework.cloud.netflix.feign;
 
-import java.util.ArrayList;
-import java.util.List;
-
-import org.apache.http.client.HttpClient;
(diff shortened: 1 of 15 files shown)


ryanjbaxter:
This is not likely ready for prime time but I wanted to get some eyes on these changes.

ryanjbaxter:
spring-cloud/spring-cloud-sleuth#446 is dependent on this change

ryanjbaxter:
One last comment ;)

These changes will also need to be merged into the 1.3.x branch.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -16,6 +16,13 @@
 
 package org.springframework.cloud.netflix.feign;
 
+import com.netflix.hystrix.HystrixCommand;
```
spencergibb: Seems odd. Why a bunch of new imports?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
```diff
@@ -74,26 +95,32 @@ public RibbonResponse execute(RibbonRequest request, IClientConfig configOverrid
 		else {
 			options = new Request.Options(this.connectTimeout, this.readTimeout);
 		}
-		Response response = request.client().execute(request.toRequest(), options);
-		return new RibbonResponse(request.getUri(), response);
+		LoadBalancedRetryPolicy retryPolicy = loadBalancedRetryPolicyFactory.create(this.getClientName(), this);
+		retryTemplate.setRetryPolicy(retryPolicy == null ? new NeverRetryPolicy()
```
spencergibb: Seems like this should only be done once.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -71,7 +68,7 @@
 	 * Flag for whether retry is supported by default (assuming the routes themselves
 	 * support it).
 	 */
-	private Boolean retryable;
+	private Boolean retryable = false;
```
spencergibb: seems unrelated.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/CachingSpringLoadBalancerFactory.java
```diff
@@ -34,11 +35,16 @@
 public class CachingSpringLoadBalancerFactory {
 
 	private final SpringClientFactory factory;
+	private final RetryTemplate retryTemplate;
+	private final LoadBalancedRetryPolicyFactory loadBalancedRetryPolicyFactory;
 
 	private volatile Map<String, FeignLoadBalancer> cache = new ConcurrentReferenceHashMap<>();
 
-	public CachingSpringLoadBalancerFactory(SpringClientFactory factory) {
+	public CachingSpringLoadBalancerFactory(SpringClientFactory factory, RetryTemplate retryTemplate,
```
spencergibb: We should keep the old signature for backwards compatibility.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
```diff
@@ -16,8 +16,12 @@
 
 package org.springframework.cloud.netflix.ribbon.apache;
 
-import java.net.URI;
-
+import com.netflix.client.RequestSpecificRetryHandler;
```
spencergibb: You need to fix your ordering in a number of files ;-)

ryanjbaxter:
@spencergibb can you take another look when you get a chance?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -107,7 +106,7 @@ public FormattingConversionService feignConversionService() {
 	@Scope("prototype")
 	@ConditionalOnMissingBean
 	public Feign.Builder feignBuilder() {
-		return Feign.builder();
+		return Feign.builder().retryer(Retryer.NEVER_RETRY);
```
spencergibb: This should go in a `@Bean` definition.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/FeignLoadBalancer.java
spencergibb: I'm still confused about where the request is used.

ryanjbaxter: So you are saying that the policy should only be created once?  The problem is we need the `Request` when `FeignRetryPolicy.open` is called so we can set the `ServiceInstance` in the Retry Context.

spencergibb: I see, the retry policy created below uses the request. Could there be a problem since the retry policy is scoped to the request and the template is a field?

ryanjbaxter: Just trying to make sure I understand the concern....

If multiple requests are executed simultaneously you are concerned about switching out the retry policy mid request?

spencergibb: That's my worry

ryanjbaxter: I agree, I guess the only good solution for now is to create a new `RetryTemplate` for each request.  I don't see any real downside to that, but I do think it can be handled better by Spring Retry.  I posed this in issue spring-projects/spring-retry#71.

PR #1519, opened 2016-12-01
Use configuration from IConfigClient reference to always use the most up to date configuration

### LINKED ISSUE #1481: Ribbon timeout not dynamically changed with Zuul
Hello,
While configuring a config-server, I noticed a bug with Ribbon timeouts. I am trying to change the timeouts but It doesn't work. The property clearly changes (as seen in /env) but the values are not used during http calls. I tried with ribbon only and everything works fine, when I add Zuul, the reloading doesn't work anymore.

How to reproduce:
An example is available here: https://github.com/garciafl/spring-cloud-timeouts
1. Load the client app while config app is working
2. The timeout configuration is correctly set, a GET on /test /sleep on the client should be OK if timeout < 1s KO else.
3. Change the value in gitConfig/client.yml and push
4. POST on localhost:8080/refresh
5. If you hit the /test/sleep again, the new timeout configuration is not used (but /env shows the correct settings).

I browsed the code around the ribbon's clients `execute` method but It seems that timeout configurations are always set when building the http client. I can't find where the issue is.

If someone could explain me where to look, I am willing to help to fix this as quick as possible :)
Thank you

ryanjbaxter:
Fixes #1481.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
@@ -75,12 +75,13 @@ public class RibbonLoadBalancingHttpClient
 	public RibbonApacheHttpResponse execute(RibbonApacheHttpRequest request,
 			final IClientConfig configOverride) throws Exception {
 		final RequestConfig.Builder builder = RequestConfig.custom();
-		if (configOverride != null) {
-			builder.setConnectTimeout(configOverride.get(
+		IClientConfig config = configOverride != null ? configOverride : this.config;
+		if (config != null) {
+			builder.setConnectTimeout(config.get(
 					CommonClientConfigKey.ConnectTimeout, this.connectTimeout));
-			builder.setSocketTimeout(configOverride.get(
+			builder.setSocketTimeout(config.get(
 					CommonClientConfigKey.ReadTimeout, this.readTimeout));
-			builder.setRedirectsEnabled(configOverride.get(
+			builder.setRedirectsEnabled(config.get(
 					CommonClientConfigKey.FollowRedirects, this.followRedirects));
 		}
 		else {
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
@@ -76,19 +76,12 @@ public class RibbonLoadBalancingHttpClient
 			final IClientConfig configOverride) throws Exception {
 		final RequestConfig.Builder builder = RequestConfig.custom();
 		IClientConfig config = configOverride != null ? configOverride : this.config;
-		if (config != null) {
-			builder.setConnectTimeout(config.get(
-					CommonClientConfigKey.ConnectTimeout, this.connectTimeout));
-			builder.setSocketTimeout(config.get(
-					CommonClientConfigKey.ReadTimeout, this.readTimeout));
-			builder.setRedirectsEnabled(config.get(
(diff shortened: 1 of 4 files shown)


ryanjbaxter:
Will also need to be merged into master

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
```diff
@@ -75,12 +75,13 @@ protected HttpClient createDelegate(IClientConfig config) {
 	public RibbonApacheHttpResponse execute(RibbonApacheHttpRequest request,
 			final IClientConfig configOverride) throws Exception {
 		final RequestConfig.Builder builder = RequestConfig.custom();
-		if (configOverride != null) {
-			builder.setConnectTimeout(configOverride.get(
+		IClientConfig config = configOverride != null ? configOverride : this.config;
+		if (config != null) {
```
ryanjbaxter: Wondering if config can ever be `null` now?  It doesn't appear to be the case now, I think we can get rid of this `if/else`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/OkHttpLoadBalancingClient.java
```diff
@@ -85,12 +83,13 @@ public OkHttpRibbonResponse execute(OkHttpRibbonRequest ribbonRequest,
 
 	OkHttpClient getOkHttpClient(IClientConfig configOverride, boolean secure) {
 		OkHttpClient.Builder builder = this.delegate.newBuilder();
-		if (configOverride != null) {
-			builder.connectTimeout(configOverride.get(
+		IClientConfig config = configOverride != null ? configOverride : this.config;
+		if (config != null) {
```
ryanjbaxter: Wondering if config can ever be `null` now?  It doesn't appear to be the case now, I think we can get rid of this `if/else`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
spencergibb: If we can add final to the field in the abstract class, then yes.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/OkHttpLoadBalancingClient.java
```diff
@@ -16,22 +16,20 @@
 
 package org.springframework.cloud.netflix.ribbon.okhttp;
 
+import okhttp3.OkHttpClient;
```
spencergibb: ordering

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/okhttp/OkHttpLoadBalancingClientTests.java
```diff
@@ -16,21 +16,20 @@
 
 package org.springframework.cloud.netflix.ribbon.okhttp;
 
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
+import okhttp3.OkHttpClient;
```
spencergibb: Ordering

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java
ryanjbaxter: it already is :)

spencergibb: Remove away!

PR #1528, opened 2016-12-05
Don’t add URL protocol when url is set using EL

### LINKED ISSUE #1522: Feign client for external service adds addition protocol to url
I noticed, that Feign client for external service adds addition protocol to url, when using EL.

For that configuration:
```
@Configuration
@EnableAutoConfiguration
@EnableFeignClients(basePackages = "com.example.connector")
public class ApplicationConfig {
    @Bean(name = "EXAMPLE_URL")
    public String prUrl() {
        return "http://example.com/api";
    }
}
```
And using that client
```
@FeignClient(name = "exampleClient", url = "#{EXAMPLE_URL}" )
public interface ExampleClient{
    @RequestMapping(value = "/example", method = RequestMethod.GET)
    String getExampleValue();
}
```
The service URL is set to `http://http://example.com/api/example` instead of `http://example.com/api/example`.

ryanjbaxter:
Fixes #1522.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsRegistrar.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsRegistrar.java
@@ -243,7 +243,7 @@ class FeignClientsRegistrar implements ImportBeanDefinitionRegistrar,
 
 	private String getUrl(Map<String, Object> attributes) {
 		String url = resolve((String) attributes.get("url"));
-		if (StringUtils.hasText(url)) {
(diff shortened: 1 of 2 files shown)


chaluja7:
Hi, a was looking at the approved fix from 2016-12-05 and noticed, that it will not handle the situation, when one would like to define url this way:

`@FeignClient(name = "exampleClient", url = "#{EXAMPLE_URL}/api/something" )`

because of `url.endsWith("}")` control. Would it be possible to solve this situation as well? Should I create new issue for this?

ryanjbaxter:
@chaluja7 thanks!  I suppose just checking fro `url.startsWith("#{")` is good enough.  Please create a new issue.

Issue #1534, opened 2016-12-06
Eureka: EurekaDiscoveryClient stopped when *any* Spring ApplicationContext is closed


brenuart:
`org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration` listens for `ContextClosedEvent` and stops the discovery client when such event is received, even if the context being closed is not the primary context.

The `onApplicationEvent()` method should be modified as follows:
```

	@EventListener(ContextClosedEvent.class)
	public void onApplicationEvent(ContextClosedEvent event) {
		if( event.getApplicationContext() == this.applicationContext ) {
		   // register in case meta data changed
		   stop();
		}
	}
```

brenuart:
Problem still there in `Dalston.SR1`. See PR above for a fix.

spencergibb:
Closed via #2087

brenuart:
This fix doesn't seem to be part of `1.3.2.RELEASE` but `2.0.0.M2` - see https://github.com/spring-cloud/spring-cloud-netflix/pull/2087

ryanjbaxter:
It was merged into master so it will be part of the 1.4.0.M1 release and yes part of the 2.0.0.M2 release.  Do you need it in the 1.3.x stream as well?  If so we can consider backporting it.

brenuart:
That's ok for us - we already have a workaround in place. I was just checking for when it could be removed. We can wait for `1.4` to be released. thx

dlts26:
1.5.6依然有这个问题？

changyangyo:
This problem still exists in version 1.4.12

PR #1535, opened 2016-12-06
Stop discovery client only when the context that owns it is closed

brenuart:
See https://github.com/spring-cloud/spring-cloud-netflix/issues/1534


spencergibb:
@brenuart this functionality has moved to `EurekaAutoServiceRegistration`. Would you rebase and update?

spencergibb:
ping @brenuart

Issue #1272, opened 2016-08-18
Allow setting LoadBalancerCommand.Builder.withServerLocator from request


lowzj:
I'm trying to use the FeignClient to send HTTP requests to target service.
The target service may have several instances grouped by different versions, so I want to write a class implemented `IRule` to choose a service instance based on a key that in the request parameters. And the 'IRule' implementation just like this:

```
public class MyLoadBalancerRule extends AbstractLoadBalancerRule {
    @Override
    public void initWithNiwsConfig(IClientConfig clientConfig) {
    }

    @Override
    public Server choose(Object key) {
        List<Server> servers = getLoadBalancer().getReachableServers();
        // choose and return a server from servers based on the key
    }
}
```

but the key is always `null`.

**LoadBalancerCommand** passes its attribute `loadBalancerKey` as parameter to the `choose` method. And **LoadBalancerCommand** object is constructed in `AbstractLoadBalancerAwareClient#executeWithLoadBalancer` without setting the attribute `loadBalancerKey`. 
So how to set/use the `loadBalancerKey` when I use FeignClient?

dsyer:
As far as I can see `LoadBalancerCommand` always has a null `loadBalancerKey` in all the out-of -the box usages, despite having a builder that allows it to be set. I guess you'd have to implement your own load balancer aware stack based on feign requests as a context/request object.

lowzj:
Thank for your response. I'm trying a solution to get the loadBalancerKey.

At first I write an annotation `@LoadBalancerKey` to specified an loadBalancerKey, then I can write the FeignClient interface like this:

```
@FeignClient(name = "UserService")
public interface UserFeignClient {
    @RequestMapping(value = "/user", method = RequestMethod.GET)
    ResponseEntity<User> getUser(@LoadBalancerKey @RequestParam("name") String name);
}
```
Secondly, I implement an `AnnotatedParameterProcessor` named `LoadBalancerKeyParameterProcessor` to process the `@LoadBalancerKey`. The implementation is very similar to `RequestHeaderParameterProcessor`. I add the `loadBalancerKey` to the request header, the key is `loadBalancerKey`. So I can get the `loadBalancerKey` from the request header.

Then I rewrite the **FeignLoadBalancer.RibbonRequest** constructor and **AbstractLoadBalancerAwareClient#executeWithLoadBalancer**:

```
// FeignLoadBalancer.RibbonRequest, set loadBalancerKey
[...]

PR #1556, opened 2016-12-16
setting LoadBalancerCommand.Builder.withServerLocator to support loadBalancerKey

lowzj:
https://github.com/spring-cloud/spring-cloud-netflix/issues/1272

This PR is just to allow load balancer client to get `loadBalancerKey` from `ClientRequest`.


spencergibb:
This is pretty invasive. I'm not sure this is a direction we want to go.

lowzj:
Yes. Another way I can think out is to update the upstream code `Ribbon`. But `Ribbon` is currently in maintenance mode,  code changes may affect the existing status quo that they want to avoid.

lowzj:
Since the upstream code has been changed, I will submit another PR to support `loadBalancerKey`(how: https://github.com/spring-cloud/spring-cloud-netflix/issues/1272#issuecomment-303087679) ASAP when there is a new release version in ribbon.

PR #1561, opened 2016-12-19
Create `sidecar.hostname` and `sidecar.ipAddress`

### LINKED ISSUE #981: Sidecar - Ability to control the hostname
This feature stems from this [stack overflow issue](http://stackoverflow.com/questions/36763171/how-spring-sidecar-works-with-docker/).

Sidecar makes the assumption that the spring boot sidecar application and the non-jvm application that you're bridging are running on the same host. Sometimes this isn't always the case.. especially if your running everything in an containerized environment with Docker.

As well as controlling the sidecar port via `sidecar.port` property, could we also have a `sidecar.hostname` property to control which host the sidecar application is running on too?

Thanks.

kakawait:
In addition to `eureka.instance.hostname` property you can now defined `sidecar.hostname` in order to set different hostname from application and its sidecar.
However if `eureka.instance.hostname` is present it will override `sidecar.hostname` (due to ` @ConfigurationProperties("eureka.instance")`)

Moreover we also now re-calculate ip address from `sidecar.hostname` (except if `sidecar.ipAddress` is defined) to correctly have ip address on eureka registry.

Fixes #981 

---

PS: About tests I can't really test ip recalculation because I'm not able to mock `InetAddress.getByName()` without `PowerMock` or something else.

### DIFF
--- a/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
+++ b/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
@@ -20,6 +20,7 @@ import static org.springframework.cloud.commons.util.IdUtils.getDefaultInstanceI
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.bind.RelaxedPropertyResolver;
@@ -35,9 +36,25 @@ import org.springframework.util.StringUtils;
 import com.netflix.appinfo.HealthCheckHandler;
 import com.netflix.discovery.EurekaClientConfig;
 
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
 /**
+ * Sidecar Configuration that setting up {@link com.netflix.appinfo.EurekaInstanceConfig}.
+ * <p>
+ * Depends on {@link SidecarProperties} and {@code eureka.instance.hostname} property. Since there is two way to
+ * configure hostname:
+ * <ol>
+ *   <li>{@code eureka.instance.hostname} property</li>
+ *   <li>{@link SidecarProperties#hostname}</li>
+ * </ol>
+ * {@code eureka.instance.hostname} will always win against {@link SidecarProperties#hostname} due to
+ * {@code @ConfigurationProperties("eureka.instance")} on {@link EurekaInstanceConfigBeanConfiguration}.
+ *
  * @author Spencer Gibb
  * @author Ryan Baxter
+ *
+ * @see EurekaInstanceConfigBeanConfiguration
  */
 @Configuration
 @EnableConfigurationProperties
@@ -80,13 +97,29 @@ public class SidecarConfiguration {
 			int port = this.sidecarProperties.getPort();
 			config.setNonSecurePort(port);
 			config.setInstanceId(getDefaultInstanceId(this.env));
-			if(StringUtils.hasText(springAppName)) {
+			if (StringUtils.hasText(springAppName)) {
 				config.setAppname(springAppName);
 				config.setVirtualHostName(springAppName);
 				config.setSecureVirtualHostName(springAppName);
 			}
-			if (StringUtils.hasText(this.hostname)) {
-				config.setHostname(this.hostname);
+			String hostname = this.sidecarProperties.getHostname();
+			String ipAddress = this.sidecarProperties.getIpAddress();
+			if (!StringUtils.hasText(hostname) && StringUtils.hasText(this.hostname)) {
+				hostname = this.hostname;
+			}
+			if (StringUtils.hasText(hostname)) {
+				if (!StringUtils.hasText(ipAddress)) {
+					try {
+						ipAddress = inetUtils.convertAddress(InetAddress.getByName(hostname)).getIpAddress();
+					} catch (UnknownHostException e) {
+						throw new IllegalStateException(
+								"Could not resolve IP address of sidecar application using hostname: " + hostname);
+					}
+				}
+				config.setHostname(hostname);
+			}
+			if (!StringUtils.hasText(ipAddress)) {
+				config.setIpAddress(ipAddress);
 			}
 			String scheme = config.getSecurePortEnabled() ? "https" : "http";
 			config.setStatusPageUrl(scheme + "://" + config.getHostname() + ":"
--- a/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
+++ b/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
@@ -118,7 +118,7 @@ public class SidecarConfiguration {
 				}
 				config.setHostname(hostname);
 			}
-			if (!StringUtils.hasText(ipAddress)) {
+			if (StringUtils.hasText(ipAddress)) {
 				config.setIpAddress(ipAddress);
 			}
 			String scheme = config.getSecurePortEnabled() ? "https" : "http";
--- a/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
+++ b/spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
@@ -108,14 +108,6 @@ public class SidecarConfiguration {
 				hostname = this.hostname;
 			}
 			if (StringUtils.hasText(hostname)) {
-				if (!StringUtils.hasText(ipAddress)) {
-					try {
-						ipAddress = inetUtils.convertAddress(InetAddress.getByName(hostname)).getIpAddress();
-					} catch (UnknownHostException e) {
-						throw new IllegalStateException(
-								"Could not resolve IP address of sidecar application using hostname: " + hostname);
-					}
-				}
 				config.setHostname(hostname);
 			}
 			if (StringUtils.hasText(ipAddress)) {
--- a/spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/SidecarApplicationTests.java
+++ b/spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/SidecarApplicationTests.java
@@ -31,8 +31,8 @@ public class SidecarApplicationTests {
 
 	@RunWith(SpringJUnit4ClassRunner.class)
 	@SpringBootTest(classes = SidecarApplication.class, webEnvironment = WebEnvironment.RANDOM_PORT, value = {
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@kakawait Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1561)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1561) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kakawait Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1561)!

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
```diff
@@ -80,13 +97,29 @@ public EurekaInstanceConfigBean eurekaInstanceConfigBean() {
 			int port = this.sidecarProperties.getPort();
 			config.setNonSecurePort(port);
 			config.setInstanceId(getDefaultInstanceId(this.env));
-			if(StringUtils.hasText(springAppName)) {
+			if (StringUtils.hasText(springAppName)) {
```
ryanjbaxter: rougue space before `if`, same with the ones below

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
```diff
@@ -80,13 +97,29 @@ public EurekaInstanceConfigBean eurekaInstanceConfigBean() {
...
+						ipAddress = inetUtils.convertAddress(InetAddress.getByName(hostname)).getIpAddress();
+					} catch (UnknownHostException e) {
+						throw new IllegalStateException(
+								"Could not resolve IP address of sidecar application using hostname: " + hostname);
+					}
+				}
+				config.setHostname(hostname);
+			}
+			if (!StringUtils.hasText(ipAddress)) {
+				config.setIpAddress(ipAddress);
```
ryanjbaxter: why set the IP address if there is nothing to set?

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
```diff
@@ -80,13 +97,29 @@ public EurekaInstanceConfigBean eurekaInstanceConfigBean() {
...
-				config.setHostname(this.hostname);
+			String hostname = this.sidecarProperties.getHostname();
+			String ipAddress = this.sidecarProperties.getIpAddress();
+			if (!StringUtils.hasText(hostname) && StringUtils.hasText(this.hostname)) {
+				hostname = this.hostname;
+			}
+			if (StringUtils.hasText(hostname)) {
+				if (!StringUtils.hasText(ipAddress)) {
+					try {
+						ipAddress = inetUtils.convertAddress(InetAddress.getByName(hostname)).getIpAddress();
```
ryanjbaxter: This is going to make a network request correct?  Have you thought about the performance impacts of this?  I guess there isn't a better way of doing this if you really need the information.  Maybe we want to consider only doing this if the user configures the sidecar app to do so?

spencergibb: We already have code that attempts to detect the hostname and ipaddress in spring cloud commons. Please remove this and stick to just setting the value if it is available.

spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/SidecarApplicationTests.java
```diff
@@ -27,22 +27,60 @@
...
-		"spring.application.name=mytest", "spring.cloud.client.hostname=mhhost",
-		"spring.application.instance_id=1", "eureka.instance.hostname=mhhost",
-		"sidecar.port=7000" })
 public class SidecarApplicationTests {
 
-	@Autowired
-	EurekaInstanceConfigBean config;
+	@RunWith(SpringJUnit4ClassRunner.class)
+	@SpringBootTest(classes = SidecarApplication.class, webEnvironment = WebEnvironment.RANDOM_PORT, value = {
+			"spring.application.name=mytest", "spring.cloud.client.hostname=mhhost",
```
spencergibb: `spring.cloud.client.hostname` doesn't map to any config. My guess is these tests work find without it.

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
kakawait: I did that because if you are registering sidecar using previous trick `eureka.instance.hostname` (in order to target application on different hostname) we get following result on Eureka server http://localhost:8761/eureka/apps

Using following configuration (tested on `Camden.SR3`)

```yml
spring:
  application:
    name: sidecar

server:
  port: 5678

sidecar:
  port: 8080

eureka:
  instance:
    hostname: google.com
```

returns

```xml
<applications>
    <versions__delta>1</versions__delta>
    <apps__hashcode>UP_1_</apps__hashcode>
    <application>
        <name>SIDECAR</name>
        <instance>
            <instanceId>127.0.0.1:sidecar:5678</instanceId>
            <hostName>google.com</hostName>
            <app>SIDECAR</app>
            <ipAddr>127.0.0.1</ipAddr>
            <status>UP</status>
            ...
        </instance>
    </application>
</applications>
```

As you can see `hostName` is correctly set but ip address does not correspond to the hostname.

Thus if you are using `eureka.instance.preferIpAddress` or other way can fail.

So if I removed the hostname resolution and client do not provide ip address via `sidecar.ipAddress` (example: `Docker` you can determine hostname but generally not the ip address that will be generated when container will be run) you assume that Eureka information will be partially wrong?

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
kakawait: Lol I may fail something 👍  I will check that

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
kakawait: What do you mean? Sorry I can't understand since https://github.com/spring-cloud/spring-cloud-netflix/blob/master/eclipse/eclipse-code-formatter.xml describe that `if` must be followed by space before brace

spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/SidecarApplicationTests.java
kakawait: Yes just copied from original test. Didn't ask myself the usage of that property. I will remove it

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
kakawait: @ryanjbaxter Yes performance will be impacted (and can fail if hostname can't be resolve when sidecar will be starting) but as I said before I favored consistency rather than performance. But let's discuss together!

Since I'm not mastering Eureka yet I may miss-understand something about its concepts, or even my sidecar vision may not be the same.

spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/SidecarApplicationTests.java
kakawait: After removed it used for `instanceId` generation if `eureka.instance.hostname` is not present so `NewPropertyEurekaTestConfigBeanTest`

```
Expected: "mhhost:mytest:1"
     but: was "127.0.0.1:mytest:1"
Expected :mhhost:mytest:1
Actual   :127.0.0.1:mytest:1
```

Thus I can't change assertion to match regex (because ip can be different depending on environment, in my case I have an alias for `lo0`) or simply rollback?

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
ryanjbaxter: > example: Docker you can determine hostname but generally not the ip address that will be generated when container will be run

@kakawait I could be wrong, but I thought that Docker sets the IP address of the container in an environment variable.  I am pretty sure that was the case when using Docker compose.

Is there another situation besides Docker where this is an issue?

kakawait: @ryanjbaxter What can I propose: remove hostname resolution but update documentation that inform user `sidecar.ipAddress` should be fill in addition to `sidecar.hostname` to prevent any futur error when trying to use `ipAddress` from `Eureka` meta informations? Like `eureka.instance.preferIpAddress`

kakawait: About docker by default (with basic `run` param) there is no exposed variable that contain `ipAddress`

```
$ docker run -it --rm alpine env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=60db1f4ab975
TERM=xterm
HOME=/root
```

spencergibb: We already have code to do this [here](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/InetUtils.java).  It is already used in the constructor of [`EurekaInstanceConfigBean`](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaInstanceConfigBean.java#L291-L296). Please remove this code.

kakawait: Yes but I will apply on server that instanciate the bean so the sidecar itself and not the application referenced by the sidecar. I'm OK to remove it but I think is important to inform user that it should specify IP address himself else sidecar IP will be referenced and not application

spencergibb: Sidecar was built to run as a process next to the application itself.

kakawait: Thus this PR is no sense, since they should share same IP address so technically same hostname

Issue #1546, opened 2016-12-12
Get InstanceInfo in EurekaInstanceRenewedEvent is always Null


darren-fu:
Try to get InstanceInfo in Listener<EurekaInstanceRenewedEvent>, but it's always null,  debug the code,  find out there is a bug here.
```java
//org.springframework.cloud.netflix.eureka.server.InstanceRegistry#renew 
// line 117
if (input.getName().equals(appName)) {
				InstanceInfo instance = null;
				for (InstanceInfo info : input.getInstances()) {
					if (info.getHostName().equals(serverId)) {
						instance = info;
						break;
					}
				}
				this.ctxt.publishEvent(new EurekaInstanceRenewedEvent(this, appName,
						serverId, instance, isReplication));
				break;
			}
```
**if (info.getHostName().equals(serverId))**
serverId contains HostName and Port, so it won't equal to info.getHostName() forever.

Hope it's helpful!

darren-fu:
Version : Brixton.SR7

ryanjbaxter:
Can you provide a sample that reproduces this?  Or better yet a PR with a fix and a test case.

PR #1574, opened 2016-12-23
Fixes issue #1546 - null instanceinfo on EurekaInstanceRenewedEvent

bijukunjummen:
This is a fix for #1546, the check for all instances of app against the renewed instance is now made using the id - this is more reliable than comparing the serverId against the app name.


spencergibb:
cherry picked to Brixton 1.x

Issue #556, opened 2015-09-25
Support Spring Data Pageable in Feign Client


edethoor:
I am using Feign for requesting a MicroService who support Spring Data Pageable functionnality.

This is my FeignClient Interface : 

``` java
 @FeignClient(FeignServiceId.SERVICE_ID)
 @RequestMapping(value = "api/document")
 public interface DocumentApi {

    @RequestMapping(method = RequestMethod.GET, value = "user/{userLogin}")
    Page<DeclarationDT> getDeclarationsByUserLogin(@PathVariable("userLogin") String userLogin, Pageable pageable);

}
```

When i make a call to the `getDeclarationsByUserLogin`, Feign do a POST request whereas i specify `RequestMethod.GET`.

I think this is due to the fact that Feign not support the Spring Data Pageable functionnality.

Is it possible to implement support for this feature ?

spencergibb:
Actually, it thinks your `Pageable` is the body. Feign assumes post with a body.

spencergibb:
@adriancole is there a way to add custom requests parameter mappings or some such?

adriancole:
I think in denominator I just made a custom iterator which kept a marker.

Like this
https://github.com/Netflix/denominator/blob/master/route53/src/main/java/denominator/route53/Route53ZoneApi.java#L41

adriancole:
Oops I meant this ;)
https://github.com/Netflix/denominator/blob/master/route53/src/main/java/denominator/route53/Route53AllProfileResourceRecordSetApi.java#L120

Dreampie:
When can solve this problem?

damienpolegato:
Hi, I'm also interested in this issue. I didn't quite get @adriancole example. Can we register a conversion somewhere ?
Thanks !

damienpolegato:
My workaround to get it work is :
1. Use PageRequest instead of Pageable
2. On backend side register a converter in ConverterRegistry in a `@Configuration` class to convert String to PageRequest
`@Bean
    public Void addPageRequestConverter(ConverterRegistry converterRegistry) {
        converterRegistry.addConverter(new StringToPageRequestConverter());
        return null;
    }`
3. Define my own Page POJO since PageImpl doesn't have noargs constructor and setters
4. Implement a custom Jackson deserializer for `Sort`

So far at least that works.

Would 3 and 4 be handled by using a HATEOAS PagedResources ?

adriancole:
[...]

PR #1604, opened 2017-01-10
Feign spring pagination support

IsNull:
Support using spring Pageable and Page<T> in Feign clients as requested in #556.

```java
@RequestMapping(value = "invoicesPaged", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
ResponseEntity<Page<Invoice>> getInvoicesPaged(Pageable pageable);
```

To support this, a Pageable Encoder and a Page Jackson Decoder has been added.

### DIFF
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -72,6 +72,7 @@
 		<dependency>
 			<groupId>org.springframework.data</groupId>
 			<artifactId>spring-data-commons</artifactId>
+			<optional>true</optional>
 		</dependency>
 		<!-- archaius dependencies that are now runtime -->
 		<dependency>
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -69,6 +69,11 @@
 			<artifactId>archaius-core</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-commons-core</artifactId>
+			<version>${spring-boot.version}</version>
+		</dependency>
 		<!-- archaius dependencies that are now runtime -->
 		<dependency>
 			<groupId>commons-configuration</groupId>
--- a/spring-cloud-netflix-core/pom.xml
+++ b/spring-cloud-netflix-core/pom.xml
@@ -71,8 +71,7 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.data</groupId>
-			<artifactId>spring-data-commons-core</artifactId>
-			<version>${spring-boot.version}</version>
+			<artifactId>spring-data-commons</artifactId>
 		</dependency>
 		<!-- archaius dependencies that are now runtime -->
 		<dependency>
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.netflix.feign;
 import java.util.ArrayList;
 import java.util.List;
 
+import com.fasterxml.jackson.databind.Module;
 import org.apache.http.client.HttpClient;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -26,6 +27,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cloud.client.actuator.HasFeatures;
+import org.springframework.cloud.netflix.feign.support.PageJacksonModule;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
@@ -55,6 +57,11 @@ public class FeignAutoConfiguration {
 		FeignContext context = new FeignContext();
 		context.setConfigurations(this.configurations);
 		return context;
+	}
+
+	@Bean
+	public Module pageJacksonModule(){
+		return new PageJacksonModule();
 	}
 
 	@Configuration
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignAutoConfiguration.java
@@ -35,6 +35,7 @@ import feign.Client;
 import feign.Feign;
 import feign.httpclient.ApacheHttpClient;
 import feign.okhttp.OkHttpClient;
+import org.springframework.data.domain.Page;
 
 /**
  * @author Spencer Gibb
@@ -60,6 +61,7 @@ public class FeignAutoConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnClass(Page.class)
 	public Module pageJacksonModule(){
 		return new PageJacksonModule();
 	}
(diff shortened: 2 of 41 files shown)


IsNull:
The tests currently fail since the assumption that the `Encoder` is an instance of `SpringEncoder` is no longer true:

>encoderDefaultCorrect(org.springframework.cloud.netflix.feign.EnableFeignClientsTests): Cannot cast org.springframework.cloud.netflix.feign.support.PageableSpringEncoder to org.springframework.cloud.netflix.feign.support.SpringEncoder
  overrideEncoder(org.springframework.cloud.netflix.feign.FeignClientOverrideDefaultsTests): Cannot cast org.springframework.cloud.netflix.feign.support.PageableSpringEncoder to org.springframework.cloud.netflix.feign.support.SpringEncoder

I could of course rework the PageableEncoder and hide it inside the `SpringEncoder`, maybe make it pluggable. What do you think?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -64,6 +62,7 @@
 	@Autowired(required = false)
 	private Logger logger;
 
+
```
ryanjbaxter: Remove rogue new line

spring-cloud-netflix-core/pom.xml
```diff
@@ -69,6 +69,10 @@
 			<artifactId>archaius-core</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
```
ryanjbaxter: should this be optional like the other dependencies?

IsNull: Now this is a good question. If I understand the optional behaviour correctly, then if the user does not specify it (directly or indirectly) he wont have this dependency. If this is the case, then the current implementation would fail horribly due to a ClassNotFoundException in

```java
public boolean supports(Object object, Type bodyType, RequestTemplate template){
        return object instanceof Pageable;
    }
```
Since `Pageable` wont be available.


So would it require a ugly check like the one below, before using a class file which has an import to the optional dependency?
```java
try {
 Class.forName( "org.springframework.data.domain.Pageable" );
} catch( ClassNotFoundException e ) {
 // not here
}
```
If so, is this really desired?

ryanjbaxter: No, you can use Spring Beans and check is the `Pageable` class is available and then include/exclude encoders based on that.

IsNull: Ah yes, this is simpler. I give it a try.

IsNull:
@ryanjbaxter I noticed several places where the Encoder is hardcoded to `SpringEncoder` in the tests.

```java
@Test
public void encoderDefaultCorrect() {
	SpringEncoder.class.cast(this.feignContext.getInstance("foo", Encoder.class));
}

// ...

@Test
public void overrideEncoder() {
	Encoder.Default.class.cast(this.context.getInstance("foo", Encoder.class));
	SpringEncoder.class.cast(this.context.getInstance("bar", Encoder.class));
}

// ...

@Test
public void testCustomHttpMessageConverter() {
	SpringEncoder encoder = this.context.getInstance("foo", SpringEncoder.class);
	assertThat(encoder, is(notNullValue()));
	RequestTemplate request = new RequestTemplate();

	encoder.encode("hi", MyType.class, request);

	Collection<String> contentTypeHeader = request.headers().get("Content-Type");
	assertThat("missing content type header", contentTypeHeader, is(notNullValue()));
	assertThat("missing content type header", contentTypeHeader.isEmpty(), is(false));

	String header = contentTypeHeader.iterator().next();
	assertThat("content type header is wrong", header, is("application/mytype"));
}
```

I think the tests should not depend on the specific type. If we need to be sure it is an encoder from Spring there might be another way. But I am not sure about the benefit of these tests.

Since my Encoder pipe will by built dynamically, I also don't know which Encoder will be instantiated so these tests are really annoying.

spring-cloud-netflix-core/pom.xml
spencergibb: +1 to the optional. use `@ConditionalOnClass` in the auto configuration.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignAutoConfiguration.java
```diff
@@ -57,6 +59,11 @@ public FeignContext feignContext() {
 		return context;
 	}
 
+	@Bean
```
spencergibb: A `@ConditionalOnClass` should probably go here as well.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientsConfiguration.java
```diff
@@ -71,12 +70,20 @@ public Decoder feignDecoder() {
...
 
 	@Bean
+	@ConditionalOnMissingClass("org.springframework.data.domain.Pageable")
 	@ConditionalOnMissingBean
 	public Encoder feignEncoder() {
 		return new SpringEncoder(this.messageConverters);
 	}
 
 	@Bean
+	@ConditionalOnClass(name = "org.springframework.data.domain.Pageable")
```
spencergibb: You can use the actual class here, boot handles it just fine.

IsNull: Thank you for the info, I will change it.

ryanjbaxter:
So are those tests now failing?  I would assume the Encoder would not be effected in the existing tests.  If not it might be an indication your changes broke something...

IsNull:
@ryanjbaxter Yes the tests are failing and I understand why and explained the reason above (at least I tried) In my opinion the tests need to be fixed, so before I do that Id like to consult with you :)

The new `Encoder` configuration looks like this and is the reason why some tests fail:
```java
@Bean
@ConditionalOnMissingClass("org.springframework.data.domain.Pageable")
@ConditionalOnMissingBean
public Encoder feignEncoder() {
	return new SpringEncoder(this.messageConverters);
}

@Bean
@ConditionalOnClass(Pageable.class)
@ConditionalOnMissingBean
public Encoder feignEncoderPageable() {
	return new PageableSpringEncoder(new SpringEncoder(this.messageConverters));
}
```

Now, there are 3 tests which expect the `Encoder` to be `SpringEncoder`, which is IMHO too specific and prevents me to do it this way. 

On a similar note, the Decoder already exposes something different than the SpringDecoder:
```java
@Bean
@ConditionalOnMissingBean
public Decoder feignDecoder() {
	return new ResponseEntityDecoder(new SpringDecoder(this.messageConverters));
}
```

So if you and @spencergibb are fine with my implementation, I go ahead and fix the tests. I just wasn't sure if there is any other reasoning why the tests explicitly expect a `SpringDecoder` instance.

ryanjbaxter:
They probably expect `SpringDecorder` because there was no other decoder.  I would expect that `SpringDecorder` would still be returned in those tests.  Is `PageableSpringEncoder` always being returned now since `Pageable` is on the classpath?

IsNull:
> They probably expect SpringDecorder because there was no other decoder.

If you mean Spring**Encoder** then yes, the tests seem that way.
(I did not change anything about the Spring**Decoder** and as a matter of fact, the tests are already not expecting `SpringDecoder` but the enhanced `ResponseEntityDecoder`)

> I would expect that SpringDecorder would still be returned in those tests. 

Well, since the `SpringDecoder` is also not returned, it would be inconsequent to imply that the Encoder must always be the `SpringEncoder`.

>Is PageableSpringEncoder always being returned now since Pageable is on the classpath?

No as per your request above, the Pageable dependency (spring-data-commons) is now optional, and thus the actual Encoder instance depends on whenever `Pageable` is available or not.
The current implementation will either yield `PageableSpringEncoder` or the original `SpringEncoder`.

IsNull:
I have fixed the unit tests, please have a look if this is feasible.

darioajr:
Any Evolution ?

IsNull:
There wont be any more work from my side. The last requested unit tests scared me off for good :)
We opted in for using shared library which is adding the required support.

darioajr:
@IsNull You can share this library ?

IsNull:
@darioajr Yes you can find it here [ElderByte-/spring-cloud-starter-bootstrap](https://github.com/ElderByte-/spring-cloud-starter-bootstrap)

PR #1617, opened 2017-01-12
Compose Simple- and DiscoveryClientRouteLocator

joshiste:
As requested #1448 was split up in two parts. This is the follow up with the second commit.

With this commit an AbstractZuulRouteLocator is added. This allows us
to remove the logic for simple routes from DiscoveryClientRouteLocator
by not extending the SimpleRouteLocator. The behaviour for the simple
and discovery routes is from now on composed and not inherited. This
follows the "favor composition over inheritance" principle and makes
the code (especially for the DiscoveryClientRouteLocator) a bit less
twisted.

### DIFF
--- a/spring-cloud-netflix-core/.gitignore
+++ b/spring-cloud-netflix-core/.gitignore
@@ -1,0 +1,1 @@
+/.apt_generated/
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/AbstractZuulRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/AbstractZuulRouteLocator.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2013-2014 the original author or authors.
+ *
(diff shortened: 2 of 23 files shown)


joshiste:
*sigh* ... I'll fix the build...

joshiste:
fixed.

spencergibb:
So reviewing this a bit more. I'm not super convinced this is helpful. Most of the functionality just moved from Simple -> Abstract. I don't see big benefits from this. I was thinking more along the lines of a true composite.

joshiste:
Problem here was both, the simple and the discovery rules both depend heavily on the `ZuulRoute`. 
Imho to decompose the RouteLocator in a true composite it may be needed to have a "StaticZuulRoute" and a "DiscoveryZuulRoute" but this would be a lot more to do. Imho the current ZuulRoute is a bit unfortunate in supplying both concepts (e.g. `getLocation()`), but decomposing the ZuulRoute would mean to break compatibility for the configuration.

spencergibb:
Sure, a true composite would be much more work.  What is the advantage of the changes here?

spencergibb:
You should be able to do this on your own with the interface.

PR #1622, opened 2017-01-16
optimize SendResponseFilter

lowzj:
optimize `SendResponseFilter#addResponseHeaders()`:
* remove needless statement `RequestContext.getCurrentContext()` except the first one
* testing dynamic config option at first before `ConcurrentHashMap#get`: testing `INCLUDE_DEBUG_HEADER` before `context.get("routingDebug")`, testing `SET_CONTENT_LENGTH` before `context.getOriginContentLength()`
  * most of the time, fetching dynamic config option's value is directly fetching its value from cache, and it's much faster than `ConcurrentHashMap#get`.
  * the default value of `INCLUDE_DEBUG_HEADER` and `SET_CONTENT_LENGTH` are `false`, it means that the optimized implementation will just test the options' value and skip `ConcurrentHashMap#get`.
* testing `INCLUDE_DEBUG_HEADER` before building the debug string.


spencergibb:
Please explain what you are doing. Optimize is vague. Tests are usually a good practice as well.

PR #1626, opened 2017-01-17
Remove check that does not apply Formatters to Collections

### LINKED ISSUE #1526: No way to format List<java.util.Date> in Feign client
Are there any way how to make custom date format now? I saw a bit similar issue https://github.com/spring-cloud/spring-cloud-netflix/issues/1178, but I think that it's rather a bug than an enhancement.

I tried to use the list of date and neither suggested approach is working for me. I have such a client
```
@FeignClient(name = "exampleClient", url = "#{EXAMPLE_URL}" )
public interface ExampleClient{
    @RequestMapping(value = "/example", method = RequestMethod.GET)
    String getExampleValue(@RequestParam(name = "date") List<Date> dates);
}
```
And I couldn't find any way how to format the date. I think it's partially because of the https://github.com/spring-cloud/spring-cloud-netflix/issues/1115 fix. The extender is not appends, and no formatter is used,

ryanjbaxter:
Currently there is a check which does not apply any `Formatter`s to `Collection`s passed in Feign Client parameters.  This problem was addressed in https://github.com/OpenFeign/feign/commit/100a9acc64dbf86760b2c55196a9bdb608ed00c0#diff-51b176e568a2cdf4a0174d10f0979da0.  We can now remove the check.

Fixes #1526

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java
@@ -241,19 +241,11 @@ public class SpringMvcContract extends Contract.BaseContract
 			}
 		}
 		if (isHttpAnnotation && data.indexToExpander().get(paramIndex) == null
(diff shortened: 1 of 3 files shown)


dsyer:
LGTM

PR #1640, opened 2017-01-24
Make 1.2.x compatible with boot 1.3 thru 1.5

spencergibb:
This works by testing the classpath for the right {Servlet|Filter}RegistrationBean. In 1.4 the old location and new are available. The old location is removed in 1.5. This change uses the old location for boot 1.3-1.4 and the new only in 1.5

There are spring boot 1.3 regression tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,6 +142,7 @@
 		<module>spring-cloud-starter-turbine-amqp</module>
 		<module>spring-cloud-starter-turbine-stream</module>
 		<module>spring-cloud-starter-zuul</module>
+		<module>spring-cloud-netflix-integration-tests</module>
 		<module>docs</module>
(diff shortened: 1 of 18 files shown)


dsyer:
I know this is a small change, but I really don't understand yet why we need it. Why would anyone want to write an app with Spring Boot 1.3 and Spring Cloud Camden?

marcingrzejszczak:
I agree with Dave but there might be some writing a Brixton app that want to bump to Camden. Otoh the migration would be a small change to do and we wouldn't need to support such hacks (I'd need to do a similar one in Sleuth)

spencergibb:
As of Camden.SR5, it requires boot 1.4 or 1.5

Issue #104, opened 2014-12-13
@EnableZuulServer showing deprecated.


ccit-spence:
I notice that @EnableZuulServer is showing that is deprecated.  What is the intended replacement? Should it be showing deprecated?

PR #1654, opened 2017-01-29
Add new Zuul filters endpoint and refactor routes endpoint

darylrobbins:
Added a new filters endpoint to list the current Zuul filters that are in effect. It returns a map keyed by filter type, containing lists of the corresponding filters.

- Create a new endpoints package to house endpoints and moved routes endpoint
- Added a test for the filters endpoint and backfilled tests for the routes endpoint
- Updated documentation to describe new filters endpoint

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/endpoints/FiltersEndpoint.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/endpoints/FiltersEndpoint.java
@@ -1,0 +1,67 @@
+package org.springframework.cloud.netflix.zuul.endpoints;
+
+import com.netflix.zuul.ZuulFilter;
+import com.netflix.zuul.filters.FilterRegistry;
+import org.springframework.boot.actuate.endpoint.Endpoint;
+import org.springframework.boot.actuate.endpoint.mvc.MvcEndpoint;
+import org.springframework.jmx.export.annotation.ManagedAttribute;
+import org.springframework.jmx.export.annotation.ManagedResource;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+import java.util.*;
+
+/**
+ * Endpoint for listing Zuul filters
+ */
+@ManagedResource(description = "List Zuul filters")
+public class FiltersEndpoint implements MvcEndpoint {
+
+    @RequestMapping(method = RequestMethod.GET)
+    @ResponseBody
+    @ManagedAttribute
+    public Map<String, List<Map<String, Object>>> getFilters() {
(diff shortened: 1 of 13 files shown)


pivotal-issuemaster:
@darylrobbins Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1654)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1654) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@darylrobbins Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1654)!

darylrobbins:
Not quite sure what is causing the test failures, but I don't think it's my changes. I saw those same failures when I first ran a build against master (before making any changes). I re-ran the same build without making any changes and it worked.

The test run works successfully on my machine for my changes. Looks like it could be a timing/state propagation issue with those tests?

tmack8001:
Yeah there have been some flaky tests on master recently I've run into as well.

darylrobbins:
As an aside, any particular reason the routes endpoint wasn't added to ZuulServer as well as ZuulProxy? You will always have routes and filters regardless of whether you add the extra Proxy glue.

tmack8001:
@EnableZuulProxy is used when you want to have a zuul setup with the reverse proxy filters
@EnableZuulServer is used to setup base vanilla zuul without any of the proxying filters that were added to ZuulProxy.

Details here https://github.com/spring-cloud/spring-cloud-netflix/issues/104

darylrobbins:
Understood, but why wouldn't I want to have both these endpoints in either case? I was more questioning why the routes endpoint was only enabled in the proxy case.

spencergibb:
Probably an oversight

spencergibb:
Rerunning build

ryanjbaxter:
@darylrobbins looks like some tests are failing, can you look into them?

darylrobbins:
Not sure this is related to my changes, but I'll try to take a look when I have a chance.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/endpoints/FiltersEndpoint.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.util.*;
+
+/**
+ * Endpoint for listing Zuul filters
+ */
+@ManagedResource(description = "List Zuul filters")
+public class FiltersEndpoint implements MvcEndpoint {
+
+    @RequestMapping(method = RequestMethod.GET)
```
spencergibb: We use tabs instead of spaces.

spencergibb:
ping @darylrobbins this will need a rebase.

spencergibb:
Superceded by #2244

Issue #1663, opened 2017-02-02
Any multipart request with a boundary in the content type raises an exception


eloipoch:
In 1.3.0.M1 seems that a bug has been introduced.

Any multipart request that in his `Content-Type` adds the `boundary` parameter (in theory a required parameter, see [RFC](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)) raises an exception with the message "MIME type may not contain reserved characters".

I think it has been introduced in the [commit](https://github.com/spring-cloud/spring-cloud-netflix/commit/d76bd84466616f53e5eee6879fee9a823d767841)  which tries to handle well a content-type with a charset due to the upgrade of the `httpclient` library.

I will try to open a PR with a fix in order to have it for this release.

hartmut-pq:
Hi, I've got same issues with 1.2.5.RELEASE

spencergibb:
Reopening to apply to Camden branch 1.2.x

hartmut-pq:
Thanks a lot!
Note: broke upgrade to spring boot `1.5.1.RELEASE` & `org.springframework.cloud:spring-cloud-dependencies:Camden.SR5`

spencergibb:
We're a workaround with Camden.SR5 and Spring Boot 1.5 https://gist.github.com/spencergibb/80a134b4738c78ca64e4d0c77214fcb6

spencergibb:
Cherry picked to 1.2.x ee953bef07034f88e352a7ebf938cbd345e2535e

prantan7:
We tried after upgrading the boot dependencies to 1.5.1.RELEASE & spring-cloud-dependencies to Camden.SR5, it still didn't work for us.

With what versions the fix for  "-MIME type may not contain reserved characters" is supposed to work

ryanjbaxter:
It looks like this was cherry picked after the release occurred, can your try `Camden.BUILD-SNAPSHOT`?

prantan7:
Thanks Ryan, I will check with the SNAPSHOT version and let you know.

Is there any other to resolve this besides updating the dependency. I saw some messages regarding using the RibbonRoutingFilter instead of the SimpleHostRoutingFilter.

spencergibb:
Yes, it will be in Dalston RC1. The gist above is a workaround.

hartmut-pq:
So there is no other Camden Release lined up including the fix? 
[...]

PR #1664, opened 2017-02-02
Allow multipart request with boundary attribute

eloipoch:
This PR fixes the issue #1663


spencergibb:
Thanks! Didn't know about `parse()`.

Issue #1414, opened 2016-10-20
Support feign SetterFactory


LucasHCruz:
Hi guys.. by default `FeignClient` uses `Hystrix`

How can I put a commandKey and commandGroup in the feign client?

Ps.: If I put the HystrixCommand in previous layer, will Hystrix create two HystrixCommand because of default integration with FeignClient?

Example:

Client:

```
@FeignClient(url = "${myurl}", name = "MyClient")//How to put hystrix name and group?
public interface MyClient {
    @RequestMapping(method = RequestMethod.POST, value = "/Do")
    Response do(ObjDTO objDTO);
}
```

Previous layer:

```
@HystrixCommand(groupKey = HystrixHelper.MY_GROUP, commandKey = "MyKey") // Will FeignClient create another  HystrixCommand?
public Response cancel(ObjDTO objDTO) {
    Response response = myClient.do(objDTO);
    return response;
}
```

spencergibb:
Feign supports it, but we haven't added support for it yet: https://github.com/OpenFeign/feign/tree/master/hystrix

LucasHCruz:
Ok... but will FeignClient use the same HystrixCommand of previous layer or will create another one?

spencergibb:
It will create another one

ErikKringen:
As a work around, since it is supported to make your own `Bean` of `Feign.Builder`, I created my own that uses a `SetterFactory`. This works with Feign 9.2+

```
@Bean
@Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
@ConditionalOnProperty(name = "feign.hystrix.enabled", matchIfMissing = true)
public Feign.Builder feignBuilder() {
  SetterFactory commandKeyIsRequestLine = (target, method) -> {
    String groupKey = target.name();
    String commandKey = method.getAnnotation(RequestLine.class).value();
    return HystrixCommand.Setter
      .withGroupKey(HystrixCommandGroupKey.Factory.asKey(groupKey))
      .andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
  };
  return HystrixFeign.builder().setterFactory(commandKeyIsRequestLine);
}
```

PR #1673, opened 2017-02-07
Add support for Feign.SetterFactory.

ErikKringen:
Fixes gh-1414

### DIFF
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
@@ -25,6 +25,7 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 
 import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.text.ParseException;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
```diff
@@ -525,12 +564,16 @@ public void testParams() {
...
-	public void testHystrixCommand() {
+	public void testHystrixCommand() throws NoSuchMethodException {
 		HystrixCommand<List<Hello>> command = this.testClient.getHellosHystrix();
 		assertNotNull("command was null", command);
 		assertEquals(
 				"Hystrix command group name should match the name of the feign client",
 				"localapp", command.getCommandGroup().name());
+		String configKey = Feign.configKey(TestClient.class,
+			TestClient.class.getMethod("getHellosHystrix", (Class<?>[]) null));
+		assertEquals("Hystrix command key name should match the feing config key",
```
spencergibb: typo: feing

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
```diff
@@ -39,12 +39,15 @@
 import feign.codec.Decoder;
 import feign.codec.Encoder;
 import feign.codec.ErrorDecoder;
+import feign.hystrix.HystrixFeign;
```
spencergibb: This concerns me a little bit. Previously, this class only dealt with feign core classes. Now this introduces a dependency on the feign-hystrix module.

ErikKringen: Another option I was looking at would be to put it in `HystrixTargeter`, so the check to see if it's a HystrixBuilder is already done

spencergibb: Would you be willing to give that a try?

ErikKringen: I'll give it a go!

ErikKringen:
@spencergibb Implemented the SetterFactory within HystrixTargeter as discussed

Issue #1672, opened 2017-02-06
SpringMvcContract does not support @RequestParam without a value when parameter type is Set


fredboutin:
We are using SpringMvcContract to generate our feign clients and since we upgraded from 1.1.x to 1.2.x we can no longer have a Set annotated @RequestParam without value set to the parameter name. A PR was made to support Maps in a GET using query params and since this PR it is broken. There is the PR : https://github.com/spring-cloud/spring-cloud-netflix/commit/d356e571340c26768e706633d082c20d9a554b64 

Removing this line fixes it :
https://github.com/spring-cloud/spring-cloud-netflix/blob/1.2.x/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringMvcContract.java#L335

I don't really know why this line is there in the first place and if I'm correct, I think it breaks other types than Sets. Does it serve a particular use case?

Thanks!

Edit :
Here is the stack : 
```
Caused by: java.lang.IllegalStateException: QueryMap parameter must be a Map: interface java.util.Set
    at feign.Util.checkState(Util.java:128)
    at feign.Contract$BaseContract.parseAndValidateMetadata(Contract.java:126)
    at org.springframework.cloud.netflix.feign.support.SpringMvcContract.parseAndValidateMetadata(SpringMvcContract.java:133)
```
And here is an example of a resource call that throws this when creating the client : 

```
@RequestMapping(method = RequestMethod.GET, value = "/test/{pathVariable}")
FooBar getTestFooBar(@PathVariable String pathVariable,
                     @RequestParam(required = false) Set<Foo> problematicSet);
```

ryanjbaxter:
@lowzj ^^^

lowzj:
@fredboutin Yes, it was caused by the [PR-1102](https://github.com/spring-cloud/spring-cloud-netflix/pull/1102).
In the `@RequestParam` docs:
> CASE 1. If the method parameter is Map<String, String> or MultiValueMap<String, String> and a parameter name is not specified, then the map parameter is populated with all request parameter names and values.
> CASE 2. If the method parameter type is Map and a request parameter name is specified, then the request parameter value is converted to a Map assuming an appropriate conversion strategy is available.

And this line:
```
BeanUtils.isSimpleProperty(parameterTypes[parameterIndex].getClass())
```
is to support `Map` arguments annotated with `@RequestParam` where the annotation does not specify a request parameter name(_case 1_). 
[...]

PR #1678, opened 2017-02-09
fix gh-1672, support @RequestParam without a value

lowzj:
see #1672


lowzj:
I think it's ok that the coverage was decreased due to the offset of the removed lines.

PR #1680, opened 2017-02-09
Add Retry Logic To Zuul Using Spring Retry

ryanjbaxter:
This PR adds retry logic back to Zuul when using the Apache HTTP Client or the Ok HTTP Client.  This is the last piece to fixing #1295.

PR #1689, opened 2017-02-13
Makes Spring Retry optional for Zuul

ryanjbaxter:
PR #1680 added the ability to retry failed requests when using the Apache or OK HTTP Clients.  However it did not assume Spring Retry would not be on the classpath.  This PR makes it so Spring Retry can be an optional depedency.


ryanjbaxter:
Yeah that was the best way I could think of that would not involve duplicating a ton of code.  There was another approach I was thinking about but that would have involved adding 2 more command handlers and command factories.  Does that mean it gets your blessing? ;)

PR #1695, opened 2017-02-14
Adds a check for an empty URI in updateToHttpsIfNeeded

### LINKED ISSUE #1690: RibbonUtils.updateToHttpsIfNeeded Should Deal With URI Equal To Empty String
I have the following configuration in my Zuul server.

```
server:
  port: 8080

hystrix:
  command:
    default:
      execution:
        timeout:
          enabled: false

zuul:
  routes:
    greeting:
      path: /greeting/**
      serviceId: greeting

ribbon:
  eureka:
    enabled: false

greeting:
  ribbon:
    listOfServers: localhost:8443
```

When I make a request to `http://localhost:8080/greeting` `RibbonLoadBalancingHttpClient.reconstructURIWithServer` will be called.  Since the `greeting` server has a port that ends in `443` we will try to set the scheme to `https` in `RibbonUtils.updateToHttpsIfNeeded`.  However the URI from the request is the empty string so when calling `uriComponentsBuilder.build(true).toUri()` the URI in `uriComponentsBuilder` is just `https` and this fails because it is not a valid URI.  I think there should be a check in `RibbonUtils.updateToHttpsIfNeeded` to make sure the URI is not the empty string.  @spencergibb what do you think?

ryanjbaxter:
Fixes #1690

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonUtils.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonUtils.java
@@ -71,7 +71,7 @@ public class RibbonUtils {
 	public static URI updateToHttpsIfNeeded(URI uri, IClientConfig config, ServerIntrospector serverIntrospector,
 			Server server) {
 		String scheme = uri.getScheme();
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Should we also port this to the 1.2.x branch?

spencergibb:
@ryanjbaxter yes.

Issue #1698, opened 2017-02-15
Fix encoding issue in SpringEncoder


tjuchniewicz:
Fix  "// TODO: set charset" in [SpringEncoder](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/support/SpringEncoder.java#L99)

Current code sends invalid request when Feign ApacheHttpClient is used. When byte array is encoded it is encoded as `StringEntity` not `ByteArrayEntity`. See [ApacheHttpClient](https://github.com/OpenFeign/feign/blob/master/httpclient/src/main/java/feign/httpclient/ApacheHttpClient.java#L139)

My dummy fix is:

```java
// do not use charset for binary data!
if (requestContentType.isCompatibleWith(MediaType.APPLICATION_OCTET_STREAM)) {
   request.body(outputMessage.getOutputStream().toByteArray(), null);
} else {
   request.body(outputMessage.getOutputStream().toByteArray(), Charset.forName("UTF-8"));
}
```

spencergibb:
@tjuchniewicz are you willing to submit a PR?

tjuchniewicz:
@spencergibb Yes. I will create PR.

spencergibb:
Thanks @tjuchniewicz!

PR #1699, opened 2017-02-15
Fix encoding issue in SpringEncoder for binary data

tjuchniewicz:
Fixes gh-1698


tjuchniewicz:
I will try. Give me more time.

spencergibb:
@tjuchniewicz NP!

tjuchniewicz:
@spencergibb Test added. Some test fail on CI but I believe this is not my fault. All tests pass on my machine.

Issue #1724, opened 2017-02-23
prefer-ip-address not works in statusPageUrlPath when management port != server port


windinn:
I set prefer-ip-address to true in a service app using eureka client. The EurekaInstanceConfigBean.statusPageUrlPath is set as http://172.18.6.27:9999/info when management port == server port, but set as http://78220986847a***:9999/info when use a different management port.

**application.yml**
```yaml
eureka. instance.prefer-ip-address: true
```

**pom.xml**:
```xml
<dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.springframework.cloud</groupId>
                <artifactId>spring-cloud-dependencies</artifactId>
                <version>Camden.SR5</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
    <dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-starter-eureka</artifactId>
        </dependency>
    </dependencies>
```

windinn:
I debug the code, it should be a bug in EurekaClientAutoConfiguration.eurekaInstanceConfigBean, which doesn't take the prefer-ip-address property into account.

PR #1725, opened 2017-02-23
fix issue: prefer-ip-address not works in statusPageUrlPath when management port != server port

windinn:
take the prefer-ip-address property into account in EurekaClientAutoConfiguration.eurekaInstanceConfigBean

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -88,12 +88,6 @@ public class EurekaClientAutoConfiguration {
 	@Value("${management.port:${MANAGEMENT_PORT:${server.port:${SERVER_PORT:${PORT:8080}}}}}")
 	private int managementPort;
 
-	@Value("${eureka.instance.hostname:${EUREKA_INSTANCE_HOSTNAME:}}")
-	private String hostname;
-
-	@Value("${eureka.instance.prefer-ip-address:${eureka.instance.preferIpAddress:false}}")
-	boolean preferIpAddress;
-
 	@Autowired
 	private ConfigurableEnvironment env;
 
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
Is there an issue open for this bug?  Describing the bug in the PR title is not very useful.  Also you will need to add some tests for this.

ryanjbaxter:
Oh I see the issue is #1724

windinn:
@ryanjbaxter thanks:)

ryanjbaxter:
Can you please provide some unit tests?

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
```diff
@@ -120,6 +123,7 @@ public EurekaInstanceConfigBean eurekaInstanceConfigBean(InetUtils inetUtils) {
 		EurekaInstanceConfigBean instance = new EurekaInstanceConfigBean(inetUtils);
 		instance.setNonSecurePort(this.nonSecurePort);
 		instance.setInstanceId(getDefaultInstanceId(this.env));
+		instance.setPreferIpAddress(preferIpAddress);
```
spencergibb: This should get automatically set since it is simply a `@ConfigurationProperties` bean. The other items set here have business logic performed.

windinn: Yes, the field will get automatically set by spring **after** `eurekaInstanceConfigBean()` method returns. However, it's too late, `eurekaInstanceConfigBean()` will use the uninitialized `preferIpAddress` to set `statusPageUrlPath`.

spencergibb: got it.

windinn:
@ryanjbaxter Unit test added.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
```diff
@@ -91,6 +91,9 @@
 	@Value("${eureka.instance.hostname:${EUREKA_INSTANCE_HOSTNAME:}}")
 	private String hostname;
 
+	@Value("${eureka.instance.prefer-ip-address:${eureka.instance.preferIpAddress:false}}")
```
spencergibb: This should be moved to use a `RelaxedPropertyReslover` in the `@Bean` method itself. The resolver from https://github.com/windinn/spring-cloud-netflix/blob/4f801d8653ca8b178a4b8cb1740736a4b0dfc435/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java#L132-L134 should be moved to the beginning of the method. The `hostname` field should be moved also. Without it, you wouldn't be able to use an ENV var like `EUREKA_INSTANCE_PREFER_IP_ADDRESS`.

windinn: done

Issue #1334, opened 2016-09-12
Eagerly create RibbonClients


KamikaZeeFu:
From what I've understood, each Ribbon Client created by Spring basically gets it's own ApplicationContext.
This context is created when the client is first used, which makes the very first request rather slow.
Is it possible to eagerly create these clients/contexts?

spencergibb:
@KamikaZeeFu not currently. In fact, we've had problems when feign clients (similar to ribbon clients) get eagerly created. Besides the first request is slow, are you experiencing problems?

jansyk13:
@spencergibb Problem is not the lazy initialization, but the fact that the creation of client is not synchronized. ```SpringClientFactory#instantiateWithConfig```

spencergibb:
And what problems does that cause @jansyk13?

KamikaZeeFu:
@spencergibb well the main problem is that it creates unpredictable delays.
I have a setup with a Zuul reverse proxy that calls a couple of backend services.
When i deploy a new version of a backend, the first request will be delayed, which causes random request to run into the zuul configured timeout. My default response time is around 300-400ms, these initial context init spikes push it to above 1s.

Of course I could just increase my timeouts, but that is rather unsatisfying.

spencergibb:
It would likely be feasible to eagerly create the zuul ribbon clients since we know all of the routes ahead of time.

jebeaudet:
In the meantime, you could use something like this to create the clients when the proxy boots, we use that to refresh them if we ever change the read/connect timeout values in archaius : 

```
@Component
public class RibbonEarlyInstantiator implements ApplicationListener<ApplicationStartedEvent>{

    @Autowired
    private ZuulProperties zuulProperties;
    @Autowired
    private SpringClientFactory springClientFactory;

    @Override
    public void onApplicationEvent(ApplicationStartedEvent event)
    {
        Map<String, ZuulRoute> routes = zuulProperties.getRoutes();
        routes.values()
              .stream()
              .filter(route -> route.getServiceId() != null)
              .map(route -> route.getServiceId())
              .distinct()
              .forEach(serviceId -> {
                  logger.info("Instantiating the context for the client '{}'", serviceId);
                  springClientFactory.getClient(serviceId, RibbonLoadBalancingHttpClient.class); //Or RestClient.class if you still use it
              });
    }
}

bijukunjummen:
[...]

PR #1744, opened 2017-03-02
GH-1334: Support for eagerly creating Ribbon Clients

bijukunjummen:
#1334  - Eagerly Creates Ribbon Clients

* Added a Spring Boot `ApplicationReadyEvent` listener, which calls `springClientFactory.createAndCacheContexts()` introduced as part of the PR here - https://github.com/spring-cloud/spring-cloud-commons/pull/182.


bijukunjummen:
Will send another PR, the approach in this PR does not look right - configurations in `NamedContextFactory` may not even be present at the point of eagerly caching the context.

PR #1762, opened 2017-03-07
Adds conditional on Ribbon class to RibbonAutoConfiguration

### LINKED ISSUE #1760: RibbonAutoConfiguration missing conditional on class Ribbon.class
```
@ConditionalOnClass({ IClient.class, RestTemplate.class })
```
The `IClient.class` is part of ribbon core package and `Ribbon.class` is part of ribbon module.
My project has dependency on Ribbon-LoadBalancer which brings ribbon core not ribbon and my code fails in run time with no class def found on Ribbon.class

The fix should be:
```
@ConditionalOnClass({ Ribbon.class, RestTemplate.class })
```

ryanjbaxter:
Fixes #1760


ryanjbaxter:
@spencergibb that is the only way i could get the regex value to work (by using a variable), not sure why though...

Issue #327, opened 2015-04-30
Zuul/Hystrix: HTTP connection leak in case of timeout


brenuart:
We came into scenarios where Zuul is leaking http connections (i.e. not returned to the pool) after execution. After a while, the pool's maximum limit is hit and Zuul can't provide any service anymore. 

**In short, it turns out that Zuul leaks HTTP connections if a service takes longer than the Hystrix execution timeout but before the ReadTimeout occurs.**

We could reproduce the case using a Zuul instance left to its default settings except for the http pool size which is set to 1 to rapidly reproduce the problem:
- Hystrix timeout = 1000ms
- ribbon.ReadTimeout = 5000ms (default value)
- ribbon.MaxTotalHttpConnections = 1
- ribbon.MaxHttpConnectionsPerHost = 1
- either use Eureka to discover the service or register it statically as explained in the documentation

Our test service exposes a simple REST API with a single argument to tell it how much time it should wait before returning a result. The service is proxied by Zuul such that a request for `http://zuul/service/hello?delay=500` is forwarded to the actual service instance. The `delay`parameter tells the service to wait 500ms before returning. Fairly simple.

### Scenario 1
Invoke the service with a delay of *500ms*. This delay (simulated execution time) is lower than all Zuul timeout (Hystrix and Ribbon ReadTimeout). 
Everything is working fine.

### Scenario 2
Invoke the service with a delay of *1500ms*. 
This should trigger the Hystrix timeout but is still within the ReadTimeout limit. A *timeout* error is returned to the client, but only after the service returns (because of the SEMAPHORE isolation), i.e. after 1500ms.
In this case, the HTTP connection is **NEVER RETURNED TO THE POOL**. 

From the logs:
```
17:17:13.283 DEBUG --- [RibbonCommand-2] com.netflix.niws.client.http.RestClient  : RestClient sending new Request(GET: ) http://localhost:8085/hello
17:17:13.283 DEBUG --- [RibbonCommand-2] c.n.http4.MonitoredConnectionManager     : Get connection: {}->http://localhost:8085, timeout = 2000
17:17:13.283 DEBUG --- [RibbonCommand-2] com.netflix.http4.NamedConnectionPool    : [{}->http://localhost:8085] total kept alive: 1, total issued: 0, total allocated: 1 out of 1
[...]

PR #1793, opened 2017-03-18
Stop retry after hystrix is timed out

### LINKED ISSUE #1772: Zuul won't stop retry after hystrix is timed out
I'm on spring cloud dependencies Camden.SR6, and a setup of zuul with the following config

```
ribbon:
    MaxAutoRetriesNextServer: 0
    MaxAutoRetries: 2
    ReadTimeout: 3000
    OkToRetryOnAllOperations: true #false to only allow get to retry

hystrix:
  command:
    sc-test-node-0:
      execution:
        isolation:
          strategy: THREAD
          thread:
            timeoutInMilliseconds: 5000
```

and I put a simple resource behind zuul which sleeps 4000ms when invoked.
```java
    @GET
    @Path("getthensleep/{sleepms}")
    @Produces(MediaType.APPLICATION_JSON)
    public Response getThenSleep(@PathParam("sleepms") int sleepms){
        LOGGER.info("get then sleep {} ms", sleepms);
        try {
            Thread.sleep(sleepms);
        } catch (InterruptedException e) {
            LOGGER.error("Sleep error", e);
        }
        return Response.ok().build();
    }
```

and I use postman to call the getthensleep resource to let it sleep 4000ms, I expect zuul's RibbonRoutingFilter times out after 3000ms and retry until hystrix times out, 2 separate call should be made from zuul. However, after hystrix times out, zuul returns with code 500, another call is made to getthensleep resource again. In total 3 calls are made by zuul. I think the zuul should stop trying when hystrix times out, and It shouldn't make other attempt after hystrix times out.

rterentiev:
Fixes #1772.

Connection leak issue https://github.com/spring-cloud/spring-cloud-netflix/issues/327 is fixed by closing response, if subscriber (hystrix command) unsubscribes (hystrix timeout error) during request execution. See ```AbstractLoadBalancingClient.getServerOperation```.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
@@ -17,107 +17,99 @@ package org.springframework.cloud.netflix.ribbon.apache;
 
 import java.io.IOException;
 import java.net.URI;
-import org.apache.commons.lang.BooleanUtils;
-import org.apache.http.HttpResponse;
-import org.apache.http.client.config.RequestConfig;
-import org.apache.http.client.methods.HttpUriRequest;
+import java.net.URISyntaxException;
+
 import org.springframework.cloud.client.ServiceInstance;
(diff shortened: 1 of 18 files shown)


spencergibb:
I'm afraid these changes significantly change the way things work and I'm of the mind to not accept them.

PR #1794, opened 2017-03-19
Make EurekaDiscoveryClient.EurekaServiceInstance constructor public

### LINKED ISSUE #1792: Make EurekaServiceInstance constructor public
We are building a [tool](http://cereebro.io) that can be used as an Eureka server add-on to document service instances. It is also used on the (discovery) client side, so we manipulate the generic `ServiceInstance` interface.

On the Eureka server side, we need to create `ServiceInstance` objects based on `com.netflix.appinfo.InstanceInfo`.  
The public class `org.springframework.cloud.netflix.eureka.EurekaDiscoveryClient.EurekaServiceInstance` is exactly what we need, but its constructor is package private.

Any objection on making the constructor public ?

michaeltecourt:
Closes #1792

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClient.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClient.java
@@ -104,7 +104,7 @@ public class EurekaDiscoveryClient implements DiscoveryClient {
 		private InstanceInfo instance;
 
 		public EurekaServiceInstance(InstanceInfo instance) {
-                        Assert.notNull(instance, "Service instance required");
+			Assert.notNull(instance, "Service instance required");
 			this.instance = instance;
 		}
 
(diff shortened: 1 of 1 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClient.java
```diff
@@ -102,7 +103,8 @@ public URI getUri() {
 	public static class EurekaServiceInstance implements ServiceInstance {
 		private InstanceInfo instance;
 
-		EurekaServiceInstance(InstanceInfo instance) {
+		public EurekaServiceInstance(InstanceInfo instance) {
+                        Assert.notNull(instance, "Service instance required");
```
spencergibb: We use tabs not spaces

michaeltecourt:
Sorry, shouldn't have used vim :sweat_smile:

michaeltecourt:
@spencergibb fixed ✅

PR #1836, opened 2017-04-04
Add the ability for ribbon clients to specify response status codes they would like to retry

### LINKED ISSUE #1540: Retry on specific http response codes
Hi
 I am trying Ribbon and Hystrix with following configuration property  based on the solution suggested here . 

http://stackoverflow.com/questions/37272996/zuul-ribbon-hystrix-not-retrying-on-different-instance?noredirect=1&lq=1

```
<ServiceId>.ribbon.listOfServers= https://approute1.g1.cloud.mycompany.net/api/v1/status,https://approute2..g2..cloud.mycompany.net/api/v1/status 
<ServiceId>.ribbon.ReadTimeout=800
<ServiceId>.ribbon.ConnectTimeout=250
<ServiceId>.ribbon.OkToRetryOnAllOperations=true
<ServiceId>.ribbon.MaxAutoRetriesNextServer=2
<ServiceId>.ribbon.MaxAutoRetries=0
hystrix.command.<ServiceId>.execution.isolation.strategy=THREAD
hystrix.command.<ServiceId>.execution.isolation.thread.timeoutInMilliseconds=100000

```
List of servers are in  Cloud foundry and in Cloud foundry returns  "404 Not Found: Requested route" when app is down. 
with the Default implementation 404 HTTP response code  does not trigger fallback within ribbon to try the next server in the list . I would like to Customize  Ribbons Response handler to  trigger fallback to the next server in the list on 404 . How can we achieve this !? 

Retry is working fine when one of the server in the list is not reachable (Connection Time out).

I would like to achieve the same in Zuul also .  Once the above question is answered I guess I can apply the same for zuul as well .

ryanjbaxter:
Fixes #1540.

Build will fail until spring-cloud/spring-cloud-commons#197 is merged

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -2448,6 +2448,22 @@ certain Ribbon properties.  The properties you can use are
 `client.ribbon.OkToRetryOnAllOperations`. See the https://github.com/Netflix/ribbon/wiki/Getting-Started#the-properties-file-sample-clientproperties[Ribbon documentation]
 for a description of what there properties do.
 
+In addition you may want to retry requests when certain status codes are returned in the
+response.  You can list the response codes you would like the Ribbon client to retry using the
+ property `clientName.ribbon.retryableStatusCodes`.  For example
+
+[source,yaml]
+----
+clientName:
+  ribbon:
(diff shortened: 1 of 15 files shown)


ryanjbaxter:
Meeds to be merged to 1.2.x as well

Issue #1838, opened 2017-04-05
Passing decode404 configuration globally using Configuration


tanmaybinaykiya:
Is there a way to pass the decode404 property to all FeignClients by instantiating a bean through a configuration class, like it is possible for ErrorDecoder or Contract etc?

spencergibb:
Not currently.

PR #1858, opened 2017-04-13
Enables Feign.Builder customization

venilnoronha:
This PR adds a new type named `FeignBuilderCustomizer` using which `Feign.Builder`s can be customized. For example, a customizer bean can be defined to enable `decode404` on all `Feign` instances created via `@FeignClient`. Refer `FeignClientTests#testFeignBuilderCustomizerTest404()` for a demonstration. Note that the test configures the customizer at the `@FeignClient` level, while it's possible to have it configured at the `@EnableFeignClients` level as well. See #1838 for additional details.

Please review and merge.

Thanks,
Venil

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignBuilderCustomizer.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignBuilderCustomizer.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


spencergibb:
I'd rather not introduce another class if possible. It is possible to have `decode404` respond to properties.

venilnoronha:
I was trying to make the solution more generic and not just for `decode404`. If we don't want that, then could you tell me which properties class you're talking about?

spencergibb:
There isn't a properties class. At this point, I'm looking for simplicity rather than increased complexity.

venilnoronha:
Do you suggest that I create a new `@ConfigurationProperties` class?

spencergibb:
I don't know if `FeignClientsRegistrar` is created before `@ConfigurationProperties` or not, so. Not sure. You can certainly use the `environment` there. I'd also add resolve placeholders to the `decode404`.

venilnoronha:
Okay, I'll have a look. Thanks!

spencergibb:
Is this superseded by #1942?

spencergibb:

Issue #1854, opened 2017-04-12
custom route property


hodeh:
Is there a way to have custom property in a Route definition at configuration and then I can use it later in a filter.
----Config File-----
```
    route-x:
      path: /v1.0/route/x/**
      service-id: route-x
      strip-prefix: false
      custom-property-x: value-x
```

ryanjbaxter:
I dont believe so

spencergibb:
No, what is the property for and how is it to be used?

hodeh:
@spencergibb the property name will be authenticationRequired (true/false).
Custom Authentication filter will read that value and  it will  authenticate the request if that route has this custom value. if the value is false then filter knows that this is a public resource and there shouldnt be any authentication

PR #1860, opened 2017-04-13
Route Secure Access Level

KevinDaHub:
This new configuration will add the possibility to authenticate your route downstream on zuul level, where the possible values are private, partial-exposed, partial-private and public
The default value is public so nothing will break and it is an optional choice

```
zuul:
  routes:
    stores:
      path: /stores/**
      secure-access-level:
    machine:
      path: /machine/**
      secure-access-level: partial-exposed
    book:
      path: /book/**
      secure-access-level: partial-private
    reservation:
      path: /reservation/**
      retryable: true
      service-id: msrental
      secure-access-level: private
```
Fixes gh-1854

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -1604,6 +1604,102 @@ NOTE: the routes should respond automatically to changes in the
 service catalog, but the POST to /routes is a way to force the change
 to happen immediately.
 
+=== The Routes Security Configuration
+To increase our performance, we implement a custom configuration security mechanism for making less network hops,
+a mechanism implemented on zuul where you can configure which route needs private, public or partial authentication. +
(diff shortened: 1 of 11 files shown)


spencergibb:
Two things:

1. Its a specific application of #1854, so I don't think it fixes it.
2. I think something like this fits better in https://github.com/spring-cloud/spring-cloud-security

KevinDaHub:
The reason I got it into Zuul is that the authentication mechanism is being implemented by Zuul filters.

spencergibb:
There is already precedent for doing zuul security in that project https://github.com/spring-cloud/spring-cloud-security/tree/master/spring-cloud-security/src/main/java/org/springframework/cloud/security/oauth2/proxy

KevinDaHub:
Ah awesome!
I will get a look at it thanks!

KevinDaHub:
Now I can put the filters in the cloud security repo but I need a extra property in the Zuulproperties class. What do you recommend?

PR #1863, opened 2017-04-14
Support for re-writing Location Headers

bijukunjummen:
#8 - Support for modifying Location headers after downstream calls

 - Modifies Location headers from downstream calls for 3XX status codes with Zuul urls
 - Adds prefix if it is stripped out by zuul downstream calls

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/Route.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/Route.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2017 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


bijukunjummen:
This is a Proof-of-concept PR for #8 , this filter may not always be appropriate so instead of auto-configuring the filter I can add in documentation on how to enable it in case somebody needs this functionality

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/Route.java
```diff
@@ -27,7 +27,7 @@
 public class Route {
 
 	public Route(String id, String path, String location, String prefix,
-			Boolean retryable, Set<String> ignoredHeaders) {
+			Boolean retryable, Set<String> ignoredHeaders, boolean prefixStripped) {
```
spencergibb: Keep the old constructor and delegate to the new one.

spencergibb:
Yeah, I'd say it's opt-in. Here's the other issue. It's all or nothing. All `Location` headers are rewritten regardless. This is one of the things (the all or nothing part) I'm fix in spring-cloud-gateway.

bijukunjummen:
I have removed the auto-configuration now @spencergibb and have added in documentation to show how this filter can be added in if required. The new `spring-cloud-gateway` project looks awesome, will check it out in the next few days.

spencergibb:

tgeens:
@spencergibb Is there anything left we can do to help get this merged ?

PR #1877, opened 2017-04-21
Add @Component for Feign Hystrix Fallback example in doc

yongsungyoon:
Add `@Component` annotation in feign hystrix fallback example in doc. Fallback factory example has `@Component` annotation, but fallback example doesn't. So people may misunderstand that fallback implementation doesn't need to be a bean.


ryanjbaxter:
Its a good point, although it is not necessary to use the `@Component` annotation.  Maybe we should also add some wording that the fallback needs a bean definition.

spencergibb:
I agree with @ryanjbaxter, `@Component` isn't necessary, but it does need to be a bean.

ryanjbaxter:
ping @yongsungyoon

yongsungyoon:
@ryanjbaxter I added a sentence to indicate the need of delaring it as a spring bean.

Issue #665, opened 2015-11-23
Rename starters to include `-netflix-`


spencergibb:
All the starters in the other projects include more context.

`spring-cloud-starter-consul-discovery`, or `spring-cloud-starter-zookeeper-config`.

I would only do this post Brixton, in the C release train.

spencergibb:
Add the new starters in Dalston, remove old starters in Edgware

daniellavoie:
I've started working on this issue. I'm having second though on `spring-cloud-starter-eureka`. When I started on Spring Cloud, I remember getting confused between `spring-cloud-starter-eureka` and `spring-cloud-starter-eureka-server`. 

How about renaming the client to `spring-cloud-starter-netflix-eureka-client` ? I know it's getting long... but we can't escape the`spring-cloud-starter-netflix-eureka-server`.

spencergibb:
@daniellavoie, yeah that makes sense.

daniellavoie:
Adjustment coming through ;) Made a first PR with the old name.

spencergibb:
NP

spencergibb:
@philwebb what does boot do when they rename/deprecate a starter? It would be nice to print a warning that the user is using a deprecated starter.

philwebb:
We update the name and description. See https://github.com/spring-projects/spring-boot/commit/fc00e4cbe0d5a1f9f83e5148df2b36e8b11943f1 for an example.

spencergibb:
Nice, so you do log a warning. Thanks @philwebb

daniellavoie:
Just pushed a new version with the shiny `spring-cloud-starter-netflix-eureka-client`

spencergibb:
@daniellavoie feign, shouldn't be `spring-cloud-starter-netflix-feign` since it's not netflix. Either keep it `spring-cloud-starter-feign` or `spring-cloud-starter-openfeign`. What do you think @adriancole.

daniellavoie:
@spencergibb I understand you point but given this logic, feign should not be part of the spring-cloud-netflix project anymore then. Still, I'll rename it once @adriancole provides us his preference.

brenuart:
`spring-cloud-starter-openfeign` looks good to me as it reflect the actual project name. If you google around `feign` alone then you will be directed to the former Netflix stuff...

My 2 cents...

spencergibb:
@daniellavoie yup, it should move out of netflix. Maybe during 2.0.0.

[...]

PR #1886, opened 2017-04-24
Rename starters to include `-netflix-`

daniellavoie:
Fixes gh-665

This is a big one. I've put the new starters inside a dedicated `spring-cloud-starter-netflix` module to avoid having too much of them at the root of `spring-cloud-netflix`. On the next release train, if required, they can be moved to a higher level once the old starters are deleted.

Old starters now import the new  ones containing `-netflix`. We'll only have to delete the old starters when the next release train comes.


spencergibb:

PR #1887, opened 2017-04-25
ServiceMapper should return a ZuulRoute

saiyedzaidi:
ServiceMapper should return a ZullRoute to allow full configuration of the routes created by it. At this time it hides everything else other than the path and location.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/discovery/DiscoveryClientRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/discovery/DiscoveryClientRouteLocator.java
@@ -113,7 +113,6 @@ public class DiscoveryClientRouteLocator extends SimpleRouteLocator
 			for (String serviceId : services) {
 				// Ignore specifically ignored services and those that were manually
 				// configured
-				String key = "/" + mapRouteToService(serviceId) + "/**";
 				if (staticServices.containsKey(serviceId)
 						&& staticServices.get(serviceId).getUrl() == null) {
 					// Explicitly configured with no URL, cannot be ignored
@@ -124,10 +123,11 @@ public class DiscoveryClientRouteLocator extends SimpleRouteLocator
 						staticRoute.setLocation(serviceId);
 					}
 				}
+				final ZuulRoute dynamicRoute = this.serviceRouteMapper.applyRoute(serviceId);
 				if (!PatternMatchUtils.simpleMatch(ignored, serviceId)
-						&& !routesMap.containsKey(key)) {
+						&& !routesMap.containsKey(dynamicRoute.getLocation())) {
 					// Not ignored
-					routesMap.put(key, new ZuulRoute(key, serviceId));
+					routesMap.put(dynamicRoute.getLocation(), dynamicRoute);
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@saiyedzaidi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1887)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1887) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@saiyedzaidi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=1887)!

saiyedzaidi:
@spencergibb Could you please review this PR ? This change is an update on the PR [Add Zuul Proxy regex serviceId to route mapping](https://github.com/spring-cloud/spring-cloud-netflix/pull/699).  I think the PatternServiceRouteMapper  is a good idea, but its interface should be more generic and allow setting all Route properties.

spencergibb:
I'm afraid I need you to explain the need for this in more detail.

saiyedzaidi:
PatternServiceMapper is an implementation of ServiceMapper. This interface allows for creating routes from the serviceId. The PatternServiceMapper specifically was made with the intention of detecting version number from the serviceId having a pattern of domainName-serviceName-vVersionNumber and map them to routes like /versionNumber/domainName/serviceName. This was is version in the master branch at the moment. It works well.
But its is limited by the fact that the ServiceMapper interface is designed just to take a String serviceId and return a String path. Hence using the interface in its current state we are stuck with the default configuration of ZullRoute, that is stripPrefix as true, retryable as false, customHeaders as false and blank.
The interface needs to be free from this specific purpose implementation and allow for a common way embedding the dynamic routes into the class DiscoveryClientRouteLocator.java that is using it now.
Also, the logic about how a serviceId gets converted into a Route should not be leaked outside the ServiceMapper implementation.
This standardisation of returning a complete ZuulRoute from ServiceMapper would eventually allow for declaring of ServiceMapper purely using Route like properties.
The older methods are retained to avoid breaking existing code. They are marked deprecated, hoping we can remove them sometime.

Please let me know if I should annotate the specific lines to show what's changed.

spencergibb:
I'm still trying to wrap my head around this. Seems like a discussion might have been a good idea before a PR.

You want to be able to set default values for `stripPrefix`, `retryable` and `sensitiveHeaders` in `DiscoveryClientRouteLocator` (since that is the only place `ServiceRouteMapper` is used).

Seems like that could be accomplished with some simple configuration properties.

saiyedzaidi:
The PR proposes that the `ZuulRoute` creation from `serviceId` should not be embedded inside `DiscoveryClientRouteLocator ` but totaly customisable using `ServiceRouteMapper `.
The default values for `stripPrefix`, `retryable `and `sensitiveHeaders` from  `DiscoveryClientRouteLocator ` could be the default/fallback values.
Even applying the default values for the configuration could not have been achieved without code change as the `DiscoveryClientRouteLocator ` is creating `ZululRoute` using `ZuulRoute(String path, String location)` ignoring other property values.
Also, I think, the `ServiceRouteMapper `should own the `ZuulRoute `it is creating, fully. This will allow for full control over these properties from the custom implementations supplied by users.
If we do not think it adds value, I am ok closing this PR down.

spencergibb:
If we decide to do this, we'll be targeting the Edgware release this summer, should be merged in the next week or two.

spencergibb:

PR #1910, opened 2017-05-02
Fix Feign Compression Config Order

### LINKED ISSUE #1580: Feign can never compress request/response.
It seems that the feign's compression won't work. My version is Spring Cloud Camden SR3.

Here goes my configuration.

application.properties

```
feign.compression.response.enabled=true
feign.compression.request.enabled=true
feign.compression.request.mime-types=text/xml,application/xml,application/json
feign.compression.request.min-request-size=2048
```

I found the compression won't work. When I debuged the feign service, I  found that the code can never be loaded. like the pic below.

Then I go into the code, in `spring-cloud-netflix-core`

```java
@Configuration
@EnableConfigurationProperties(FeignClientEncodingProperties.class)
@ConditionalOnClass(Feign.class)
@ConditionalOnBean(ApacheHttpClient.class)
@ConditionalOnProperty(value = "feign.compression.request.enabled", matchIfMissing = false)
@AutoConfigureBefore(FeignAutoConfiguration.class)
public class FeignContentGzipEncodingAutoConfiguration {

    @Bean
    public FeignContentGzipEncodingInterceptor feignContentGzipEncodingInterceptor(FeignClientEncodingProperties properties) {
        return new FeignContentGzipEncodingInterceptor(properties);
    }
}
```

In this line:

```
@ConditionalOnBean(ApacheHttpClient.class)
```

this class ApacheHttpClient, is in the jar `feign-httpclient` .

but the  dependency is even not in the classpath. Here goes my POM.XML

```xml
  <dependencies>
    <dependency>
      <artifactId>spring-boot-starter-web</artifactId>
    </dependency>
    <dependency>
      <artifactId>spring-boot-starter-actuator</artifactId>
    </dependency>

    <dependency>
      <artifactId>spring-cloud-starter-eureka</artifactId>
    </dependency>

    <dependency>
      <artifactId>spring-cloud-starter-feign</artifactId>
    </dependency>
  </dependencies>

  <!-- 引入spring cloud的依赖 -->
  <dependencyManagement>
    <dependencies>
      <dependency>
        <artifactId>spring-cloud-dependencies</artifactId>
        <version>Camden.SR3</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

So, I added the denpency mannally.

```xml
    <dependency>
      <groupId>io.github.openfeign</groupId>
      <artifactId>feign-httpclient</artifactId>
    </dependency>
```

BUT even I did this, Feign's compression can never work.

## My Question

* What's wrong?
* Is it a bug?
* feign.compression.request.min-request-size=2048    Does it mean 2048 byte?

ryanjbaxter:
Fixes #1580

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/FeignAcceptGzipEncodingAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/encoding/FeignAcceptGzipEncodingAutoConfiguration.java
@@ -18,7 +18,8 @@ package org.springframework.cloud.netflix.feign.encoding;
 
 import feign.Feign;
 import feign.httpclient.ApacheHttpClient;
-import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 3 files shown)


spencergibb:
Does it work moving from before to after?

ryanjbaxter:
According to @jpuigsegur and @pgonzalezgua, yes

spencergibb:
OK, wonder if there is a way to test. If not, LGTM.

ryanjbaxter:
@spencergibb I wrote up a simplistic test which makes sure the configuration is enabled.

Issue #1598, opened 2017-01-08
SimpleRouteLocator: Change fields from private to protected to support extension


nafiux:
Hi team, I want to implement my own `public Route getMatchingRoute(final String path)` method, and I would like to extend the functionality from **SimpleRouteLocator**, the problem is that the fields are **privated** and I should implement all the code again.

I think that changing **private** to **protected** will do the trick. I don't want to redo all the code since that implies more responsibility to maintain a healthily code.

Please let me know if it's suitable. Thanks.

spencergibb:
Pull requests welcome, though I'd prefer protected getters rather than protected fields.

PR #1927, opened 2017-05-09
Expose some private fields via protected getters to allow easier customization. Fixes gh-1598

TheSentinel454:


TheSentinel454:
Sorry, it looks like I pulled in some unwanted changes.  Closing while I clean this up.

PR #1928, opened 2017-05-09
Expose some private fields via protected getters to allow easier cust…

TheSentinel454:
…omization. Fixes gh-1598

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
@@ -22,7 +22,6 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


ryanjbaxter:
Not sure I see the need for these changes, can you elaborate on the use case?

TheSentinel454:
Basically, if I want to create a custom Route Locator, but base it off of a lot of the SimpleRouteLocator functionality, I can't do so without copying and pasting a lot of the functionality that is in the Simple Route Locator.  Instead, if we make the fields available via protected getters, we can inherit a lot of the core functionality and just customize the portion we want, such as the getMatchingRoute method.

Does that help?

ryanjbaxter:
I can buy some of this, but some of those variables are just used in logging in `getMatchingRoute`, maybe we could add a utility method for that functionality and make that protected?

TheSentinel454:
Yea, that makes sense. Let me try to rework this a bit.

Thanks for the feedback!

-Luke Tornquist

TheSentinel454:
@ryanjbaxter I've updated the PR based on your suggestions.  I also did some updating of the logic to kind of clean it up, such as adding the optional return type for the getZuulRoute method.

Let me know what you think.

spencergibb:
Master is still Java 7, please remove the optional reference

PR #1951, opened 2017-05-16
GH-1916: Spring Boot Auto-configuration based Zuul Server and Proxy

### LINKED ISSUE #1916: Move ZuulConfiguration use of @ConditonalOnMissionBean to auto-config, ala eureka client.
So it seems that `ZuulConfiguration.java` is no AutoConfiguration Bean but still uses `@ConditionOnMissingBean` for checking for the absence of certain beans.

But `@ConditionalOnMissionBean`s seems to be discouraged for non-AutoConfiguration Beans according to the docs:

> The condition can only match the bean definitions that have been processed by the application context so far and, as such, it is strongly recommended to use this condition on auto-configuration classes only. If a candidate bean may be created by another auto-configuration, make sure that the one using this condition runs after.

bijukunjummen:
Fixes #1916 

* Moves Zuul Proxy and Zuul Server configuration to be based on Spring-Boot Auto-configuration


spencergibb:

PR #1965, opened 2017-05-20
simplify org.springframework.cloud.netflix.zuul.filters.ProxyRequestH…

spccold:
simplify org.springframework.cloud.netflix.zuul.filters.ProxyRequestHelper#setResponse

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
@@ -157,22 +157,14 @@ public class ProxyRequestHelper {
 		}
 
 		boolean isOriginResponseGzipped = false;
-        if (headers.containsKey(CONTENT_ENCODING)) {
-            List<String> collection = headers.get(CONTENT_ENCODING);
-            for (String header : collection) {
-                if (HTTPRequestUtils.getInstance().isGzipped(header)) {
-                    isOriginResponseGzipped = true;
-                    break;
-                }
-            }
-        }
-		context.setResponseGZipped(isOriginResponseGzipped);
-
 		for (Entry<String, List<String>> header : headers.entrySet()) {
 			String name = header.getKey();
 			for (String value : header.getValue()) {
-				context.addOriginResponseHeader(name, value);
-				if (name.equalsIgnoreCase(CONTENT_LENGTH)) {
+				if (name.equalsIgnoreCase(HttpHeaders.CONTENT_ENCODING)
+						&& HTTPRequestUtils.getInstance().isGzipped(value)) {
+					isOriginResponseGzipped = true;
+				}
+				if (name.equalsIgnoreCase(HttpHeaders.CONTENT_LENGTH)) {
 					context.setOriginContentLength(value);
 				}
 				if (isIncludedHeader(name)) {
@@ -180,6 +172,7 @@ public class ProxyRequestHelper {
 				}
 			}
 		}
+		context.setResponseGZipped(isOriginResponseGzipped);
 	}
 
 	public void addIgnoredHeaders(String... names) {


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java
```diff
@@ -156,23 +156,16 @@ public void setResponse(int status, InputStream entity,
 			context.setResponseDataStream(entity);
 		}
 
-		HttpHeaders httpHeaders = new HttpHeaders();
```
spencergibb: You'll have to explain why you removed this other than simplify.

spccold: no need to create HttpHeaders and copy all headers from headers, isOriginResponseGzipped can get from headers directly

spccold: the second headers is the paramter(MultiValueMap<String, String> headers)

spencergibb: Does it work case insensitive?

spccold: i rewrite setResponse and it work case insensitive

mukteshkrmishra: Can you include some more test as well.

spencergibb: Actually, a benchmark whether this actually improves performance or not would be useful. My gut tells me calling `equalsIgnoreCase` is probably slower than the built-in case insensitivity in Spring's `HttpHeaders`.

spccold: paramter(MultiValueMap<String, String> headers) come from two place, one is `org.springframework.cloud.netflix.zuul.filters.route.SimpleHostRoutingFilter.revertHeaders(Header[])`, headers is a `LinkedMultiValueMap` type, another is `org.springframework.cloud.netflix.ribbon.RibbonHttpResponse.getHeaders()`, headers is a `HttpHeaders` type, so i write a performance test between improved setResponse and original setResponse(`some common unnecessary be removed`), code like below:

~~~java
import java.io.IOException;
import java.io.InputStream;
import java.util.List;
import java.util.Map.Entry;

import org.springframework.http.HttpHeaders;
import org.springframework.util.MultiValueMap;

import com.netflix.zuul.context.RequestContext;
import com.netflix.zuul.util.HTTPRequestUtils;

/**
 * @author spccold
 */
public class App {

	public static void main(String[] args) throws Exception {
		MultiValueMap<String, String> headers = new HttpHeaders();
		//MultiValueMap<String, String> headers = new LinkedMultiValueMap<>();
		headers.add(HttpHeaders.CONTENT_ENCODING, "gzip");
		headers.add(HttpHeaders.CONTENT_LENGTH, "10");
		headers.add(HttpHeaders.CONTENT_TYPE, "application/json");
		
		// initialize RequestContext
		RequestContext.getCurrentContext();
		
		int time = 100000;
		System.out.print("time consume(ms): ");
		for (int i = 0; i < 10; i++) {
			long start = System.currentTimeMillis();
			for (int j = 0; j < time; j++) {
				improvedSetResponse(null, headers);
				//originalSetResponse(null, headers);
			}
			long end = System.currentTimeMillis();
			System.out.print((end - start)+" ");
		}
	}

	public static void improvedSetResponse(InputStream entity, MultiValueMap<String, String> headers)
			throws IOException {
		RequestContext context = RequestContext.getCurrentContext();
		if (entity != null) {
			context.setResponseDataStream(entity);
		}

		boolean isOriginResponseGzipped = false;
		for (Entry<String, List<String>> header : headers.entrySet()) {
			String name = header.getKey();
			for (String value : header.getValue()) {
				if (name.equalsIgnoreCase(HttpHeaders.CONTENT_ENCODING)
						&& HTTPRequestUtils.getInstance().isGzipped(value)) {
					isOriginResponseGzipped = true;
				}
				if (name.equalsIgnoreCase(HttpHeaders.CONTENT_LENGTH)) {
					context.setOriginContentLength(value);
				}
			}
		}
		context.setResponseGZipped(isOriginResponseGzipped);
	}

	public static void originalSetResponse(InputStream entity, MultiValueMap<String, String> headers)
			throws IOException {
		RequestContext context = RequestContext.getCurrentContext();
		if (entity != null) {
			context.setResponseDataStream(entity);
		}

		HttpHeaders httpHeaders = new HttpHeaders();
		for (Entry<String, List<String>> header : headers.entrySet()) {
			List<String> values = header.getValue();
			for (String value : values) {
				httpHeaders.add(header.getKey(), value);
			}
		}
		boolean isOriginResponseGzipped = false;
		if (httpHeaders.containsKey(HttpHeaders.CONTENT_ENCODING)) {
			List<String> collection = httpHeaders.get(HttpHeaders.CONTENT_ENCODING);
			for (String header : collection) {
				if (HTTPRequestUtils.getInstance().isGzipped(header)) {
					isOriginResponseGzipped = true;
					break;
				}
			}
		}
		context.setResponseGZipped(isOriginResponseGzipped);

		for (Entry<String, List<String>> header : headers.entrySet()) {
			String name = header.getKey();
			for (String value : header.getValue()) {
				context.addOriginResponseHeader(name, value);
				if (name.equalsIgnoreCase(HttpHeaders.CONTENT_LENGTH)) {
					context.setOriginContentLength(value);
				}
			}
		}
	}
}
~~~

test result:

~~~
headers type            setResponse implementation   time consume(ms) for 10 time tests

HttpHeaders              improvedSetResponse         76 51 12 12 12 11 15 16 12 12
LinkedMultiValueMap      improvedSetResponse         94 49 18 13 13 14 13 17 11 12
HttpHeaders              originalSetResponse         316 281 242 246 111 146 106 174 367 107
LinkedMultiValueMap      originalSetResponse         354 273 230 228 147 139 106 149 238 108
~~~

So my  RP is useful!!!

PR #1972, opened 2017-05-22
FeignClient Health Indicator Feature

khannedy:
The idea of this feature is to automatically create `HealthIndicator` for FeignClient

```java
@FeignClient(name = "healthClient", health = "healthMethod")
public interface HealthClient {

    @RequestMapping(method = RequestMethod.GET, value = "/health")
    ResponseEntity<String> healthMethod();

}
```


spencergibb:
Can you explain more about why you want this feature and what exactly this feature does?

khannedy:
Sometimes our app depend on other app, so to make our app run well, we need to make sure that other app is live. To do that, we need to create `HealthIndicator` manually.

With the ability to automatically create `HealthIndicator` in `FeignClient`, it's really helpful because we  don't need to create `HealthIndicator` manually

spencergibb:
I've marked it for team discussion, I don't know if this is something everyone would want/need.

andirdju:
cmiiw, at the moment, there are no feign-based built-in health-indicators,

https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html#_auto_configured_healthindicators

is it possible to make this as one of those?

@khannedy , @spencergibb

ryanjbaxter:
I dont understand why you wouldn't just use the health endpoint from actuator.

khannedy:
@ryanjbaxter actually this code will automatically register HealthIndicator for actuator, so it will automatically displayed in actuator endpoint

ryanjbaxter:
@khannedy it looked to me like you were creating another endpoint, `/health`, I was confused why you needed a new endpoint.

khannedy:
@ryanjbaxter not it's just feign endpoint for target server, maybe target server using different endpoint for healthcheck, so we can customize the endpoint, but in the app, the health indicator is registered in actuator feign 

for example we have `auth service`

`http://auth-service/`

with health endpoint

`http://auth-service/healthcheck`

in our app, example `retail service`, so we can create feign for auth service

```java
@FeignClient(name="authService", url="http://auth-service", health="healthMethod")
interface AuthFeign {

   @RequestMapping("/healthcheck")
   ResponseEntity<String> healthMethod();

}
```

our `retail-service` endpoint still using actuator health endpoint

`http://retail-service/health`

and the result for example 

```json
{
   "status" : "UP",
   "authFeign" : {
      "status" : "UP",
      "responseCode" : 200,
      "responseBody" : "..."
   },
   "redis" : {
      "status" : "UP"
   },
   "..." : {
   }
}
```

spencergibb:
I think a generic declarative HTTP based health indicator might be something we could think about, but we don't want to maintain a Feign based health indicator.

PR #1907, opened 2017-05-02
change HttpServletResponse to HttpServletRequest

spccold:

PR #2007, opened 2017-06-01
change HttpServletResponse to HttpServletRequest in SendResponseFilter

shuiky:
to check Servlet API 3.1 support is wrong.
i saw same bug in #1907


lowzj:
It maybe not the same as #1907. In #1907, it checks if `getContentLengthLong` exists in `HttpServletRequest`, and here it's correct to check if `setContentLengthLong` exists in `HttpServletResponse`.

Issue #2026, opened 2017-06-07
Centralize Configuration And Creation Of HTTP Clients


ryanjbaxter:
There are multiple places in Spring Cloud Netflix where we create an HTTP Client and each one is different.  We should try to centralize the configuration and creation of these clients in order to be consistent in behavior.

- `SimpleHostRoutingFilter`

- `OkHttpLoadBalancingClient`
- `FeignLoadBalancer`
- `RibbonLoadBalancingHttpClient`

ryanjbaxter:
Would also be nice to address #1684 as part of this work.

Issue #1684, opened 2017-02-10
Zuul with ribbon routing unable to validate custom ssl certificates


atiwari54:
In my application i want to search service with service id but my zuul is not connecting with eureka service, can any one help me out. 
1. when i hit http://localhost:8761/   then zuul is not showing there
2. If i comment the serviceid and enable the url then it will run as per requirement

here is my application.yml for zuul

```
spring:
  application:
    name: ZUUL

server:
  port: 8090

ribbon:
  eureka:
    enabled: true

eureka:
  client:
    serviceUrl:
      defaultZone: http://localhost:8761/eureka/

zuul:
  routes:
    test:
      path: /**
      serviceId: main
#     url: http://localhost:9080
```

spencergibb:
@atiwari54 unfortunately, we don't have enough information to help you. Please be more descriptive when opening issues. Include, versions, steps to reproduce and, possibly, a project that recreates the issue (not code pasted as comments).

Do you have `@EnableDiscoveryClient`? Do you have the eureka starter? Is zuul registered with eureka?

atiwari54:
@spencergibb, yes i have added `@EnableDiscoveryClient` in my zuul app. No zuul not getting register with eureka, all other microservice are  getting register.

i am using
1. spring 4.2.2.RELEASE version and spring-cloud-netflix  1.0.7.RELEASE version for ZUUL
2. spring 4.2.2.RELEASE version and spring-cloud-netflix  1.2.3.RELEASE version for EUREKA and MAIN

atiwari54:
@spencergibb, after certain changes zuul get register with eureka but now it giving me 

`com.netflix.zuul.exception.ZuulException: Forwarding error
	at org.springframework.cloud.netflix.zuul.filters.route.RibbonRoutingFilter.forward(RibbonRoutingFilter.java:140)
	at org.springframework.cloud.netflix.zuul.filters.route.RibbonRoutingFilter.run(RibbonRoutingFilter.java:105)
	at com.netflix.zuul.ZuulFilter.runFilter(ZuulFilter.java:112)
	at com.netflix.zuul.FilterProcessor.processZuulFilter(FilterProcessor.java:197)
	at com.netflix.zuul.FilterProcessor.runFilters(FilterProcessor.java:161)
	at com.netflix.zuul.FilterProcessor.route(FilterProcessor.java:120)
	at com.netflix.zuul.ZuulRunner.route(ZuulRunner.java:84)
[...]

PR #2033, opened 2017-06-10
SimpleHostRoutingFilter: Add overridable ssl context provider method

stfsy:
## Problem
We're having trouble with the hardcoded instance of **SSLContext** and our **self-signed certificates**. Tried using **ssl.keystore** etc., but our SSL Traffic is terminated by the load balancer and forwarded using http. Pretty sure we all heard that often enough, so i'll cut it short. 😄 

Setting [HttpsURLConnection.DefaultSocketFactory](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/HttpsURLConnection.html#setDefaultSSLSocketFactory(javax.net.ssl.SSLSocketFactory)) works but i'd prefer to costumize the built in SimpleHostRoutingFilter.

## Solution
Added a new method that provides the default SSLContext and can be overriden by subclasses.


ryanjbaxter:
This is actually one of the issues I plan on addressing with #2026 and there is an issue for this specific problem as well #1684.  I am of the opinion that we can make this change in the `1.3.x` branch and then in master we should implement the fix for #2026 (and #1684).  @spencergibb what do you think?

spencergibb:
Superseded by #2026

PR #2074, opened 2017-06-25
Add options for RibbonCommand to use separate thread pools for hystrix

yongsungyoon:
#### Background

When zuul creates `RibbonCommand`, it assigns `"RibbonCommand"` as a HystrixCommandGroupKey for hystrix. When using semaphore isolation, it's not a big deal. But in case of using THREAD isolation strategy, it will make all RibbonCommands to be executed in a single thread pool for hystrix. In some environments or by  an architectural decision, users might want to use separate thread pool for each service id. This PR is about providing options for this.

#### Implementation

I've added two properties of Zuul for thread isolation strategy.
```
zuul.threadPool.useSeparateThreadPools : boolean (false as default)
zuul.threadPool.threadPoolKeyPrefix : String ("" as default)
```

If setting `useSeparateThreadPools` to true, RibbonCommands will have HystrixThreadPoolKey whose value is same with commandKey. As a result, each RibbonCommand will be executed in a single thread pool that it is associated with based on commandKey (namely service ID). 

Additionally, this PR provides `threadPoolKeyPrefix` property for the convenience. If setting to a string, the string will be used as prefix for HystrixThreadPoolKey and it would be useful for the management purpose.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
@@ -359,9 +359,6 @@ public class ZuulProperties {
 		
 	}
 
-	@Data
-	@AllArgsConstructor
-	@NoArgsConstructor
 	public static class HystrixThreadPool {
 		/**
 		 * Flag to determine whether RibbonCommands should use separate thread pools for hystrix.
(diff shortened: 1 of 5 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -357,6 +359,26 @@ public boolean isCustomSensitiveHeaders() {
 		
 	}
 
+	@Data
```
ryanjbaxter: We are moving away from Lombok so we should not be adding any additional uses of it.

yongsungyoon: okay, I will remove lombok annotation and add related information into the documentation about this PR soon.

yongsungyoon:
@ryanjbaxter I've removed lombok annotation and added doc about this properties.

yongsungyoon:
@ryanjbaxter I've added test cases to verify thread pool key according to properties.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -88,12 +92,13 @@ protected static Setter getSetter(final String commandKey,
 					.getIntProperty(name, zuulProperties.getSemaphore().getMaxSemaphores());
 			setter.withExecutionIsolationSemaphoreMaxConcurrentRequests(value.get());
 		} else	{
-			// TODO Find out is some parameters can be set here
+			if (zuulProperties.getThreadPool().isUseSeparateThreadPools()) {
```
spencergibb: Should be `else if`

yongsungyoon:
@spencergibb I've fixed it. :-)

PR #2080, opened 2017-06-28
Centralizing HTTP client creation and configuration

### LINKED ISSUE #2026: Centralize Configuration And Creation Of HTTP Clients

ryanjbaxter:
Fixes #2026
Requires https://github.com/spring-cloud/spring-cloud-commons/pull/222
Uses Apache HTTP client factories in Spring Cloud Commons to create and use Apache HTTP Clients throughout Netflix.  The PR currently just replaces the Apache clients.  Still need to do the work to do the same for the OK HTTP clients.


docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -11,6 +11,10 @@
 
 include::intro.adoc[]
 
+== HTTP Clients
```
spencergibb: Is this a todo?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RibbonCommandFactoryConfiguration.java
```diff
@@ -124,10 +120,6 @@ public OnRibbonOkHttpClientCondition() {
 			super(ConfigurationPhase.PARSE_CONFIGURATION);
 		}
 
-		@Deprecated //remove in Edgware"
-		@ConditionalOnProperty("zuul.ribbon.okhttp.enabled")
```
spencergibb: We should mention deprecations removed in the release notes.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -105,51 +86,78 @@
...
 
 	@PostConstruct
 	private void initialize() {
-		this.httpClient = newClient();
-		this.connectionManagerTimer.schedule(new TimerTask() {
-			@Override
-			public void run() {
-				if (SimpleHostRoutingFilter.this.connectionManager == null) {
-					return;
+		if(!customHttpClient) {
```
spencergibb: Do you think we need this filter creating an http client anymore? Maybe deprecate it for removal later?

spring-cloud-netflix-core/src/test/java/OkHttpClientConfigurationTests.java
```diff
@@ -0,0 +1,160 @@
...
+
+	@Autowired
+	LoadBalancerFeignClient feignClient;
+
+	@Autowired
+	OkHttpRibbonCommandFactory okHttpRibbonCommandFactory;
+
+	@Test
+	public void testFactories() {
+		assertTrue(OkHttpClientConnectionPoolFactory.class.isInstance(connectionPoolFactory));
```
spencergibb: I'd prefer `Assertions.assertThat(connectionPoolFactory).isInstanceOf(OkHttpClientConnectionPoolFactory.class)`. That way, if it fails, we have a decent message :-)

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/FeignHttpClientUrlTests.java
```diff
@@ -58,7 +58,7 @@
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringBootTest(classes = FeignHttpClientUrlTests.TestConfig.class, webEnvironment = WebEnvironment.DEFINED_PORT, value = {
 		"spring.application.name=feignclienturltest", "feign.hystrix.enabled=false",
-		"feign.okhttp.enabled=false" })
+		"spring.cloud.httpclientfactories.ok.enabled=false" })
```
spencergibb: Are the properties documented somewhere?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RibbonCommandFactoryConfiguration.java
ryanjbaxter: I will label this as `needs release note`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
ryanjbaxter: I did this to support the pattern where the app provides its own client.  If the app does nothing and accepts all the defaults than we will use the factories from commons.  So yes we still need it.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/FeignHttpClientUrlTests.java
ryanjbaxter: Yes, they are documented in Spring Cloud Commons.  However, this property doesnt even need to be set so I am going to reverse it :)

PR #2081, opened 2017-06-28
Use constant utility classes instead of constant interfaces

gzurowski:
Basically just following best practices and using constant utility classes instead of interfaces for exporting constants.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixConstants.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixConstants.java
@@ -19,8 +19,12 @@ package org.springframework.cloud.netflix.hystrix;
 /**
  * @author Spencer Gibb
  */
-public interface HystrixConstants {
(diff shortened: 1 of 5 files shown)


ryanjbaxter:
Not sure I see the need for these changes

gzurowski:
When looking through the code I appreciated that these constants are now available, but thought that they shouldn't be exposed via interfaces. As initially stated, it's just for following best practices.

ryanjbaxter:
I mean I can see the argument that some people make about not putting constants in an interface but from a practical perspective I dont see much of a benefit personally.

gzurowski:
Item 19 in Joshua Bloch's Effective Java goes into detail about this, but the main reason is that if a class that implements a constant interface changes further down the road, i.e. does not need these constants, it must still implement it for maintaining backwards compatibility. As this can be simply avoided by using constant utility classes, I created this PR. But again, it's more about best practices, so I won't struggle over this one. ;)

ryanjbaxter:
So we discussed this with the team and the decision was made to merge it.  Could you just make sure to use static imports where appropriate?

Issue #2084, opened 2017-06-30
Eureka registration prefer hostname instead of IP


dhanvi:
Is there a way for eureka to register via hostname instead of IP, I have tried setting eureka.instance.preferIpAddress to false but it takes hostname some times and IP address sometime. [Related stackoverflow Question](https://stackoverflow.com/questions/44689202/springboot-client-unable-register-with-eureka-using-docker-container-id)

Using hostname is useful when running micro services with docker swarm, as docker-id (which is same as hostname of the container) of the container is resolvable to all the containers (even if the containers are on different nodes/VM's)

I am aware of setting preferIpAddress as true and [ignoring networks](http://cloud.spring.io/spring-cloud-static/spring-cloud.html#ignore-network-interfaces) but I am looking for a straight forward solution.

Thanks in Advance.

ryanjbaxter:
Could you provide your configuration?

dhanvi:
Eureka
```
eureka:
  client:
    registerWithEureka: false
    fetchRegistry: false
    serviceUrl:
      defaultZone: http://discovery-eureka:8761/eureka/
```
client
```
eureka:
  client:
    healthcheck:
      enabled: true
    serviceUrl:
      defaultZone: http://discovery-eureka:8761/eureka/
  instance:
    preferIpAddress: false
```

Here discovery-eureka is the docker service name (In docker swarm you can hit the containers using the service name) and we have only one instance (container) of eureka running.

spencergibb:
We try and guess. What we get comes from java's `NetworkInterfaces` class [see here](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/commons/util/InetUtils.java#L78-L128). In some cases, when the hostname can't be determined by java, the IP is returned as the hostname. The only **sure** way to ensure the hostname is to **set** the hostname in `eureka.instance.hostname`.

spencergibb:
If the hostname is available as an env var (say `THE_HOST`) you can do `eureka.instance.hostname=${THE_HOST}`.

dhanvi:
Thank you! @spencergibb for mentioning setting the environment variable. By setting the environment variable it's working exactly as I needed.

I am setting the environment variable at the run time of the container using a shell script as the ENTRYPOINT in Dockerfile, the shell script sets the hostname environment variable and starts the jar.

dhanvi:
[...]

PR #2092, opened 2017-07-05
Document prefer hostname

dhanvi:
#2084


dhanvi:
@spencergibb changed it to tip instead of section

PR #2105, opened 2017-07-08
Provide more Zuul route details in the routes actuator endpoint

gzurowski:
Add a query parameter to control the level of details returned with the routes actuator endpoint. If requested, return all available Zuul route details including ID, prefix, sensitive headers, etc.

Example:

A detailed rendition of the routes endpoint can be requested with `/routes?format=details`. The output looks as follows:

```
 {
 	"/stores/**": {
 		"id": "stores",
 		"fullPath": "/stores/**",
 		"location": "http://localhost:8081",
 		"path": "/**",
 		"prefix": "/stores",
 		"retryable": false,
 		"customSensitiveHeaders": false,
 		"prefixStripped": true
 	}
 }
```

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RoutesEndpoint.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RoutesEndpoint.java
@@ -18,11 +18,11 @@ package org.springframework.cloud.netflix.zuul;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-import lombok.Data;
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RoutesEndpoint.java
```diff
@@ -59,4 +64,55 @@ public RoutesEndpoint(RouteLocator routes) {
...
+		}
+		return map;
+	}
+
+	/**
+	 * Container for exposing Zuul {@link Route} details as JSON.
+	 */
+	@JsonPropertyOrder({ "id", "fullPath", "location" })
+	@JsonInclude(JsonInclude.Include.NON_EMPTY)
+	@Data
```
ryanjbaxter: We are slowly phasing out Lombok, could you avoid using it here?

gzurowski:
Thanks for your feedback, @ryanjbaxter.  I have removed Lombok and updated the documentation.

PR #2128, opened 2017-07-14
Remove Lombok from spring-cloud-netflix-eureka-server module

gzurowski:
This removes Lombok from the spring-cloud-netflix-eureka-server module.

Please note that I have used Apache Common's EqualsBuilder, HashCodeBuilder and ToStringBuilder in `@ConfigurationProperties` annotated configuration property classes in order to simplify code and future maintenance.

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDashboardProperties.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDashboardProperties.java
@@ -16,8 +16,9 @@
 
 package org.springframework.cloud.netflix.eureka.server;
 
-import lombok.Data;
-
+import org.apache.commons.lang.builder.EqualsBuilder;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+import org.apache.commons.lang.builder.ToStringBuilder;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
@@ -26,7 +27,6 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * @author Dave Syer
  */
 @ConfigurationProperties("eureka.dashboard")
-@Data
 public class EurekaDashboardProperties {
 
 	/**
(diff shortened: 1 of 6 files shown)


gzurowski:
No, they are not necessary, but I have added them so that IDEs (both Eclipse and IntelliJ) will not reformat specific areas of the code when applying the formatter.  This preserves the areas of the code that are more readable when formatted manually, often useful for code blocks that make use of builders or longer method invocation chains.  I can remove this, if it's unwanted.

ryanjbaxter:
We havent used it in the past, so I say remove it

gzurowski:
@ryanjbaxter, removed the formatter markers as requested.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDashboardProperties.java
```diff
@@ -16,8 +16,9 @@
 
 package org.springframework.cloud.netflix.eureka.server;
 
-import lombok.Data;
-
+import org.apache.commons.lang.builder.EqualsBuilder;
```
bijukunjummen: @gzurowski, @ryanjbaxter - Isn't commons-lang3 a better choice for `Equals`, `HashCode` and `ToString` - so the imports will be `org.apache.commons.lang3.builder.EqualsBuilder`. I see that commons-lang3 is already transitively imported.

gzurowski: Please note that commons-lang3 is not (transitively) imported on the Eureka modules, so as long as there is nothing that we actually need from lang3, I would just leave this as-is.

spencergibb: As a point, we try not to rely on libraries like commons-lang and guava. See 31d19fd51ad7bfdc035c09928cd677669639e764

spencergibb: As a point, we try not to rely on libraries like commons-lang and guava. See 31d19fd51ad7bfdc035c09928cd677669639e764

PR #2145, opened 2017-07-21
Incorrect statusPageUrl and healthCheckUrl are reported to Eureka if …

Aloren:
…management.context-path is not set

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -24,7 +24,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.Map;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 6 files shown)


spencergibb:
So doesn't contextPath affect management if management.port is not set?

Aloren:
Forgot about that cases. Thanks. I will add handling of other cases as well.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
```diff
@@ -113,38 +113,46 @@ public EurekaClientConfigBean eurekaClientConfigBean() {
...
-		int managementPort = Integer.valueOf(propertyResolver.getProperty("management.port", String.valueOf(nonSecurePort)));
-		String managementContextPath = propertyResolver.getProperty("management.contextPath", propertyResolver.getProperty("server.contextPath", "/"));
+		Integer managementPort = propertyResolver.getProperty("management.port", Integer.class);// nullable. should be wrapped into optional
+		String managementContextPath = propertyResolver.getProperty("management.contextPath");// nullable. should be wrapped into optional
+
 		EurekaInstanceConfigBean instance = new EurekaInstanceConfigBean(inetUtils);
 		instance.setNonSecurePort(nonSecurePort);
 		instance.setInstanceId(getDefaultInstanceId(propertyResolver));
 		instance.setPreferIpAddress(preferIpAddress);
-		if (managementPort != nonSecurePort && managementPort != 0) {
```
Aloren: Why there was a check for managementPort != nonSecurePort?
Shouldn't we set urls if management port is equal to nonSecurePort?

spencergibb:
Looks like a rebase is in order. Also, there should be a test case for management context path set, but not management port.

Aloren:
Rebased to masted and added test

ryanjbaxter:
@Aloren sorry but this needs another rebase.

Aloren:
:) I've refactored my solution a bit to return `ManagementMetadata` and rebased. Please check also test cases I may have missed smth. Thanks!

Issue #1635, opened 2017-01-20
SimpleZuulProxyApplicationTests.getOnSelfWithComplexQueryParam fails during build


spencergibb:
Doesn't fail running individually in IDE.

```
org.junit.ComparisonFailure: expected:<foo=%[7B%22project%22%3A%22stream%22%2C%22logger%22%3A%22javascript%22%2C%22platform%22%3A%22javascript%22%2C%22request%22%3A%7B%22url%22%3A%22https%3A%2F%2Ffoo%]2Fadmin> but was:<foo=%[257B%2522project%2522%253A%2522stream%2522%252C%2522logger%2522%253A%2522javascript%2522%252C%2522platform%2522%253A%2522javascript%2522%252C%2522request%2522%253A%257B%2522url%2522%253A%2522https%253A%252F%252Ffoo%25]2Fadmin>
```

spencergibb:
Ignored in 5d0b2fae861fbdf9ae0808925abbbeb40567d4ac

Needs to be fixed as part of 1.3.0

OlgaMaciaszek:
Already done.

PR #2146, opened 2017-07-21
Restore ignored test - works cleanly for a full build

bijukunjummen:
Restored ignored test for issue #1635 , appears to work without issues locally. Sending a PR to run it on Circle Ci also.


bijukunjummen:
The test appears to be good @spencergibb, tested it standalone, in local and now on circleci!

PR #2158, opened 2017-07-25
Refresh Zuul routes on property changes

### LINKED ISSUE #705: Zuul routes cannot be removed dynamically
I noticed that it is possible to add routes via Spring cloud Config (or by changing the yml directly), but it is not possible to remove routes.

Scenario:
- routed defined:

```
zuul:
ignored-services: "*"
  routes:
    echo3:
      path: /echo3/**
      url: http://www.aaa.com
```
- route shows fine in /admin/routes
- remove route from yml
- hit /admin/refresh() endpoint. The response mentions that changes were detected about the route
- Problem: /admin/routes still shows the route

bijukunjummen:
Fixes #705

Zuul routes and mappings will be refreshed on  configuration change. There are two issues addressed here:
1. `ZuulProperties` were not being refreshed correctly - the routes were being updated when new routes were added, but was not being removed.
2. `ZuulHandlerMapping` which uses `ZuulProperties` for the routes configuration was again able to handle new routes but not able to remove them. This was because the underlying `handlerMap` in `AbstractUrlHandlerMapping` is not exposed to be modified. 

The fix is to add `@RefreshScope` to both these beans.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulServerAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulServerAutoConfiguration.java
@@ -33,6 +33,7 @@ import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.cloud.client.actuator.HasFeatures;
 import org.springframework.cloud.client.discovery.event.HeartbeatEvent;
 import org.springframework.cloud.client.discovery.event.HeartbeatMonitor;
+import org.springframework.cloud.context.config.annotation.RefreshScope;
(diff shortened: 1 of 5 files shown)


bijukunjummen:
The build failure appears unrelated to the changes @spencergibb, will check and see what else could be the issue here.

bijukunjummen:
Yes, will try @spencergibb, the issue that I saw specifically with `ZuulHandlerMapping` is that the underlying `handlerMap` of `AbstractUrlHandlerMapping` was not exposed and did not appear to have anything to remove from it. Will review once more though.

spencergibb:
Yup, we're aware of that. @dsyer may have had some ideas.

bijukunjummen:
Closing the PR for now @spencergibb, we can revisit it for 2.0 timelines.

PR #2160, opened 2017-07-26
Fix check then act issue

obourgain:
If two threads are calling the `buildContentData()` method concurrently, thanks to the synchronized, they will not execute concurrently, but the second thread to enter the method will either overwrite the work that was done by the first thread with garbage values or hit an exception.

By checking again inside the synchronized that the `contentData` is still null, we know that we are not going to encounter this issue.

`contentData` also has to be marked `volatile` and written last to have proper double checked locking.

I suspect that the code would be simpler and the performance not sensibly different if the unsynchronized null check was removed and every method was calling `buildContentData()` (with the null check inside). I can change this PR to do this if you also think it is better and simpler.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/FormBodyWrapperFilter.java
@@ -140,7 +140,7 @@ public class FormBodyWrapperFilter extends ZuulFilter {
 
 		private HttpServletRequest request;
 
-		private byte[] contentData;
+		private volatile byte[] contentData;
 
 		private MediaType contentType;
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@obourgain Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2160)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2160) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@obourgain Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2160)!

ryanjbaxter:
@obourgain can you merge the latest changes from master into your branch?  It should fix the circleci build.

PR #2169, opened 2017-07-29
RouteLocator to use object instead of string literal for matching route.

msamad:
This is to open ways for fine-grained route matching instead of only
path. e.g. matching route based on path and http method or adding
header values into the mix in the future. This will also give the ability
to separate out GET, PUT, POST etc and scale accordingly if needed.
Also, header values can be added to the mix later to route to
different endpoints.


ryanjbaxter:
Can you provide some examples of how this will be used?

msamad:
Sure, this is the first step to enable a fine-grained route matching - planning to do changes in small PRs to keep it simple.
Next step will be:
- In SimpleRouteLocator, use bean for PathMatcher so matching logic can be customized.

With the above in place, we can
- have the ability to limit routing to back-end for specific http methods.
```yaml
zuul:
  routes:
    user-service:
      path: /user/**
      # Only GET and PUT are allowed
      methods: GET, PUT
      url: http://localhost:8080
```
- have the ability to route to different endpoints based on http methods.
```yaml
zuul:
  routes:
    user-service-read:
      path: /user/**
      # Only GET is allowed
      methods: GET
      url: http://localhost:8080
    user-service-write:
      path: /user/**
      # Only PUT and POST are allowed
      methods: PUT, POST
      url: http://localhost:8081
```

- More complex example would be to route based on header values.
```yaml
zuul:
  routes:
    user-service-read-debug-server:
      path: /user/**
      # Only GET is allowed
      methods: GET
      header:
        debugserver: true
      url: http://localhost:8090
    user-service-read:
      path: /user/**
      # Only GET is allowed
      methods: GET
      url: http://localhost:8080
```

Also, JWTs can be used to further filter the routing but don't want to complicate it too much at this point - that is for another day.

Hope the examples above make sense, let me know if you want me to put more examples.

Cheers!

ryanjbaxter:
Thanks.  I have labeled this as "for team discussion" so I can get the opinions of others.  We will probably talk about it later this week.

In the mean time can you add some documentation about how this will be used to the docs?  Also can you merge in the latest changes from master into your branch, it should fix the circleci build.

msamad:
Thanks. The circleci build is passing now.

msamad:
Hi @ryanjbaxter,
About the docs, I still need to do the second part of the implementation. I was thinking of doing that in another PR, the docs would not make much sense without that. Do you still want me to the docs now or I can do them with the second PR? 
Or, I can do the full change in this PR - a complete set with the docs?
Your views?

ryanjbaxter:
After discussing with the team today, the general consensus was that we were in favor of this functionality :)

The one thing we wanted to see as far as an implementation is to create a new interface instead of changing `RouteLocator`.

msamad:
Thanks @ryanjbaxter . I'll do a new interface but could you please elaborate a bit more on that. The idea was to get it working in SimpleRouteLocator hence the change in RouteLocator. How do you see the new interface working with SimpleRouteLocator? Just have it extend the new interface or do you have something else in mind?

ryanjbaxter:
If you extend `RouteLocator` and add the new method and then have `SimpleRouteLocator` implement that new interface shouldnt that work?

spencergibb:
I don't think `SimpleRouteLocator` needs to be involved necessarily. The new interface should use the actual `Request` rather than an object that gathers information from a request.

msamad:
I like the idea of using the actual `Request` and think using `RequestContext` will give more freedom. It will allow to even do some manipulation (if needed) before the request matching happens.

msamad:
By the way, there is another matching happening in `ZuulHandlerMapping.lookupHandler - PatternMatchUtils.simpleMatch(ignored, urlPath)` - this one is just a check though
```java
@Override
	protected Object lookupHandler(String urlPath, HttpServletRequest request) throws Exception {
		if (this.errorController != null && urlPath.equals(this.errorController.getErrorPath())) {
			return null;
		}
		String[] ignored = this.routeLocator.getIgnoredPaths().toArray(new String[0]);
		if (PatternMatchUtils.simpleMatch(ignored, urlPath)) {
			return null;
		}
		RequestContext ctx = RequestContext.getCurrentContext();
		if (ctx.containsKey("forward.to")) {
			return null;
		}
		if (this.dirty) {
			synchronized (this) {
				if (this.dirty) {
					registerHandlers();
					this.dirty = false;
				}
			}
		}
		return super.lookupHandler(urlPath, request);
	}
```

msamad:
Hi guys,
I am done for now, have updated the docs as well. Could anyone please review and see if I've made a mess of all or have actually done something that will work? :)

All yours now.

Cheers!

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/CompositeRouteLocator.java
```diff
@@ -69,6 +71,17 @@ public Route getMatchingRoute(String path) {
 	}
 
 	@Override
+	public Route getMatchingRoute(RequestContext requestContext) {
```
spencergibb: I was thinking the actual `HttpServletRequest`, rather than a zuul context object. That can be accessed at any point since it's a static threadlocal bound instance.

msamad:
Hi @spencergibb,
I've made the change to use `HttpSerlvetRequest` instead of `RequestContext`. I squashed my commit with the old one to keep it in 1 commit and that has made your comment outdated, didn't know it will do that. Will try avoiding squashing next time :)

Cheers!

msamad:
And not sure why circleci failed this time, all tests are passing on my local.

spencergibb:
Squashing is fine. I can get to the outdated code.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/Route.java
```diff
@@ -27,7 +27,7 @@
 public class Route {
 
 	public Route(String id, String path, String location, String prefix,
-			Boolean retryable, Set<String> ignoredHeaders) {
+			Boolean retryable, Set<String> ignoredHeaders, Set<String> methods) {
```
spencergibb: We should leave existing constructors, deprecate and add new.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/RouteLocator.java
```diff
@@ -22,7 +22,7 @@
 /**
  * @author Dave Syer
  */
-public interface RouteLocator {
+public interface RouteLocator extends RouteMatcher {
```
spencergibb: I'm not sure if this is quite what I was looking for. I'll have to noodle on it.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -51,8 +55,10 @@
 	private String dispatcherServletPath = "/";
 	private String zuulServletPath;
 
-	private AtomicReference<Map<String, ZuulRoute>> routes = new AtomicReference<>();
+	private AtomicReference<Map<String, List<ZuulRoute>>> routes = new AtomicReference<>();
```
spencergibb: Can you explain this change?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -51,8 +55,10 @@
...
 	private int order = DEFAULT_ORDER;
+	
+	private UrlPathHelper urlPathHelper = new UrlPathHelper();
```
spencergibb: Did this come from a merge commit?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -83,15 +92,23 @@ public SimpleRouteLocator(String servletPath, ZuulProperties properties) {
...
-	public Route getMatchingRoute(final String path) {
-
-		return getSimpleMatchingRoute(path);
+	public Route getMatchingRoute(HttpServletRequest request) {
+		return getSimpleMatchingRoute(request);
+	}
 
+	@Override
+	public Route getMatchingRoute(final String path) {
+		// Nothing to do here, will do matching using RequestContext
```
spencergibb: Either this should continue to work or throw an exception, doing nothing is a bad experience for a developer expecting this to work.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/discovery/DiscoveryClientRouteLocator.java
```diff
@@ -93,18 +94,23 @@ public void addRoute(ZuulRoute route) {
 	}
 
 	@Override
-	protected LinkedHashMap<String, ZuulRoute> locateRoutes() {
-		LinkedHashMap<String, ZuulRoute> routesMap = new LinkedHashMap<String, ZuulRoute>();
+	protected LinkedHashMap<String, List<ZuulRoute>> locateRoutes() {
```
spencergibb: Again, why the change to list?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -244,8 +244,11 @@ public void init() {
 
 		private boolean customSensitiveHeaders = false;
 
+		private Set<String> methods = new LinkedHashSet<>();
```
spencergibb: I prefer a way to do this, so that this class doesn't change for every new way of matching.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -106,27 +123,49 @@ protected Route getSimpleMatchingRoute(final String path) {
...
+		}
+		return null;
+	}
+	
+	protected ZuulRoute matchRoute(String path, Entry<String, List<ZuulRoute>> entry,
+			HttpServletRequest request) {
+		String method = request.getMethod();
+		String pattern = entry.getKey();
+		List<ZuulRoute> zuulRouteList = entry.getValue();
+		for (ZuulRoute zuulRoute : zuulRouteList) {
```
spencergibb: It might be worthwhile to look at https://github.com/spring-cloud/spring-cloud-gateway/blob/2.0.x/spring-cloud-gateway-core/src/main/java/org/springframework/cloud/gateway/config/GatewayProperties.java to see how I've done similar things. Hard coding matching code seems like a maintenance problem moving forward and combining them should be simpler than nested if's. What if I want to match on method without path? or host and method?

spencergibb: I think this is where I was thinking a new interface and new classes for matching would be helpful rather than shoehorning this into the old way. So the old way == path only, new way == multiple options.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/Route.java
msamad: I'll fix that.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
msamad: this is to do with how `locateRoutes` method works and creates the map.

Old implementation
```java
	protected Map<String, ZuulRoute> locateRoutes() {
		LinkedHashMap<String, ZuulRoute> routesMap = new LinkedHashMap<String, ZuulRoute>();
		for (ZuulRoute route : this.properties.getRoutes().values()) {
			routesMap.put(route.getPath(), route);
		}
		return routesMap;
	}
```
Since this is using path as the key, having multiple routes of the same path overwrites the value for the key.
```yaml
zuul:
  routes:
    users-read:
      path: /myusers/service
      url: http://user-service/readusers/newservice
      methods: GET
    users-write:
      path: /myusers/service
      url: http://user-service/readusers/newservice
      methods: PUT, POST
```
The will end up as a single route for key `/myusers/service`, so had to create a list to have multiple routes against the key `/myusers/service`. The `locateRoute` method is called by `DiscoveryClientRouteLocator` as well so creating a list seemed less dangerous than making an overall change for `locateRoutes`.

But now that I think about it, this may not be the ideal solution. We'll get `404 Not Found` if path + method is not found e.g. in the above yaml `PATCH` will give `404` while ideally it should return `405 Method Not Allowed` since resource exists but method is not available.

The above leads me to think that it should be something like below
```yaml
zuul:
  routes:
    users:
      path: /myusers/service
      url: http://default-user-service
      allowed-methods: GET, PUT, POST, PATCH
      - methods: GET
        url: http://read-user-service
      - methods: PUT, POST
        url: http://write-user-service
```
* If method is not in the list of `allowed-methods` e.g. `DELETE` in this case, then Zuul should throw `405 Method not Allowed`. 
* If method is allowed but not specified explicitly e.g. `PATCH` in this case, then default url should take effect.
* And of coarse, if nothing is specified then allow all methods to default url.

I might have taken it too far but this will really make it shine. Let me know what you think about it?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
msamad: This came from `PreDecorationFilter` and `LocationRewriteFilter`. Both were using `urlPathHelper.getPathWithinApplication(ctx.getRequest())` to get the requestUri.

The following piece of code is important as well to get the correct requestUri. `this.urlPathHelper.setRemoveSemicolonContent(properties.isRemoveSemicolonContent());`

Since extraction of requestUri has now moved to SimpleRouteLocator, this was required to move as well.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
msamad: This piece of code is unreachable. I'll throw exception here.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/discovery/DiscoveryClientRouteLocator.java
msamad: See comments above

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
msamad: I'll change this to an object so that can be used.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
msamad: Do you mean something like below
* Create interface `RouteMatcher`
* Create class `FancyRouteMatcher` implements `RouteMatcher`
* Create class `DiscoveryClientRouteMatcher` extends `FancyRouteMatcher`
* Create class `CompositeRouteMatcher`
* Update `PreDecorationFilter` to use old `RouteLocator` or new `RouteMatcher`

I think at least the above will need to be created to separate out a complete new functionality. Your views?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
msamad: Hi @spencergibb, what are you views on `Route` object as a wrapper/decorator over `ZuulRoute` instead of populating its own fields?

msamad:
Hi @spencergibb,
So I've taken your suggestion of creating a new set of classes to do the matching. I've committed a change, it is just a draft commit. I want to get your opinion on it before I go ahead and shine it a bit and write further tests. I'll fix/squash the commit later - easier to discuss this way.

Now 
* Everything works as it was as a default implementation i.e. matching using the path. It uses the existing `RouteLocator`.
* `zuul.route-matcher.fancy` property has been introduced which when set to `true` disables the simple matching using `RouteLocator` and enables the fancy matching using `RouteMatcher`.  I had to do a bit more to get all this working together so you would see some classes being created, some being extended. It was all because the `RouteLocator` was quite embedded in those classes and had to create new ones to use `RouteMatcher`.

`FancyZuulProxyApplicationTests` class would be the one to see how it works.

Yaml will look like below
```yaml
zuul:
  routes:
    users:
      path: /myusers/service
      url: http://default-user-service
      routeOptions:
        allowedMethods: GET, PUT, POST, PATCH
        methodOptions:
          GET:
            url: http://read-user-service
          PUT:
            url: http://write-user-service
          POST:
            url: http://write-user-service
```
* `GET` will end up with `http://read-user-service`
* `PUT` and `POST` will end up with `http://write-user-service`
* `PATCH` will end up with `http://default-user-service` i.e. default url since a specific one is not specified.

I'll later combine all the beans in one class to enable and disable simple and fancy route matching using a single property instead of current implementation where I've put the property on individual beans. Just in case you notice that. :)

Cheers!

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/RoutesRefreshedEventV2.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * Version 2: Use {@link RouteMatcher} instead of {@link RouteLocator}
+ * 
+ * @author Dave Syer
+ * @author Mustansar Anwar
+ */
+@SuppressWarnings("serial")
+public class RoutesRefreshedEventV2 extends ApplicationEvent {
```
spencergibb: No `V2`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulProxyAutoConfiguration.java
```diff
@@ -80,17 +85,34 @@ public HasFeatures zuulFeature() {
 
 	@Bean
 	@ConditionalOnMissingBean(DiscoveryClientRouteLocator.class)
+	@ConditionalOnProperty(value = "zuul.route-matcher.fancy", havingValue="false", matchIfMissing=true)
```
spencergibb: :-) probably not fancy, maybe advanced?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/ZuulServerAutoConfiguration.java
```diff
@@ -95,34 +99,69 @@ public HasFeatures zuulFeature() {
...
+	
+	@Bean
+	@ConditionalOnProperty(value = "zuul.route-matcher.fancy", havingValue="false", matchIfMissing=true)
 	public ApplicationListener<ApplicationEvent> zuulRefreshRoutesListener() {
 		return new ZuulRefreshListener();
 	}
+	
+	@Bean
+	@ConditionalOnProperty(value = "zuul.route-matcher.fancy", havingValue="true")
+	public ApplicationListener<ApplicationEvent> zuulRefreshRoutesListenerV2() {
```
spencergibb: Having two of everything is bad.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/MethodOption.java
```diff
@@ -0,0 +1,11 @@
+package org.springframework.cloud.netflix.zuul.filters;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
```
spencergibb: No more lombok, please.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/RouteOptions.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+import lombok.Data;
+
+@Data
+public class RouteOptions {
+
+	private Set<String> allowedMethods = new LinkedHashSet<>();
```
spencergibb: Having anything hardcoded, precludes a user from adding their own. Again, see my previous link to spring-cloud-gateway.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/routematcher/FancyRouteMatcher.java
```diff
@@ -0,0 +1,310 @@
...
+		if (log.isDebugEnabled()) {
+			log.debug("route matched=" + route);
+		}
+		// Set a default location for route
+		String location = route.getLocation();
+		RouteOptions routeOptions = route.getRouteOptions();
+
+		String requestMethod = request.getMethod();
+		if (routeOptions != null) {
+			if (!routeOptions.getAllowedMethods().isEmpty() && !routeOptions
```
spencergibb: Again, the hard coding here precludes custom options.

msamad:
Cool, I'll fix the above mentioned issues, just wanted to see if there are any major concerns about the approach - no big concerns raised so I'll take it as a yes :)

msamad:
Hi @spencergibb,
I've committed some changes that you asked for - you will see some code copied from the spring-cloud-gateway as well :). Wasn't able to spend much time on it last two weeks but hoping to get it all sorted this coming weekend. 

I've got rid of duplicate code and made route options dynamic instead of hard coded. I hope I am on a right track now - let me know what you think?

By the way, I've squashed the commits again to keep it to 1 single commit.
Cheers!

pom.xml
```diff
@@ -62,8 +63,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>${maven-compiler-plugin.version}</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
```
ryanjbaxter: I dont think we want to bump the java version

msamad:
Thanks @ryanjbaxter, I've got it back to java 1.7 now. Have updated the docs and added some tests to make sure it can read the routes properly from the yml file. 
Let me know if you guys want me to fix anything else. I'll squash all commits into one once you guys are happy with it.

Cheers!

msamad:
Thanks @ryanjbaxter. I've updated my branch, resolved conflicts and squashed all my commits into one.

This is the end result. :)
```yaml
zuul:
  route-matcher:
    advanced: true
  routes:
    advance-routing-compact:
      path: /advanceroutescompact/**
      url: http://user-default-service
      routeOptions:
        allowedMethods: PUT, POST, GET
        routeConditions:
        - user-get-operation=http://user-get-service,Method=GET
        - user-post-operation=http://user-post-service,Method=POST
    advance-routing-expanded:
      path: /advanceroutesexpanded/**
      url: http://user-default-service
      routeOptions:
        allowedMethods: PUT, POST, GET
        routeConditions:
        - id: user-get-operation
          url: http://user-get-service
          predicates:
          - Method=GET
        - id: user-post-operation
          url: http://user-post-service
          predicates:
          - Method=POST
```

Cheers!

msamad:
Hi @spencergibb, your thoughts?

msamad:
Hi @spencergibb @ryanjbaxter, 
Is there anything I need to do to have this merged please?

spencergibb:
Looks like another rebase is in order. I'll review it next week.

msamad:
Hi @spencergibb 
Rebase done.

msamad:
Hi @spencergibb 
Could you please review this one? I really want to get this one in before the end of the year.

msamad:
Closing the pull request as it's been too long and the modules are now in maintenance mode.

PR #2190, opened 2017-08-08
Dynamic filtering on Zuul SimpleRouteLocator.getRoutes()?

tcawley:
I have a need for dynamically filtering out zuul routes when `SimpleRouteLocator/DiscoveryClientRouteLocator` assembles its routes in the `getRoutes()` and `getMatchingRoute()` methods.

The need is security based in that I don't think all routes should be advertised, if, say, a particular user (Principal) doesn't have access to the given routes/paths.  For instance, I want to allow fine-grained route access to users via a custom JWT claim that specifies what routes/paths he/she has access to.  I _could_ prune routes in my application _after_ calling `routeLocator.getRoutes()`, but I'd have to do that in multiple places potentially.  And I'm sure I'd forget completely about such post-filtering after a week or so and end up leaking unauthorized routes elsewhere in my code.  I'd much rather specify a filter at configuration time that `SimpleRouteLocator` can use so that I'm guaranteed that all calls to `routeLocator.getRoutes()` returns only the routes the user is allowed to see.


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -71,12 +71,18 @@ public SimpleRouteLocator(String servletPath, ZuulProperties properties) {
...
-			values.add(getRoute(route, path));
+			if (acceptRoute(route)) {
+				String path = route.getPath();
+				values.add(getRoute(route, path));
+			}
 		}
 		return values;
 	}
 
+	@Override public boolean acceptRoute(final ZuulRoute route) {
```
ryanjbaxter: Formatting

spencergibb:
I'd prefer to not add another level of abstraction here. Can subclassing `SimpleRouteLocator` or having another `RouteLocator` that delegates and filters work?

tcawley:
@spencergibb yeah, I understand, no worries.  Can I propose an even simpler change to `SimpleRouteLocator` that will allow me to subclass it and have the control over the route assembly that I'm seeking?

**FROM THIS:**
```
public List<Route> getRoutes() {
	if (this.routes.get() == null) {
		this.routes.set(locateRoutes());
	}
	List<Route> values = new ArrayList<>();
	for (String url : this.routes.get().keySet()) {
		ZuulRoute route = this.routes.get().get(url);
		String path = route.getPath();
		values.add(getRoute(route, path));
	}
	return values;
}
```
**TO THIS:**
```
public List<Route> getRoutes() {
	if (this.routes.get() == null) {
		this.routes.set(locateRoutes());
	}
	return assembleRoutes(Collections.unmodifiableMap(this.routes.get()));
}

protected List<Route> assembleRoutes(Map<String, ZuulRoute> routes) {
	List<Route> values = new ArrayList<>();
	for (String url : routes.keySet()) {
		ZuulRoute route = routes.get(url);
		String path = route.getPath();
		values.add(getRoute(route, path));
	}
	return values;
}
```

**Sample Subclass:**
```
private static class FilteringRouteLocator extends SimpleRouteLocator {
	public FilteringRouteLocator(String servletPath, ZuulProperties properties) {
		super(servletPath, properties);
	}

	@Override 
	protected List<Route> assembleRoutes(Map<String, ZuulRoute> routes) {

		List<Route> values = new ArrayList<>();
		for (String url : routes.keySet()) {
			ZuulRoute route = routes.get(url);
			if (acceptRoute(route)) {
				String path = route.getPath();
				values.add(getRoute(route, path));
			}
		}
		return values;
	}

	private boolean acceptRoute(ZuulRoute route) {
		return route != null && !(route.getId().equals("foo"));
	}
}

```

tcawley:
Hi Spencer,

I absolutely could, but it isn't the most desirable or efficient solution.  Let me explain.

The big sticking point is that `SimpleRouteLocator` controls the cache of the routes in a privately held `AtomicReference` that subclasses don't have access to.  That's appropriate.  However, if I were to override `SimpleRouteLocator.getRoutes()`, I'd have to call `super.getRoutes()` in my subclass, which will assemble _all_ the `Route`s and then in my subclass I'd have to re-iterate over all the `Route`s and remove the ones that I want to filter.  That's why I say it is inefficient.  

In my PR I'm introducing a callback method, `assembleRoutes()`, that allows subclasses the opportunity to build the route list directly from the cache, without having access to the cache, and iterating over the routes only once.

Did I clarify, or were you thinking of something entirely different?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -68,9 +69,13 @@ public SimpleRouteLocator(String servletPath, ZuulProperties properties) {
 		if (this.routes.get() == null) {
 			this.routes.set(locateRoutes());
 		}
+		return assembleRoutes(Collections.unmodifiableMap(this.routes.get()));
```
Aloren: Maybe it also makes sense to add test that if somebody tries to modify routes inside assembleRoutes it will fail with error? wdyt?

spencergibb:
There's two other protected methods that access `this.routes.get()`. Maybe we make a protected `getRoutesMap()` so we could satisfy overriding the other methods as well, rather than a special purpose `assembleRoutes()` method.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
tcawley: I think that sounds good.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -94,14 +87,19 @@ public Route getMatchingRoute(final String path) {
...
+
 	protected Route getSimpleMatchingRoute(final String path) {
 		if (log.isDebugEnabled()) {
 			log.debug("Finding route for path: " + path);
 		}
 
-		if (this.routes.get() == null) {
-			this.routes.set(locateRoutes());
-		}
+		getRoutesMap();
```
spencergibb: A note here that this is called for the initialization done in getRoutesMap()

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -94,14 +87,19 @@ public Route getMatchingRoute(final String path) {
 
 	}
 
+	protected Map<String, ZuulRoute> getRoutesMap() {
+		if (this.routes.get() == null) {
+			this.routes.set(locateRoutes());
+		}
+		return Collections.unmodifiableMap(this.routes.get());
```
spencergibb: I think the `unmodifiableMap` should stay in `getRoutes()`.

tcawley: Hi Spencer,

I'm not sure what you mean. I introduced the unmodifiableMap in getRoutes() in my prior PR commit that used the assembleRoutes callback. Since that was a no go, I moved it to the getRoutesMap() method in this PR commit so that subclasses won't be able to modify the ZuulRoutes map.

spencergibb: I don't think the protected methods need the map to be unmodifiable.

Issue #892, opened 2016-03-09
Zuul post retry on 503


yukatan:
We have a service that returns 503 responses when some dependencies are not available. We are observing that in this situation zuul retry one time.

The problem is that the request is a POST with json body and the first time the body is correctly sent, but when zuul retries, the body is sent empty and it results in a bad request to the client. We are in Angel.SR3 but the response is the same with SR6

```
{
  "result": {
    "code": 400,
    "info": "Bad Request"
  },
  "errors": [
    {
      "code": "requestBodyInvalid",
      "description": "request body is invalid"
    }
  ]
}
```

Any idea of how to retry with body?? it seems a serius bug to me

Thanks in advance!!

spencergibb:
Can you show more of your configuration please? Retries are the responsibility of ribbon in zuul. By default it only retries GET.

yukatan:
Hi,

After some test I push a demo with the bug running here [https://github.com/yukatan/zuul-503-retry-bug](url)
Steps to reproduce
- Run Eureka
- Run Zuul
- Launch Service503Test testRepeatUntilBadRequest() [https://github.com/yukatan/zuul-503-retry-bug/blob/master/service/src/test/java/org/yukatan/service/Service503Test.java](url)
- Look the test log until you get something similar to this:


> 2016-03-09 22:31:13.755  INFO 14706 --- [nio-8080-exec-1] org.yukatan.service.Application503       : Body arrived to the service: {"content":"hello"}

-

> 2016-03-09 22:31:13.817  INFO 14706 --- [           main] org.yukatan.service.Service503Test       : Response:<400 Bad Request,{"timestamp":1457559073768,"status":400,"error":"Bad Request","exception":"org.springframework.http.converter.HttpMessageNotReadableException","message":"Required request body content is missing: org.springframework.web.method.HandlerMethod$HandlerMethodParameter@113443","path":"/test"},{Server=[Apache-Coyote/1.1], X-Application-Context=[zuul:8000], Date=[Wed, 09 Mar 2016 21:31:13 GMT], Content-Type=[application/json;charset=UTF-8], Transfer-Encoding=[chunked], Connection=[close]}

-

> 2016-03-09 22:31:18.825  INFO 14706 --- [nio-8080-exec-3] org.yukatan.service.Application503       : Body arrived to the service: {"content":"hello"}

-

[...]

PR #2207, opened 2017-08-17
gh-892 fix zuul POST retry

doernbrackandre:


doernbrackandre:
I tried to implement the ResettableServletInputStreamWrapper into the FormBodyWrapperFilter and faced some issues so I've moved it to an own filter. Maybe it is still possible to merge both filters..

doernbrackandre:
@ryanjbaxter sure. I've simplified the fix to one change in the RibbonCommandContext like mentioned above. 

Gimme a moment to revert all changes and re-implement the fix :)

doernbrackandre:
@ryanjbaxter the ContextAwareRequest depends on the RibbonCommandContext which is part of the zuul.filter.route package. Is there a reason why the context is not moved to a more generic package so that the context aware http clients no not depend on zuul?

doernbrackandre:
@ryanjbaxter sorry. my IDE did strange thinks. I really hope you will get through this

ryanjbaxter:
@doernbrackandre np, i understand.  If you can fix the formatting we will be happy to take a look at it.

PR #2209, opened 2017-08-17
Fix zuul post retry

doernbrackandre:
fixes gh-892

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -2619,6 +2619,9 @@ certain Ribbon properties.  The properties you can use are
 `client.ribbon.OkToRetryOnAllOperations`. See the https://github.com/Netflix/ribbon/wiki/Getting-Started#the-properties-file-sample-clientproperties[Ribbon documentation]
 for a description of what there properties do.
 
+WARNING: Enabling `client.ribbon.OkToRetryOnAllOperations` includes retring POST requests wich can have a huge impact
+on the server's resources.
+
 In addition you may want to retry requests when certain status codes are returned in the
 response.  You can list the response codes you would like the Ribbon client to retry using the
  property `clientName.ribbon.retryableStatusCodes`.  For example
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -2619,8 +2619,8 @@ certain Ribbon properties.  The properties you can use are
 `client.ribbon.OkToRetryOnAllOperations`. See the https://github.com/Netflix/ribbon/wiki/Getting-Started#the-properties-file-sample-clientproperties[Ribbon documentation]
 for a description of what there properties do.
 
-WARNING: Enabling `client.ribbon.OkToRetryOnAllOperations` includes retring POST requests wich can have a huge impact
-on the server's resources.
+WARNING: Enabling `client.ribbon.OkToRetryOnAllOperations` includes retring POST requests wich can have a impact
+on the server's resources due to the buffering of the request's body.
 
 In addition you may want to retry requests when certain status codes are returned in the
 response.  You can list the response codes you would like the Ribbon client to retry using the
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
@@ -48,18 +48,19 @@ public class RibbonCommandContext {
 	 * Kept for backwards compatibility with Spring Cloud Sleuth 1.x versions
 	 */
 	@Deprecated
-	public RibbonCommandContext(String serviceId, String method, String uri,
-								Boolean retryable, MultiValueMap<String, String> headers,
+	public RibbonCommandContext(String serviceId, String method,
+								String uri, Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity) {
 		this(serviceId, method, uri, retryable, headers, params, requestEntity,
-				new ArrayList<RibbonRequestCustomizer>(), null);
+			new ArrayList<RibbonRequestCustomizer>(), null);
 	}
 
 	public RibbonCommandContext(String serviceId, String method, String uri,
 								Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity,
 								List<RibbonRequestCustomizer> requestCustomizers) {
-		this(serviceId, method, uri, retryable, headers, params, requestEntity, requestCustomizers, null);
+		this(serviceId, method, uri, retryable, headers, params, requestEntity,
+			requestCustomizers, null);
 	}
 
 	public RibbonCommandContext(String serviceId, String method, String uri,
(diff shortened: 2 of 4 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
```diff
@@ -16,18 +16,20 @@
 
 package org.springframework.cloud.netflix.zuul.filters.route;
 
+import com.google.common.io.ByteStreams;
```
spencergibb: No Guava please

doernbrackandre:
@ryanjbaxter every time I used the eclipse formatter XML the code was screwed. It was easier for me to close the old PR and open a clean one again. sorry for any inconvenience

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
```diff
@@ -124,9 +126,18 @@ public Boolean getRetryable() {
...
 	}
 
-	public InputStream getRequestEntity() {
-		return requestEntity;
-	}
+    public InputStream getRequestEntity() {
+        try {
+            if (!(requestEntity instanceof ResettableServletInputStreamWrapper)) {
+                requestEntity = new ResettableServletInputStreamWrapper(
+                        ByteStreams.toByteArray(requestEntity));
```
spencergibb: Wondering what this will do to large file upload support.

doernbrackandre:
@spencergibb You're right. POST isn't idempotent. But today there are valid use cases for POST retries. A oauth2 token API, for example. There you have mostly POST requests and they are idempotent by nature. (At least, this is our use case)

In the end it is a business decision, I think. The fact that you can enable POST retries does not mean that you have to use it. Also this behavior is linked to the ribbon.OkToRetryAllOperations property which is false on default.

Best regards,
Andre

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
doernbrackandre: I will have a look for that. Thanks

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
doernbrackandre: I will have a look for that. Thanks

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
doernbrackandre: Can be a problem, true. But again, enabling ribbon.OkToRetryAllOperations should be an active decision

ryanjbaxter:
This seems OK to me since its not enabled by default.

spencergibb:
Yes, since it's opt it, that's ok. Still need to get rid of guava.

doernbrackandre:
@spencergibb, @ryanjbaxter 

I've finally removed guava. 

Also I've tried to replace the byte[] in the input stream wrapper with a BufferedReader() to improve the memory management. The problem I've faced is that the mark() method needs to know for how long the mark is valid. 
In theory we want to mark the beginning of the stream forever. I just don't want to use any magic number to do so. Do you have any ideas? 

Cheers

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
```diff
@@ -129,9 +129,9 @@ public Boolean getRetryable() {
     public InputStream getRequestEntity() {
         try {
             if (!(requestEntity instanceof ResettableServletInputStreamWrapper)) {
-                requestEntity = new ResettableServletInputStreamWrapper(
-                        ByteStreams.toByteArray(requestEntity));
-            }
+					requestEntity = new ResettableServletInputStreamWrapper(
```
ryanjbaxter: Fix the spacing here

ryanjbaxter:
@doernbrackandre I dont really have any suggestions.  Do you see it being a huge performance hit?

Also could you perhaps rerun the CircleCI build, I am unsure if the failure is real.

lkwg82:
we (@doernbrackandre and I) discussed whether the retry and the size buffer could trigger an OOM. The size of the buffer, the number of post retries and concurrent post requests could bloat memory temporarily.

doernbrackandre:
ok, it happened again. The pushed code looks differently from what I see.  No formatter or something is active. I need to fix this

doernbrackandre:
@ryanjbaxter In general, I think we should use the BufferedReader instead of reading the whole request body to a byte array. Maybe we can set the mark validity to 1GB or something or make it configurable with a property but then, there is still the question what a good default value could be

ryanjbaxter:
Yeah I see the problem.  What would happen then if we hit that configurable limit?  The request just fails?

doernbrackandre:
@ryanjbaxter The BufferedReader will throw an exception because the mark is invalid. In this case the request  body will be lost and I think, the retry should be aborted

ryanjbaxter:
Sounds like something we are just going to have to deal with.  I would like to confirm the behavior though and add a warning to the documentation on what will happen.

doernbrackandre:
@ryanjbaxter Is there anything I can do now?

ryanjbaxter:
> In this case the request body will be lost and I think, the retry should be aborted

Can you confirm the above statement is correct and add a warning to the documentation about this scenario?

doernbrackandre:
@ryanjbaxter Sorry I lost track a bit. Do you agree, that we should implement the BufferedReader and a static / configurable POST body limit?
I'll take care for the documentation. No problem.

ryanjbaxter:
To be honest I see the benefits to the BufferedReader approach but I think we should stick to the `byte[]` for now.  If we end up running into issues than we can consider changing it.

doernbrackandre:
@ryanjbaxter I've added a warning to the docs.  Not sure why the test failed. All tests passed on my mashine.

ryanjbaxter:
Thanks.  The tests can be a little flakey sometimes on CircleCI.  Can you just try rerunning the build?

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -2619,6 +2619,9 @@ certain Ribbon properties.  The properties you can use are
 `client.ribbon.OkToRetryOnAllOperations`. See the https://github.com/Netflix/ribbon/wiki/Getting-Started#the-properties-file-sample-clientproperties[Ribbon documentation]
 for a description of what there properties do.
 
+WARNING: Enabling `client.ribbon.OkToRetryOnAllOperations` includes retring POST requests wich can have a huge impact
```
ryanjbaxter: Rather than saying it will have a "huge" impact on server resources, I think it might be more useful to explain why server resources would be effected.

doernbrackandre:
I dont have permissions to rerun a build but I can push a fake commit ;)

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/support/ResettableServletInputStreamWrapper.java
```diff
@@ -0,0 +1,38 @@
+package org.springframework.cloud.netflix.zuul.filters.support;
```
ryanjbaxter: Sorry one more thing, we need a license header here

doernbrackandre: my bad. done :)

doernbrackandre:
@ryanjbaxter Is there anything left I can do? Looking forward to get this merged

PR #2219, opened 2017-08-21
[Turbine] Extract separate class TurbineClustersProvider

Aloren:
Extract separate class TurbineClustersProvider to be able to provide custom solution for retrieving configured clusters

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -687,6 +687,9 @@ turbine:
   appConfig: customers
 ----
 
+If you need to customize which cluster names should be used by Turbine (you don't want to store cluster names in
+`turbine.aggregator.clusterConfig` configuration) provide a bean of type `TurbineClustersProvider`.
+
 The `clusterName` can be customized by a SPEL expression in `turbine.clusterNameExpression` with root an instance of `InstanceInfo`. The default value is `appName`, which means that the Eureka serviceId ends up as the cluster key (i.e. the `InstanceInfo` for customers has an `appName` of "CUSTOMERS").  A different example would be `turbine.clusterNameExpression=aSGName`, which would get the cluster name from the AWS ASG name. Another example:
 
 ----
--- a/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/ConfigurationBasedTurbineClustersProvider.java
+++ b/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/ConfigurationBasedTurbineClustersProvider.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2013-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.turbine;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.util.List;
+
+/**
+ * Provides clusters names for Turbine based on configuration value.
+ *
+ * @author Anastasiia Smirnova
+ */
+public class ConfigurationBasedTurbineClustersProvider implements TurbineClustersProvider {
+
+	private static final Log log = LogFactory.getLog(ConfigurationBasedTurbineClustersProvider.class);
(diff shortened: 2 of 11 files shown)


spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/ConfigurationBasedTurbineClustersProvider.java
```diff
@@ -0,0 +1,28 @@
+package org.springframework.cloud.netflix.turbine;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ConfigurationBasedTurbineClustersProvider implements TurbineClustersProvider {
+
+    private final TurbineAggregatorProperties properties;
```
Aloren: moved this configuration from archaius based DynamicProperty to @ConfigurationProperties class

spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/TurbineLifecycle.java
```diff
@@ -29,11 +30,13 @@
 public class TurbineLifecycle implements SmartLifecycle, Ordered {
 
 	private final InstanceDiscovery instanceDiscovery;
+	private final ClusterMonitorFactory<?> factory;
 
-	private boolean running;
+	private volatile boolean running;
```
Aloren: I think, missing volatile was a bug, since stop and isrunning methods can be called from different threads

spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/TurbineHttpConfiguration.java
```diff
@@ -55,8 +56,24 @@ public TurbineProperties turbineProperties() {
 	}
 
 	@Bean
-	public TurbineLifecycle turbineLifecycle(InstanceDiscovery instanceDiscovery) {
-		return new TurbineLifecycle(instanceDiscovery);
+	public TurbineAggregatorProperties turbineAggregatorProperties() {
+		return new TurbineAggregatorProperties();
+	}
+
+	@Bean
```
Aloren: I would also like to add @ConditionalOnMissingBean for the beans inside this configuration to ba able to override them in user application. @ryanjbaxter what do you think?

ryanjbaxter: Generally a good idea, yes

spencergibb: That's probably OK. You can always replace with the same name.

spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/TurbineLifecycle.java
spencergibb: Possibly, though I doubt you'd see it in practice. I've used `AtomicBoolean` too.

spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/TurbineHttpConfiguration.java
```diff
@@ -55,8 +56,24 @@ public TurbineProperties turbineProperties() {
...
+	public TurbineLifecycle turbineLifecycle(InstanceDiscovery instanceDiscovery,
+											 ClusterMonitorFactory<?> factory) {
+		return new TurbineLifecycle(instanceDiscovery, factory);
+	}
+
+	@Bean
+	public ClusterMonitorFactory springAggregatorFactory(TurbineClustersProvider clustersProvider) {
```
spencergibb: Change name to match return type.

Aloren:
Thanks for the review.
I will add documentation tomorrow.
> maybe an example of providing your own TurbineClustersProvider
I've added `EurekaBasedTurbineClustersProvider` that provides clusters based on services names registered in Eureka. Based on some configuration property I can add it to `TurbineHttpConfiguration`. (It will register clusters with names of all services) If that's not ok, I will provide some simple example in javadoc section of `TurbineClustersProvider`.

Aloren:
Fixed all comments. Please review.

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -678,7 +678,11 @@ eureka:
 ----
 
 
-The configuration key `turbine.appConfig` is a list of eureka serviceIds that turbine will use to lookup instances.  The turbine stream is then used in the Hystrix dashboard using a url that looks like: `http://my.turbine.sever:8080/turbine.stream?cluster=<CLUSTERNAME>` (the cluster parameter can be omitted if the name is "default"). The `cluster` parameter must match an entry in `turbine.aggregator.clusterConfig`. Values returned from eureka are uppercase, thus we expect this example to work if there is an app registered with Eureka called "customers":
+The configuration key `turbine.appConfig` is a list of eureka serviceIds that turbine will use to lookup instances.
```
spencergibb: Can you undo the formatting change here.

Aloren:
rebased and fixed. thanks

PR #2223, opened 2017-08-23
[Turbine] Turbine controller

Aloren:
I think it makes sense to have some sort of endpoint that gives users information about currently monitored clusters. Currently if you have many services with different configured clusters it's very difficult to know which streams are actually working and what are the links to them.
For now I added only cluster name and link to the monitoring stream just to give an idea. WDYT?

### DIFF
--- a/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/ClusterInformation.java
+++ b/spring-cloud-netflix-turbine/src/main/java/org/springframework/cloud/netflix/turbine/ClusterInformation.java
@@ -1,0 +1,44 @@
+package org.springframework.cloud.netflix.turbine;
+
+import java.util.Objects;
+
+public class ClusterInformation {
+
+    private final String name;
+    private final String link;
+
+    public ClusterInformation(String name, String link) {
+        this.name = name;
+        this.link = link;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getLink() {
+        return link;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        ClusterInformation that = (ClusterInformation) o;
+        return Objects.equals(name, that.name) &&
+                Objects.equals(link, that.link);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, link);
+    }
+
+    @Override
+    public String toString() {
+        return "ClusterInformation{" +
+                "name='" + name + '\'' +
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
I am questioning why you wouldnt know which clusters are configured if you configured them to begin with.

Aloren:
Let me explain all the features that we are currently in need in Turbine. One of them is having dynamic monitoring dashboard.

If we have ~10 microservices -- services monitoring is not an issue. But as soon as number of microservices grows -- amount of configuration rapidly increases:
1. You need to add cluster configuration for your service to: `turbine.aggregator.clusterConfig`
2. You need to add service configuration to: `turbine.appConfig`

Currently we have Turbine dashboard static page, that stores all monitoring links, but this is not convenient, because each time we add new service with hystrix stream we need to manually add valid monitoring link to it. It means when adding new microservice to Turbine we also have one more additional step: add configuration for Turbine dashboard.

By this pull request I would like to omit the last step in our configuration: retrieving configured streams from Turbine via REST API gives us possibility to remove static configuration from Turbine Dashboard.

Aloren:
Ok, thanks!. I will add tests for the feature in a couple of days then.

spencergibb:
Would it make more sense for this to be an actuator endpoint rather than a bare controller?

Aloren:
Hm, I'm not sure. If in future UI will be added for this information (smth like Eureka dashboard), then it is better not to stick to management port, I think. If you insist I will change to actuator endpoint, currently it does not matter.

spencergibb:
I won't support adding a turbine UI here, but that's beside the point (netflix no longer supports turbine). `/turbine.stream` isn't an actuator, so I'm fine with it not being actuator.

ryanjbaxter:
@Aloren wondering if you could provide some clarification for me, as I cannot seem to get this to work.  The `RestController` is not being created for me when I have Turbine on the classpath, and there is no `TurbineInformationService` bean created in any configuration.  So does the user need to do this all themselves in order to get this to work?  At very least we need some docs around this.

Aloren:
@ryanjbaxter i haven't seen this pr was merged. it doesn't has any tests or configurations for setting up any beans, so i can not guarantee that it works. do you want me to create another PR with cleanup for this one?

ryanjbaxter:
@Aloren Yes please, that would be great!

Aloren:
@ryanjbaxter is there any deadline for the changes? :) so i could plan my time )

ryanjbaxter:
@Aloren I just submitted a PR for this, see https://github.com/spring-cloud/spring-cloud-netflix/pull/2966

Issue #1957, opened 2017-05-17
Ribbon and tests with context


mariuszluciow:
Hi,

I noticed unexpected behaviour while testing Ribbon. I have two tests with separate, different contexts. During the tests Ribbon always uses properties from context that started first, even when test is run from the second context. 

I have example project with test showing this behaviour: https://github.com/mariuszluciow/ribbon-spring-issue

In more detail, first test:
```
@RunWith(SpringRunner.class)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, properties = "client.ribbon.listOfServers=http://localhost:123")
public class ADemoApplicationTests {
    @Test
    public void contextLoads() {
    }
}
```

and the second one:
```
@RunWith(SpringRunner.class)
@EnableFeignClients("com.example")
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
        properties = {"server.port=8083", "client.ribbon.listOfServers=http://localhost:${server.port}"})
public class DemoApplication2Tests {

    @Autowired
    Client client;
    @Test
    public void contextLoads() {
        client.test();
    }

    @FeignClient("client")
    public interface Client {
        @RequestMapping(value = "/test", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
        String test();
    }
}
```

When started, Ribbon logs 
```
DynamicServerListLoadBalancer for client client initialized: DynamicServerListLoadBalancer:{NFLoadBalancer:name=client,current list of Servers=[localhost:123],
...
```
while it should initialise list of servers with localhost:8083.

spencergibb:
Use `@DirtiesContext`

mariuszluciow:
Hi @spencergibb, thanks for the answer. `@DirtiesContext` helps, but the question is if this is the right solution, as the first test doesn't suggest any context *polluting* .

spencergibb:
We use it fairly regularly.

spencergibb:
Ribbon has some oddities

machowski:
@DirtiesContext is often considered to be bad practice 
- it makes your test suite run way slower
- and means that you somehow integrate inside your code under tests, instead of just testing it's behaviour

There are cases when it's useful, but shouldn't be considered as solution to problem reported by mariuszluciow

spencergibb:
Suggestions welcome. It's what we use in spring cloud.

[...]

PR #2226, opened 2017-08-24
Changes IClientConfig to be Environment based rather than Archaius based

bijukunjummen:
This is a POC, I will clean this PR up if it sounds like the right approach to take.

This PR changes the `IClientConfig` implementation to be based on Spring Environment rather than the default archaius based `DefaultClientConfigImpl`. A big benefit of doing this is that some of global state that Archaius maintains is no longer required, as a side benefit it fixes issues like #1957. All the tests appear to be passing, but I am sure I am missing something. Can you please review @spencergibb, @ryanjbaxter and see if this sounds directionally okay.


spencergibb:
It's an interesting idea. Not sure how I feel about it.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/EnvBasedClientConfig.java
```diff
@@ -0,0 +1,100 @@
...
+	public void loadProperties(String restClientName) {
+		setClientName(restClientName);
+		loadDefaultValues();
+	}
+
+	@Override
+	public void loadDefaultValues() {
+		super.loadDefaultValues();
+		putDefaultStringProperty(DeploymentContextBasedVipAddresses, this.getClientName());
+		putDefaultStringProperty(RETRYABLE_STATUS_CODES, "");
```
spencergibb: What this default?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/EnvBasedClientConfig.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	@Override
+	protected void putDefaultIntegerProperty(IClientConfigKey propName,
```
spencergibb: These all seem weird to me.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/EnvBasedClientConfig.java
bijukunjummen: The implementation is yet to be cleaned up @spencergibb :-), right now it extends `DefaultClientConfigImpl` and tries to let it do as much as possible - `loadDefaultValues` comes from `DefaultClientConfigImpl` and loads up all the default values for a ribbon specific client name - essentially it tries to fix up the precedence, anything with `clientname.ribbon.*` has a higher precedence than `ribbon.*` which has higher precedence than defaults built into the code.

If approach looks okay still, I can directly implement the interface `IClientConfig` instead and clean up this implementation.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/EnvBasedClientConfig.java
bijukunjummen: Yes, agreed, these are because I am extending my `IClientConfig` from `DefaultClientConfigImpl`, these specific overrides bypass some of the archaius calls in the base class. However if the approach itself looks okay I can try a clean room implementation of IClientConfig that does not depend on DefaultCleintConfiImpl at all

ryanjbaxter:
You mention 

> A big benefit of doing this is that some of global state that Archaius maintains is no longer required

What exactly does that mean?

bijukunjummen:
Meant these type of calls @ryanjbaxter in DefaultClientConfigImpl  - ```ConfigurationManager.getConfigInstance().getString(..)```  here the archaius configuration being returned by ```ConfigurationManager.getConfigInstance()``` is a singleton.

spencergibb:
One of the things we are thinking about doing is removing the need for Archaius 1 altogether in Finchley.

bijukunjummen:
Oh nice!, but transitively it will be still be required right - I thought the netflix core libs have a deep dependency on Archaius. Or is the thought to migrate to Archaius 2.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfigDefaults.java
```diff
@@ -0,0 +1,79 @@
+package org.springframework.cloud.netflix.ribbon;
+
+import java.util.concurrent.TimeUnit;
+
+public abstract class RibbonClientConfigDefaults {
```
ryanjbaxter: Are these defaults taken from the underlying Ribbon library?  What if they change, how will we know?

bijukunjummen: Yes, they are taken from `DefaultClientConfigImpl` @ryanjbaxter - that is a good point, don't see a good way other than referring back to `DefaultClientConfigImpl`. I was trying to remove all references to this class in the codebase and decided to replicate even the defaults. But may be we can shadow the defaults in `RibbonClientConfigDefaults` by pointing them to `DefaultClientConfigImpl`, this way the only reference to this class remains in `RibbonClientConfigDefaults`. Still a POC though, I still feel that something is going to break with this approach that is not captured in tests

bijukunjummen: Changed the default to shadow`DefaultClientConfigImpl` now @ryanjbaxter

bijukunjummen:
Any thoughts on whether this PR is viable at all @spencergibb, @ryanjbaxter - I have all the changes in place that I had planned for this PR.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfigDefaults.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.netflix.client.config.DefaultClientConfigImpl;
+
+/**
+ * This is shadowing the defaults defined in {@link DefaultClientConfigImpl}, this way the
+ * rest of the code can refer to this class over {@link DefaultClientConfigImpl} and any
+ * changes to the defaults carry over.
+ */
+
+public class RibbonClientConfigDefaults {
```
ryanjbaxter: There is still the problem that if new properties are added we wont know.  This solves the problem if the values change.

ryanjbaxter:
I think its ok, but I am just worried about changes (additions) to properties.  I suppose that wont be a huge deal as we would then have to use them somewhere anyways and if we are not using them it doesnt really matter.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfigDefaults.java
bijukunjummen: sorry, totally missed out your review comment @ryanjbaxter . Yes, it is true that we will not catch other upstream changes to `DefaultClientConfigImpl` . Considering that we there are other plans to move away from Archaius 1 and too many files touched by this change  I am good to close this PR without merging if you are okay with it @ryanjbaxter

ryanjbaxter: Yeah I think that makes sense

PR #2239, opened 2017-08-31
Remove Lombok from spring-cloud-netflix-core module

gzurowski:
Removes usage of Lombok from module `spring-cloud-netflix-core`. This concludes the removal of Lombok from spring-cloud-netflix. I will follow up with another commit that will also remove all references to Lombok from the documentation.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.netflix.feign;
 
 import java.util.Map;
+import java.util.Objects;
 
 import org.springframework.beans.BeanUtils;
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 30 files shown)


ryanjbaxter:
Awesome thanks!

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
```diff
@@ -275,4 +273,99 @@ public boolean isSingleton() {
...
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(applicationContext, decode404, fallback, fallbackFactory,
+				name, path, type, url);
+	}
+
+	@Override
+	public String toString() {
+		return new StringBuilder("FeignClientFactoryBean{")
```
spencergibb: I learned that Spring has a [ToStringCreator](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/core/style/ToStringCreator.html) that we should enforce in the future.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -17,9 +17,9 @@
 package org.springframework.cloud.netflix.zuul.filters;
 
 import com.netflix.hystrix.HystrixCommandProperties.ExecutionIsolationStrategy;
-import lombok.AllArgsConstructor;
-import lombok.Data;
-import lombok.NoArgsConstructor;
+import org.apache.commons.lang.builder.EqualsBuilder;
```
spencergibb: Arg

Issue #2258, opened 2017-09-07
Zuul Routes not correctly refreshed under heavy traffic


jcarvalho:
In an application using a custom `RefreshableRouteLocator` implementation, publishing a `RoutesRefreshedEvent` during a period of heavy traffic can cause new Routes to not be correctly applied.

This is caused by a race-condition in [ZuulHandlerMapping](https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.3.4.RELEASE/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/web/ZuulHandlerMapping.java#L70-L75), in which the mapping is recorded as `dirty` prior to refreshing the routes.

This means that if a new request comes in prior to the `RefreshableRouteLocator.refresh()` method finishing, the `dirty` flag is seen as `true` [here](https://github.com/spring-cloud/spring-cloud-netflix/blob/v1.3.4.RELEASE/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/web/ZuulHandlerMapping.java#L90-L97), the old routes are retrieved from the `RouteLocator`, and the `dirty` flag is set back to `false`, meaning that the new routes will not be picked-up.

I am able to work around this by manually invoking my `RefreshableRouteLocator.refresh()` method prior to publishing the event, which will ensure that `ZuulHandlerMapping` will already see the new route configuration. This works, but is rather wasteful.

Within Spring Cloud Netflix, this can be fixed by only setting the `dirty` flag after refreshing the Route Locator.

Note: This behavior is only triggered when the `refresh()` method takes longer than the 'next' request to come in, hence this only happening in heavy traffic situations.

ryanjbaxter:
@jcarvalho sine you have already done the analysis of the problem would you be willing to submit a PR with your proposed changes?

jcarvalho:
@ryanjbaxter Yes, I'm just sorting things out regarding the Contribution Agreement. Or is it OK to open the PR first?

ryanjbaxter:
Whatever works for you.  You can submit the PR and then accept the contribution agreement after you sort that out.

spencergibb:
This module has entered [maintenance mode](https://github.com/spring-cloud/spring-cloud-netflix#modules-in-maintenance-mode). This means that the Spring Cloud team will no longer be adding new features to the module. We will fix blocker bugs and security issues, and we will also consider and review small pull requests from the community.

PR #2259, opened 2017-09-07
Reload Zuul routes prior to registering handlers

jcarvalho:
When routes are reloaded, ZuulHandlerMapping needs to register the handlers for the new route's paths. This is accomplished by marking the Handler Mapping as 'dirty', which causes the new paths to be registered on the next HTTP request.

However, prior to this commit, the 'dirty' flag was being set prior to actually refreshing the routes, which meant that any request that came in before the route reload was finished, is going to set the 'dirty' flag back to `false`, without actually seeing the new routes, causing their mappings to be lost. This situation is common when the Route Locator's `refresh` method takes a non-trivial amount of time to complete and the Handler Mapping is receiving high amounts of traffic.

This PR ensures that the 'dirty' flag is only set after the reload is complete, ensuring that any subsequent request will see the new set of routes.

Fixes gh-2258

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/web/ZuulHandlerMapping.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/web/ZuulHandlerMapping.java
@@ -73,10 +73,10 @@ public class ZuulHandlerMapping extends AbstractUrlHandlerMapping {
 	}
 
 	public void setDirty(boolean dirty) {
-		this.dirty = dirty;
 		if (this.routeLocator instanceof RefreshableRouteLocator) {
 			((RefreshableRouteLocator) this.routeLocator).refresh();
 		}
+		this.dirty = dirty;
 	}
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@jcarvalho Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2259)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2259) to manually synchronize the status of this Pull Request.

spencergibb:
Two thoughts, if it's a synchronization issue, maybe an atomic Boolean or some other method for dealing with the race condition.

Also, tests, please

pivotal-issuemaster:
@jcarvalho Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2259)!

Issue #2298, opened 2017-09-26
Separate Zuul Into Its Own Submodule


ryanjbaxter:

spencergibb:
Closed by #2332

ryanjbaxter:
Thanks!

PR #2332, opened 2017-09-29
Move zuul to its own module.

spencergibb:
fixes gh-2298


spencergibb:
@ryanjbaxter, can you give this a post merge review? Since it's 2.0.x we have time to make changes.

ryanjbaxter:
@spencergibb could moving `RibbonCommandContext` to a different package be considered a breaking change?  Looks like Sleuth uses that class.

Other than that this looks fine to me.

spencergibb:
It is a breaking change. RibbonCommandContext has zuul specific references. If we can think of an idea to move it back, I'm ok with it.

Issue #1878, opened 2017-04-21
Allow RibbonLoadBalancedRetryPolicy to participate to the ServerStats connection failure count


nithril:
It appears that `org.springframework.cloud.netflix.ribbon.RibbonLoadBalancedRetryPolicy#registerThrowable`
does not participate to the `ServerStats` connection failure count. Thus the `com.netflix.loadbalancer.BaseLoadBalancer#chooseServer` cannot circuit breaks `Server` with too much failure according to the `niws.loadbalancer.default.connectionFailureCountThreshold` property.

Does it make sense?

Thanks

ryanjbaxter:
Potentially we will have to look at it.

nithril:
I wonder, what is the advantage of using spring retry, over the native ribbon retry?

ryanjbaxter:
There was too much inconsistency between various ways requests were made in Spring Cloud Netflix, so we centralized on Spring Retry across all of them.  For a good summary of the history behind this see #1290.

When I added the feature I did not try using the [AvailabilityFilteringRule](https://github.com/Netflix/ribbon/wiki/Working-with-load-balancers#availabilityfilteringrule) (which I assume you are using).  Have you tried using this rule without using Spring Retry?

nithril:
Without spring retry the `AvailabilityFilteringRule` works as intended. Note this rule is added by default.

> There was too much inconsistency between various ways requests were made in Spring Cloud Netflix

Have you considered putting the retry logic into a custom `LoadBalancerCommand` ?

I have look on how to use `ServerStats` into the spring cloud retry logic, but it does not seems straightforward as ribbon is not aware of the next new `chooseServer` calls.

ryanjbaxter:
I need to spend some time looking at how this works, I am not to familiar with it.  I will post some updates here hopefully tomorrow.

ryanjbaxter:
Are you testing this by proxying requests through Zuul?

nithril:
yes

pway99:
Has there been any progress on this issue?   We are experiencing unnecessary client side exceptions while attempting to gracefully shutdown an service instance.  Since the server stats are only updated after retry exhausts its retry limit it is rare for a circuit breaker to open and common for an unavailable server to be chosen by the LoadBalancer.

ryanjbaxter:
Sorry this fell off my radar.  I will try to carve out some time to look at it.

pway99:
[...]

PR #2334, opened 2017-09-29
Bugfix/ribbon load balanced retry policy update server stats

pway99:
This pr is associated with https://github.com/spring-cloud/spring-cloud-netflix/issues/1878 where the RibbonLoadBalancedRetryPolicy does not update the LoadBalancer ServerStats when registering CircuitTrippingExceptions.

note the RetryableRibbonLoadBalancingHttpClient has its own RetryPolicy which extends the FeignRetryPolicy where the ServiceInstance does not have a reference the Server.  For this reason obtaining a reference to LoadBalancer ServerStats requires looping through the load balancers server list.


spencergibb:
Not sure I like having to loop through all servers. That could be expensive with large lists.

pway99:
@spencergibb I agree with your concern on looping through all of the servers.  If there is interest in accepting this PR I would not mind fixing that.  Thanks!

ryanjbaxter:
We are certainly interested in the PR, so go ahead and make the changes.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
```diff
@@ -105,17 +118,14 @@ public void registerThrowable(LoadBalancedRetryContext context, Throwable throwa
...
-    					&& lbServer.getPort() == serviceInstance.getPort()){
-    				ServerStats serverStats = lbContext.getServerStats(lbServer);
-    				serverStats.incrementSuccessiveConnectionFailureCount();
-    				serverStats.addToFailureCount();    				
-    				logger.debug(lbServer.getHostPort() + " RetryCount: " + context.getRetryCount() 
-    					+ " Successive Failures: " + serverStats.getSuccessiveConnectionFailureCount() 
-    					+ " CirtuitBreakerTripped:" + serverStats.isCircuitBreakerTripped());
-    				break;
-    			}
+    		Server lbServer = ribbonServerListChangeListener.getServer(serviceInstance.getHost() + ":" + serviceInstance.getPort());
```
ryanjbaxter: I would like to see this code encapsulated in its own method

pway99: thanks.  I will make that change

spencergibb:
I'm still super hesitant about this. You're doing all this work because of an issue with Feign. Maybe it would be beneficial to fix the issue with feign rather than introducing all of this complexity into ribbon.

pway99:
Thanks for your concern and attention to this issue.  I will spend some time looking for a solution to the feign issue.

spencergibb:
Can you more precisely describe the issue with feign? Maybe you can avoid some unneeded work.

spencergibb:
I can't find the connection between `RetryableRibbonLoadBalancingHttpClient` and feign.

pway99:
there is quite a bit going in with this issue following is a description of why ribbon is so far the best place for solving this problem.
 
zuul creates a retryable ribbon load balancing client which chooses a server and sets the requests host and port.  Then the request is the captured by the RetryTemplate where it is executed by its doWithRetry method. Since the RetryTemplates RestClient is executing the request and capturing the exceptions and then retrying it is responsible for notifying the LoadBalancer of circuit tripping exceptions.

In this case the original server was chosen by the netflix RetryableRibbonLoadBalancingHttpClient which updates request's host and port.  Since the RetryTemplate does not have a reference to the original Server Instance it is a challenge for it  to update LoadBalancers ServerStats.

to the best of my knowledge a solution not involving the RibbonLoadBalancerRetryPolicy would require a change to the RetryableRibbonLoadBalancingHttpClient so that it allows the RetryTemplate to choose the server.

Feing is only involved to the extent that the RetryableRibbonLoadBalancingHttpClient has its own RetryPolicy which extends the FeignRetryPolicy

spencergibb:
We need to find an alternative way to pass the server through. I don't want to maintain another list of servers outside of what ribbon already does.

pway99:
I agree on not maintaining another list of servers and will look for a better solution.  thanks again we really appreciate your attention to this issue

pway99:
@spencergibb your hesitation for the previous solution was justified since it was not fixing the root cause.  After a more thorough review I found that modifying the RetryableRibbonLoadBalancingHttpClient to use the execute method vs executeWithLoadBalancer allows the RetryTemplate to choose the server and solves the problem. 
There are some new test failures which require investigation.

pway99:
@spencergibb and @ryanjbaxter  I was able to resolve this issue by updating the AbstractRibbonCommand so it uses the request.isRetryable() attribute for deciding between the spring-retry and netflix-loadbalancer libraries for retry and loadbalancing concerns.  I am curious if you have any concerns regarding this approach.   Thanks again for you help and attention to this issue and PR.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicyFactoryTests.java
```diff
@@ -36,6 +36,8 @@
 import org.springframework.cloud.netflix.ribbon.RibbonLoadBalancerClient.RibbonServer;
 
 import java.io.IOException;
+import java.net.SocketException;
```
spencergibb: Accidental import?

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClientTests.java
```diff
@@ -247,7 +247,7 @@ public void testRetrySameServerOnly() throws Exception {
 		doReturn(uriRequest).when(request).toRequest(any(RequestConfig.class));
 		RibbonApacheHttpResponse returnedResponse = client.execute(request, null);
 		verify(delegate, times(2)).execute(any(HttpUriRequest.class));
-		verify(lb, times(0)).chooseServer(eq(serviceName));
+		verify(lb, times(1)).chooseServer(eq(serviceName));
```
spencergibb: I'm a little concerned that the number of times `chooseServer()` is called has increased. We've seen issues where, when the number of servers is small (2), one is always skipped since choose was called multiple times. Can you explain why this changed (and for the 4 below as well)?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -114,8 +114,16 @@ protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
-		RS response = this.client.executeWithLoadBalancer(request, config);
-
+		RS response;
+		// Note: using isRetriable to determine if spring-retry or netflix is are managing loadbalancing concerns.
+		if (request != null && request.isRetriable()) {
```
spencergibb: Can you explain in a bit more detail about why `isRetriable()` says it's spring retry vs not? As far as I know, it's if you are using ribbon `RestClient` or not.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClientTests.java
pway99: Actually the number of servers chosen by the LoadBalancer is the same.  Prior to this change the first server instance was chosen by the LoadBalancerCommand.   Now that the AbstractLoadBalancerAwareClient is invoking the execute method for retriable requests the first server is chosen by the InterceptorRetryPolicy.

This can be verified by resetting the RetryPolicy for both the RetryableRibbonLoadBalancingHttpClient and RetryableOkHttpLoadBalancingClient back to the FeignRetryPolicy.  In that scenario the first request will fail because a server has not been chosen and the host and port on the request are undefined.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicyFactoryTests.java
pway99: looks like a unit test was lost during the merge I will re-add that test

spencergibb: If you add this test back soon, I can merge this.

spencergibb:
@pway99 looks like some stray commits made it in via a rebase, can you fix them? I was about to merge.

pway99:
@spencergibb yes I will look for the stray commits and fix them now.  Thanks again

spencergibb:
NP, looking forward to getting this merged.

spencergibb:
any update @pway99?

pway99:
@spencergibb  still working on it.  sorry its taking a bit of time doing my best

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
ryanjbaxter: I didnt see any clarification on this...

spencergibb: Thanks @ryanjbaxter. @pway99?

pway99:
having issues with git merge and rebase.  Going to create a new clean PR

spencergibb:
Sounds good

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
pway99: @spencergibb and @ryanjbaxter this is a good question.  There are 3 configurations I have found that affect the restclient configuration.  The first is having a RestTemplate defined and the second is the zuul.retryable property.   

1 Having a RetryTemplate defined a RetryableRibbonLoadBalancingHttpClient is used otherwise its a RibbonLoadBalancingHttpClient.
2 The zuul.retryable property is used as a flag in several locations for defining a request as retryable or not. 
3 The ribbon.httpclient.enabled property enables the HttpClientRibbonConfiguration

looking at these three configurations I would say that the ribbon.httpclient.enabled property most likely defines the RestClient as ribbon or not.

overall its a challenging configuration, I will give some more thought to inline comments.

Thanks again

pway99:
@spencergibb and @ryanjbaxter In an effort to get past the merge issues and stray commits I have closed this pr and opened a new PR (https://github.com/spring-cloud/spring-cloud-netflix/pull/2390).  I am currently looking into some unit test failures on new pr.

Please let me know if there any action is required on either of these PRs.  

Thanks again for your assistance,
Pat

PR #2390, opened 2017-10-24
Allowing RibbonLoadbalancedRetryPolicy to update ServerStats for circuit tripping exceptions

pway99:
This pull request is to fix a bug where the LoadBalancedRetryPolicy is not updating LoadBalancer ServerStats for circuit tripping exceptions.  There are three changes required for this fix.

1. Update the AbstractRibbonCommand to only execute with the LoadBalancerCommand when the request is not retryable
2. Update Retryable Http and OK ribbon load balancing clients to use  InterceptingLoadBalancingRetryPolicy
3. Update the RibbonLoadBalancedRetryPolicy so that it will update the LoadBalancer ServerStats status for circuit tripping exceptions

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
@@ -75,6 +75,9 @@ public class RibbonClientConfiguration {
 	@Value("${ribbon.client.name}")
 	private String name = "client";
 
+	@Value("${ribbon.restclient.enabled:false}")
+	public static boolean RIBBON_RESTCLIENT_ENABLED;
+	
 	// TODO: maybe re-instate autowired load balancers: identified by name they could be
 	// associated with ribbon clients
 
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
@@ -75,9 +75,6 @@ public class RibbonClientConfiguration {
 	@Value("${ribbon.client.name}")
 	private String name = "client";
 
-	@Value("${ribbon.restclient.enabled:false}")
-	public static boolean RIBBON_RESTCLIENT_ENABLED;
-	
 	// TODO: maybe re-instate autowired load balancers: identified by name they could be
 	// associated with ribbon clients
 
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
@@ -16,17 +16,24 @@
 
 package org.springframework.cloud.netflix.ribbon;
 
-import com.netflix.client.config.CommonClientConfigKey;
-import com.netflix.client.config.IClientConfig;
-import com.netflix.client.config.IClientConfigKey;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.loadbalancer.LoadBalancedRetryContext;
 import org.springframework.cloud.client.loadbalancer.LoadBalancedRetryPolicy;
 import org.springframework.cloud.client.loadbalancer.ServiceInstanceChooser;
+import org.springframework.cloud.netflix.ribbon.RibbonLoadBalancerClient.RibbonServer;
 import org.springframework.http.HttpMethod;
 import org.springframework.util.StringUtils;
 
-import java.util.ArrayList;
-import java.util.List;
+import com.netflix.client.config.CommonClientConfigKey;
+import com.netflix.client.config.IClientConfig;
+import com.netflix.client.config.IClientConfigKey;
+import com.netflix.loadbalancer.Server;
+import com.netflix.loadbalancer.ServerStats;
 
 /**
  * {@link LoadBalancedRetryPolicy} for Ribbon clients.
@@ -41,6 +48,8 @@ public class RibbonLoadBalancedRetryPolicy implements LoadBalancedRetryPolicy {
 	private RibbonLoadBalancerContext lbContext;
 	private ServiceInstanceChooser loadBalanceChooser;
 	List<Integer> retryableStatusCodes = new ArrayList<>();
+
+	protected final Log logger = LogFactory.getLog(getClass());
 
 	public RibbonLoadBalancedRetryPolicy(String serviceId, RibbonLoadBalancerContext context, ServiceInstanceChooser loadBalanceChooser) {
 		this.serviceId = serviceId;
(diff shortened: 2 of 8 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -114,8 +114,16 @@ protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
-		RS response = this.client.executeWithLoadBalancer(request, config);
-
+		RS response;
+		// Note: using isRetriable to determine if spring-retry or netflix is are managing loadbalancing concerns.
```
pway99: I know @ryanjbaxter  and @spencergibb have an outstanding concern around the 'isRetryable()' flag and in-line comments.
>Can you explain in a bit more detail about why isRetriable() says it's spring retry vs not? As far as I know, it's if you >are using ribbon RestClient or not.

@spencergibb and @ryanjbaxter this is a good question. There are 3 configurations I have found that affect the restclient configuration.

1 Having a RetryTemplate defined determines if a RetryableRibbonLoadBalancingHttpClient is used otherwise its a RibbonLoadBalancingHttpClient.
2 The zuul.retryable property is used as a flag in several locations for defining a request as retryable or not.
3 The ribbon.httpclient.enabled property enables the HttpClientRibbonConfiguration

looking at these three configurations I would say that the ribbon.httpclient.enabled property most likely defines the RestClient as ribbon or not. Is this correct?

overall its a challenging configuration, I will give some more thought to inline comments.

Thanks again

spencergibb: `ribbon.restclient.enabled` is the only way to determine if `RestClient` is used.

pway99: Thanks Spencer.. I am thinking the following single comment might be more appropriate.  would you agree?
> // using isRetriable flag to select either the LBC client or LoadBalancerCommand for executing the request and handling both retry and loadbalancing concerns.

spencergibb: I guess I'm questioning that `isRetriable()` is the right thing to use here. I haven't seen any evidence that it is correct.

pway99: I agree with your concern, both the AbstractLoadBalancerAwareClient implementation and the LoadBalancerCommand should be retriable provided they are configured to do so.  With that understanding using 'isRetriable()' to switch between them may not be valid.  I will take some time to review the code with a focus on the ribbon.restclient.enabled property in search for either solid evidence or a better solution.

spencergibb:
Why did you move away from using `ribbon.restclient.enabled`?

pway99:
There were quite a few test failures with ribbon.restclient.enabled, although in hindsight I was not also checking to ensure the request was retryable which might have been part of the problem.   As I continued working the problem I developed some concern for how the FeignLoadBalancer would behave and decided to move the decision logic to the Retryable clients.  I will give the ribbon.restclient.enabled another review where the AbstractRibbonCommand is also checking the IsRetryable flag and get back to you

pway99:
@spencergibb Just confirmed that the solution using the ribbon.restclient.enabled flag along with checking the request.isRetryable() property works well.  All of the unit tests pass and the HttpClientRibbonRetryIntegrationTests do invoke the client.execute(requst,config) code.
Do you have any concerns for the RetryableFeignLoadBalancer?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonClientConfiguration.java
```diff
@@ -75,6 +75,9 @@
 	@Value("${ribbon.client.name}")
 	private String name = "client";
 
+	@Value("${ribbon.restclient.enabled:false}")
+	public static boolean RIBBON_RESTCLIENT_ENABLED;
```
spencergibb: This seems odd to me that this is a static. Please remove static and just make it a normal field. In fact, it probably needs to live in a different place.

tkvangorder: Agreed, having this static is pretty strange, I guess the thing I struggle with is how to correctly inject this flag down into the command? 

The first version of this pull request, Pat had kind of flipped the problem around and just changed how "executeWithLoadBalancer()" behaves in the two retriable clients. That approach is nice because then we don't have to figure out how to inject that value.

We are totally willing to change it, just trying to figure out the best approach for injecting that flag down into the command.

Would you prefer we do something like: 

```java
public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwareClient<RQ, RS>, RQ extends ClientRequest, RS extends HttpResponse>
		extends HystrixCommand<ClientHttpResponse> implements RibbonCommand {
	protected final LBC client;
	protected RibbonCommandContext context;
	protected ZuulFallbackProvider zuulFallbackProvider;
	protected IClientConfig config;
	protected Boolean ribbonRestclientEnabled = false;
```

And then in the RestClientRibbonCommand: 

```java
public class RestClientRibbonCommand extends AbstractRibbonCommand<RestClient, HttpRequest, HttpResponse> {
	public RestClientRibbonCommand(String commandKey, RestClient client,
			RibbonCommandContext context, ZuulProperties zuulProperties) {
		super(commandKey, client, context, zuulProperties);
		ribbonRestclientEnabled = Boolean.TRUE;
	}
```

spencergibb: I'd prefer something like a method in `AbstractRibbonCommand` like `boolean isExecuteWithLoadBalancer()` that by default returns false. `RestClientRibbonCommand` can then override.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -108,14 +108,18 @@ protected static Setter getSetter(final String commandKey,
...
+	}
 
 	@Override
 	protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
 		RS response;
-		if (!RibbonClientConfiguration.RIBBON_RESTCLIENT_ENABLED
+		if (isExecuteWithLoadBalancer()
```
pway99: this should be !isExecuteWithLoadBalancer()

spencergibb:
Unfortunately you now have a polluted history

pway99:
@spencergibb noticed some conflicks this morning, rebased master which resulted in an extra merge.. ughhh.  please advise on how I can resolve this situation.  Thanks again for all of your assistance.  Best,  Pat

ryanjbaxter:
@pway99 we move what was the old master branch to a branch called 1.4.x and the new master branch is what was 2.0.x.  If you want this code in the next Edgware release, which would make sense to me, you need to submit the PR against the 1.4.x branch.

pway99:
Thanks @ryanjbaxter, I will put together a PR for these changes on the 1.4.x branch.

pway99:
@ryanjbaxter I have created a PR for the 1.4.x branch.  Please let me know if any action is required for this pr.  Thanks again. Pat

ryanjbaxter:
@pway99 we can close this PR since we will merge the changes from the 1.4.x branch forward.

PR #2409, opened 2017-10-30
Regression fix for hystrix stream without eureka registration

daniellavoie:


spencergibb:
Closed with some polish and tests via a805b53d865e2233b423113e97c0d78b1e2d1a96

PR #2417, opened 2017-11-02
Adds feign-* missing managed dependencies

agseijas:
Probably we could import a bom from feign.

### DIFF
--- a/spring-cloud-netflix-dependencies/pom.xml
+++ b/spring-cloud-netflix-dependencies/pom.xml
@@ -369,6 +369,46 @@
 				<version>${feign.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-gson</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-jackson-jaxb</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-jackson</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-java8</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-jaxb</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-jaxrs</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.github.openfeign</groupId>
+				<artifactId>feign-ribbon</artifactId>
+				<version>${feign.version}</version>
+			</dependency>
+			<dependency>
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@agseijas Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2417)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2417) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@agseijas Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2417)!

ryanjbaxter:
Why are these needed?  What problem is this solving?

agseijas:
Hi Ryan, maybe I missed something else that could resolve the problem I noticed but I don't yet know of. Feel free to close the issue. Meanwhile I'll try to explain the situation. This is just to add the other missing feing dependencies (feign-*), because as they are not defined on the bom module (spring-cloud-dependencies) you could be in this situation for example:

Use spring cloud on your own project through importing the spring-cloud-dependencies pom on your dependenciesManagement and then adding the libs that you really need on the dependencies section, without having headaches with versions. One of those libs is feign, adding also the feign-gson lib.
In that case you have to declare manually feign-gson's version, potentially producing a missalignment of the spring-cloud's version and your declared version if you upgrade your spring-cloud-dependencies version. With this, you resolve this situation, because you have the proper aligned version already.

ryanjbaxter:
We would never suggest that you add `spring-cloud-netflix-dependencies` to you POM.  You should be using the "starters" only.  `spring-cloud-netflix-dependencies` are dependencies that we need in Spring Cloud Netflix.

agseijas:
Sorry to bother you again but dependency-importing of the starter (spring-cloud-starter-feign) won't resolve the potential misalignment of versions. I guess you're referring to doing this on the pom.xml (which is what I'm doing):

```
<dependencies>
...
    <dependency>
      <artifactId>spring-cloud-starter-feign</artifactId>
      <version>1.3.5.RELEASE</version>
    </dependency>
...
</dependencies>
```

If this is the case, I would suggest to add these missing dependencies on the starter somehow.
Or are you referring to doing something else ?

ryanjbaxter:
What I am saying is that Spring Cloud Netflix does not need those dependencies to provide the functionality we support with Feign so that is why they are not there.  We are not going to add those dependencies and cause unnecessary bloat to the jars in the situation that some users might need them.  If a user is going to using Spring Cloud Feign and needs those dependencies they are just going to have to make sure the versions align.

spencergibb:
I think the request is for dependency management, not a compile time dependency since we already have the feign version. Boot does this. I think it's a reasonable request.

spencergibb:
Or if it wasn't a request for dependency management, I'd be ok with doing dependency management, not compile.

ryanjbaxter:
@spencergibb in which case they should all be made optional?

spencergibb:
It's part of `<dependencyManagement>` not `<dependencies>`. They won't be included in a jar unless the user adds a `<dependency>`. This change allows them to do
```
<dependency>
 	<groupId>io.github.openfeign</groupId>
 	<artifactId>feign-java8</artifactId>
</dependency>
```
without a version.

ryanjbaxter:
Ah I forgot this was happening in the `dependecyManagement` section.  Makes sense.  Sorry about the confusion.

Issue #2420, opened 2017-11-05
Feign adds body which in turn converts the Request method GET to POST


thirunar:
I have created a FeignClient and it requires dynamic headers to be passed as part of request. So, I have added `@HeaderMap` in the method and pass the header. 

I'm aware the `@HeaderMap` is not part of `spring-cloud` and we have to use `@RequestHeader` to pass the headers. 

Now the question is, if we add a method parameter and lets assume the Request method is `GET`, our Feign client adds it in the body, because of which, the request method is converted to `POST`.

I have written a Test in `FeignClientTests` to verify this, 
A Feign Client
```
	@FeignClient(name = "localapp", configuration = TestClientConfig.class)
	protected interface TestClient {
		@RequestMapping(method = RequestMethod.GET, path = "/helloheadersplaceholdersMap", produces = "application/json")
		Map getHelloHeadersPlaceholdersMap(@HeaderMap Map headers);
	}

```
Test
```
	@Test
	public void testHeaderPlaceholdersMap() {
		ImmutableMap<String, String> map = ImmutableMap
				.of("myPlaceholderHeader", "value");
		Map header = this.testClient.getHelloHeadersPlaceholdersMap(map);
		assertNotNull("header was null", header);
	}

```
Following are the logs, when `@HeaderMap` is used in the client
```
[TestClient#getHelloHeadersPlaceholdersMap] ---> GET http://localapp/helloheadersplaceholdersMap HTTP/1.1
[TestClient#getHelloHeadersPlaceholdersMap] Accept: application/json
[TestClient#getHelloHeadersPlaceholdersMap] Content-Type: application/json;charset=UTF-8
[TestClient#getHelloHeadersPlaceholdersMap] Content-Length: 31
[TestClient#getHelloHeadersPlaceholdersMap] myheader1: myheader1value
[TestClient#getHelloHeadersPlaceholdersMap] myheader2: myheader2value
[TestClient#getHelloHeadersPlaceholdersMap] 
[TestClient#getHelloHeadersPlaceholdersMap] {"myPlaceholderHeader":"value"}
[TestClient#getHelloHeadersPlaceholdersMap] ---> END HTTP (31-byte body)

```
When `@RequestHeader` is used in the client
```
[TestClient#getHelloHeadersPlaceholdersMap] ---> GET http://localapp/helloheadersplaceholdersMap HTTP/1.1
[TestClient#getHelloHeadersPlaceholdersMap] Accept: application/json
[...]

PR #2435, opened 2017-11-10
Fixes gh-2420 | Added to test to verify open feign annotations are ignored

thirunar:
This is reference to the issue #2420

### DIFF
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
@@ -34,9 +34,12 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.collect.ImmutableMap;
+import com.netflix.hystrix.exception.HystrixRuntimeException;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@thirunar Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2435)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2435) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@thirunar Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2435)!

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/valid/FeignClientTests.java
```diff
@@ -34,9 +34,12 @@
 import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.collect.ImmutableMap;
```
spencergibb: We do not allow `com.google` classes.

thirunar: I will update to regular Map. The reason for the test is, I was stuck in a situation where I was using HeaderMap in the FeignClient, which converted the Http method from GET to POST, as it was adding it to the body. Had this test been there, I wouldn't have spent time in researching. Let me know if this helps?

spencergibb:
We document that feign annotations don't work be default. If you think the documentation needs clarifying, great. But we don't need a test to verify this.

PR #2436, opened 2017-11-10
Adds backoff policies when retrying requests

### LINKED ISSUE #2195: Support for retry-backoffs
While Spring Retry integrated into Ribbon provide a retry-mechanism that even works in a load-balanced scenario ("ribbon.MaxAutoRetries", "ribbon.MaxAutoRetriesNextServer" properties), back-offs are not supported (the "ribbon.BackoffTimeout" property does not work)

Would it be possible to implement the backoff-feature? Apparently Netflix-Ribbon provides two backoff strategies out-of-the-box (exponential and bounded). It would be nice to have either (or both) in Spring Cloud Netflix.

ryanjbaxter:
Fixes #2195

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/CachingSpringLoadBalancerFactory.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/CachingSpringLoadBalancerFactory.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.netflix.feign.ribbon;
 
 import java.util.Map;
 
+import org.springframework.cloud.client.loadbalancer.LoadBalancedBackOffPolicyFactory;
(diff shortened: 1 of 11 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/ribbon/CachingSpringLoadBalancerFactory.java
```diff
@@ -32,11 +33,13 @@
  *
  * @author Spencer Gibb
  * @author Dave Syer
+ * @author Ryan Baxter
  */
 public class CachingSpringLoadBalancerFactory {
 
 	private final SpringClientFactory factory;
 	private final LoadBalancedRetryPolicyFactory loadBalancedRetryPolicyFactory;
+	private LoadBalancedBackOffPolicyFactory loadBalancedBackOffPolicyFactory;
```
spencergibb: nit final

Issue #2419, opened 2017-11-03
Slice testing of Zuul along the lines of @WebMvcTest


bijukunjummen:
Just a proposal for a feature -

I am working with a customer who has a normal web-app which has a zuul proxy added in to proxy certain uri paths.

Now to test the Zuul parts of the application, loading up the entire Spring Boot context using `@SpringBootTest` leads to a lot of wiring issues. If in some form only the Zuul slice can be loaded up along the lines of say `@WebMvcTest` in Spring Boot that can really help with such tests, this way only the Zuul Configuration along with the configuration to load up the embedded server can be started up and wired in. 

This can potentially be added for other features in spring-cloud-netflix also.

j2ali:
@bijukunjummen How did you work around the wiring issues. I am experiencing the same thing

spencergibb:
> leads to a lot of wiring issues

I'd like to know what those are.

bijukunjummen:
Not zuul wiring issues @spencergibb, customer also has a bunch of auto-configurations which gets in the way. So I would ideally like to prevent customer configurations and auto-configurations from coming in the way. @j2ali I have got this set of configurations (with auto-configuration turned off) which appear to work well for me, these can potentially be added to the slice test.

```
   @Configuration
    @EnableZuulProxy
    @Import({
            DispatcherServletAutoConfiguration.class,
            ZuulProxyAutoConfiguration.class,
            EmbeddedServletContainerAutoConfiguration.class,
            SimpleDiscoveryClientAutoConfiguration.class,
            UtilAutoConfiguration.class,
            PropertyPlaceholderAutoConfiguration.class,
            RibbonAutoConfiguration.class,
            ArchaiusAutoConfiguration.class
    })
    public static class SpringConfig...
```

spencergibb:
Ah, makes sense.

bijukunjummen:
I have a strawman implementation here - https://github.com/bijukunjummen/spring-cloud-netflix/blob/GH-2419/spring-cloud-netflix-core-test/src/test/java/org/springframework/cloud/netflix/zuul/slice/ZuulProxySliceTest.java, still quite rough though, will clean up and send a PR.

spencergibb:
[...]

PR #2437, opened 2017-11-11
Support slice testing of zuul components

bijukunjummen:
Supports #2419 - Slice testing of Zuul filter and properties. If this looks good I can send another PR to update the documentation.


bijukunjummen:
Note that Discovery based Ribbon `ServerList` is disabled for the slice test, this way there is no dependency on needing any registry for these slice tests and the target zuul endpoint needs to be explicitly specified for tests using `listOfServers` - `serviceId.ribbon.listOfServers=localhost:${wiremock.server.port}`. If we absolutely need to support Discovery for Ribbon an option could be to hook up `SimpleDiscoveryClient` based `ServerList` and use it instead(such an auto-configuration does not exist but can be added if desired)

ryanjbaxter:
Is there some kind advantage to doing it this way as opposed to using spring cloud contract?

bijukunjummen:
Yes, good point @ryanjbaxter.  S-C-Contract  would assume that we have contract's defined for all downstream services that Zuul is fronting, which likely may not be available + typically since Zuul is used for hiding legacy services typically those may be say SOAP based services. 

The approach in the PR is to leave it to the user during test to define a stub behavior for the downstream service (using `@AutoConfigureWireMock`) and then test purely using MockMvc

ryanjbaxter:
OK maybe I misunderstood the use case.  This is not for testing our Zuul implementation in SC Netflix its for testing apps using Zuul, right?

bijukunjummen:
Yes, that is correct @ryanjbaxter - I faced an issue at a customer location where we added Zuul proxy to an existing application and this application uses some shared libraries with its own set of auto-configurations. Now to test the Zuul specific configuration we could exclude all the auto-configurations that the customer has OR test it by activating only Zuul specific auto-configurations. 

These specific auto-configurations which worked for at the customer location are now part of this slice test

ryanjbaxter:
Makes sense, could we get some documentation then on how its used?

bijukunjummen:
Oh yes, sounds good @ryanjbaxter, will update the PR with documentation on how to use this

bijukunjummen:
Added in documentation now @ryanjbaxter

PR #2439, opened 2017-11-13
Add support loadBalancerKey in RibbonRoutingFilter of Zuul for Canary test

yongsungyoon:
**Background**

To support canary testing in Zuul, we need to pass the information called `loadBalancerKey` to Ribbon that could be used in `IRule` to choose a server for special treatment. But it was not easy until Ribbon 2.2.2 because it doesn't pass `loadBalancerKey` in `ClientRequest` to `serverLocator(loadBalancerKey)` in `LoadBalancerCommand`.

From Ribbon 2.2.3 (currently 2.2.4 is the latest), it is possible by Ribbon's [PR](https://github.com/Netflix/ribbon/pull/309) now. 
This PR is about supporting to pass `loadBalancerKey`  from Zuul to Ribbon.

**Description**

This PR doesn't change any behavior of Zuul unless there is `loadBalancerKey` in `RequestContext` of Zuul. 
If `RibbonRoutingFilter` finds `loadBalancerKey` from `RequestContext`, it will pass it into `RibbonCommandContext`. And `ContextAwareRequest` will include this value from `RibbonCommandContext` and it will be passed to `LoadBalancerCommand` finally.

**How to Use**

User can add `loadBalancerKey` into `RequestContrext` in his own pre-filter like below.
```
RequestContext.getCurrentContext().set(LOAD_BALANCER_KEY, "any value you want");
```
Because user can access much information of HTTP request like headers and query params in pre-filter, user can use any information of HTTP request to determine `loadBalancerKey`.

The value for `loadBalancerKey` could be any `Object` and it will be passed when `IRule` is called. User can use this value to determine the address for special routing. 

**Note**

There is another subclass of `AbstractLoadBalancerAwareClient` -  `FeignLoadBalancer`. I think that we can support the same feature for Feign with the similar way. But there is no easy way for user to add his own logic for adding `loadBalancerKey`. So we need to consider more about how to support it in Feign.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
@@ -32,6 +32,7 @@ import java.util.Objects;
 
 /**
  * @author Spencer Gibb
+ * @author Yongsung Yoon
  */
 public class RibbonCommandContext {
 	private final String serviceId;
@@ -43,6 +44,7 @@ public class RibbonCommandContext {
 	private final List<RibbonRequestCustomizer> requestCustomizers;
 	private InputStream requestEntity;
 	private Long contentLength;
+	private Object loadBalancerKey;
 
 	/**
 	 * Kept for backwards compatibility with Spring Cloud Sleuth 1.x versions
@@ -52,7 +54,7 @@ public class RibbonCommandContext {
 								String uri, Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity) {
 		this(serviceId, method, uri, retryable, headers, params, requestEntity,
-			new ArrayList<RibbonRequestCustomizer>(), null);
+			new ArrayList<RibbonRequestCustomizer>(), null, null);
 	}
 
 	public RibbonCommandContext(String serviceId, String method, String uri,
@@ -60,13 +62,23 @@ public class RibbonCommandContext {
 								MultiValueMap<String, String> params, InputStream requestEntity,
 								List<RibbonRequestCustomizer> requestCustomizers) {
 		this(serviceId, method, uri, retryable, headers, params, requestEntity,
-			requestCustomizers, null);
+			requestCustomizers, null, null);
+	}
+
+	@Deprecated
+	public RibbonCommandContext(String serviceId, String method, String uri,
+								Boolean retryable, MultiValueMap<String, String> headers,
+								MultiValueMap<String, String> params, InputStream requestEntity,
+								List<RibbonRequestCustomizer> requestCustomizers, Long contentLength) {
+		this(serviceId, method, uri, retryable, headers, params, requestEntity,
+			requestCustomizers, contentLength, null);
 	}
 
 	public RibbonCommandContext(String serviceId, String method, String uri,
 								Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity,
-								List<RibbonRequestCustomizer> requestCustomizers, Long contentLength) {
+								List<RibbonRequestCustomizer> requestCustomizers, Long contentLength,
+								Object loadBalancerKey) {
 		Assert.notNull(serviceId, "serviceId may not be null");
 		Assert.notNull(method, "method may not be null");
 		Assert.notNull(uri, "uri may not be null");
@@ -82,6 +94,7 @@ public class RibbonCommandContext {
 		this.requestEntity = requestEntity;
 		this.requestCustomizers = requestCustomizers;
 		this.contentLength = contentLength;
+		this.loadBalancerKey = loadBalancerKey;
 	}
 
 	public URI uri() {
@@ -155,6 +168,14 @@ public class RibbonCommandContext {
 		this.contentLength = contentLength;
 	}
 
+	public Object getLoadBalancerKey() {
+		return loadBalancerKey;
+	}
+
+	public void setLoadBalancerKey(Object loadBalancerKey) {
+		this.loadBalancerKey = loadBalancerKey;
+	}
+
 	@Override
 	public boolean equals(Object o) {
 		if (this == o)
@@ -169,13 +190,14 @@ public class RibbonCommandContext {
 			.equals(params, that.params) && Objects
 			.equals(requestEntity, that.requestEntity) && Objects
 			.equals(requestCustomizers, that.requestCustomizers) && Objects
-			.equals(contentLength, that.contentLength);
+			.equals(contentLength, that.contentLength) && Objects
+			.equals(loadBalancerKey, that.loadBalancerKey);
 	}
 
 	@Override
 	public int hashCode() {
 		return Objects.hash(serviceId, method, uri, retryable, headers, params,
-			requestEntity, requestCustomizers, contentLength);
+			requestEntity, requestCustomizers, contentLength, loadBalancerKey);
 	}
 
 	@Override
@@ -190,6 +212,7 @@ public class RibbonCommandContext {
 		sb.append(", requestEntity=").append(requestEntity);
 		sb.append(", requestCustomizers=").append(requestCustomizers);
 		sb.append(", contentLength=").append(contentLength);
+		sb.append(", loadBalancerKey=").append(loadBalancerKey);
 		sb.append('}');
 		return sb.toString();
 	}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
@@ -65,7 +65,6 @@ public class RibbonCommandContext {
 			requestCustomizers, null, null);
 	}
 
-	@Deprecated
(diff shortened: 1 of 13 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
```diff
@@ -43,30 +44,33 @@
...
 	public RibbonCommandContext(String serviceId, String method, String uri,
 								Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity,
 								List<RibbonRequestCustomizer> requestCustomizers) {
 		this(serviceId, method, uri, retryable, headers, params, requestEntity,
-			requestCustomizers, null);
+			requestCustomizers, null, null);
 	}
 
 	public RibbonCommandContext(String serviceId, String method, String uri,
```
ryanjbaxter: To avoid breaking anyone I think we should create a new constructor with the new parameter.  You could also mark the old constructor deprecated if it is not necessary anymore.

yongsungyoon: Okay. I will add a new constructor and some documentation soon.

lowzj:
I have a idea about how to add loadBalancerKey in Feign a year ago, see https://github.com/spring-cloud/spring-cloud-netflix/issues/1272#issuecomment-240812591 .
And I suggest that sc-netflix should have another new class to inherit `AbstractLoadBalancerAwareClient`. It will be convenient for other modules to support loadBalancerKey.
More detailed see https://github.com/spring-cloud/spring-cloud-netflix/issues/1272#issuecomment-303087679 , and the implementation: https://github.com/spring-cloud/spring-cloud-netflix/pull/1556 .

yongsungyoon:
@ryanjbaxter I've updated code and added some documentation

yongsungyoon:
@lowzj Hi. Thanks for your comment.
I've seen your suggest and here is what I think.
Only change what we need in those classes is passing `ClientRequest`'s `loadBalancerKey` to `LoadBalancerCommand`'s `serverLocator`(internally also called loadBalancerKey). And currently only 2 direct subclass of `AbstractLoadBalancerAwareClient` exist at this moment - `AbstractLoadBalancingClient` and `FeignLoadBalancer`. The other one `RestClient` is already deprecated. So, I think that just adding one override method to existing classes is better than adding new sub classes.

Personally I think that those changes (passing `loadBalancerKey` between `ClientRequest` and `LoadBalancerCommand` should be done in `AbstractLoadBAlancerAwareClient` in Ribbon as default someday.  Because two fields - `loadBalancerKey` in `ClientRequest` and `serverLocator` in `LoadBalancerCommand` seems to be defined for the same purpose. If it was passed in `AbstractLoadBAlancerAwareClient` class, we even don't need to override one method.  But at this moment (with ribbon 2.2.4), overriding one method is the safest way to support this feature as I think.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonCommandContext.java
```diff
@@ -52,21 +54,31 @@ public RibbonCommandContext(String serviceId, String method,
...
 								Boolean retryable, MultiValueMap<String, String> headers,
 								MultiValueMap<String, String> params, InputStream requestEntity,
 								List<RibbonRequestCustomizer> requestCustomizers) {
 		this(serviceId, method, uri, retryable, headers, params, requestEntity,
-			requestCustomizers, null);
+			requestCustomizers, null, null);
 	}
 
+	@Deprecated
```
ryanjbaxter: Can I change my mind on this?  Sorry, maybe we should keep this constructor around since the command key is optional based on the use case.  Sorry again.

yongsungyoon: NP.

yongsungyoon:
@ryanjbaxter I've add the integration test for loadBalancerKey support. Also deleted `@deprecated` annotation that you requested to remove.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/test/ApacheHttpClientConfigurationTests.java
```diff
@@ -124,7 +124,7 @@ public void testHttpClientWithFeign() {
 	public void testRibbonLoadBalancingHttpClient() {
 		RibbonCommandContext context = new RibbonCommandContext("foo"," GET", "http://localhost",
 				false, new LinkedMultiValueMap<String, String>(), new LinkedMultiValueMap<String, String>(),
-				null, new ArrayList<RibbonRequestCustomizer>(), 0l);
+				null, new ArrayList<RibbonRequestCustomizer>(), 0l, null);
```
ryanjbaxter: you can change these tests back to using the old constructor now

yongsungyoon:
@ryanjbaxter I've fixed.

zywj:
@yongsungyoon 
Zuul -> WebService -> APIService
The ribbon rule of Zuul can get the `loadbalancekey` to select the canary WebService as your TEST.
BUT, WebService cann't get the `loadbalancerkey` from Zuul to select the canary APIservice.
In other words， the `loadbalancerkey` have not been passed down

zywj:
OK. I find the result. 
https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancerClient.java#L138

zywj:
@yongsungyoon 
The canary test call chain seems not being opened.
Zuul -> WebService -> APIService
The `LoadBalancerKey` isn't passd to the WebService, so it cann't using canary test for selecting APIService.

xiaominger:
> **Background**
> 
> To support canary testing in Zuul, we need to pass the information called `loadBalancerKey` to Ribbon that could be used in `IRule` to choose a server for special treatment. But it was not easy until Ribbon 2.2.2 because it doesn't pass `loadBalancerKey` in `ClientRequest` to `serverLocator(loadBalancerKey)` in `LoadBalancerCommand`.
> 
> From Ribbon 2.2.3 (currently 2.2.4 is the latest), it is possible by Ribbon's [PR](https://github.com/Netflix/ribbon/pull/309) now.
> This PR is about supporting to pass `loadBalancerKey` from Zuul to Ribbon.
> 
> **Description**
> 
> This PR doesn't change any behavior of Zuul unless there is `loadBalancerKey` in `RequestContext` of Zuul.
> If `RibbonRoutingFilter` finds `loadBalancerKey` from `RequestContext`, it will pass it into `RibbonCommandContext`. And `ContextAwareRequest` will include this value from `RibbonCommandContext` and it will be passed to `LoadBalancerCommand` finally.
> 
> **How to Use**
> 
> User can add `loadBalancerKey` into `RequestContrext` in his own pre-filter like below.
> 
> ```
> RequestContext.getCurrentContext().set(LOAD_BALANCER_KEY, "any value you want");
> ```
> 
> Because user can access much information of HTTP request like headers and query params in pre-filter, user can use any information of HTTP request to determine `loadBalancerKey`.
> 
> The value for `loadBalancerKey` could be any `Object` and it will be passed when `IRule` is called. User can use this value to determine the address for special routing.
> 
> **Note**
> 
> There is another subclass of `AbstractLoadBalancerAwareClient` - `FeignLoadBalancer`. I think that we can support the same feature for Feign with the similar way. But there is no easy way for user to add his own logic for adding `loadBalancerKey`. So we need to consider more about how to support it in Feign.



> **Background**
> 
> To support canary testing in Zuul, we need to pass the information called `loadBalancerKey` to Ribbon that could be used in `IRule` to choose a server for special treatment. But it was not easy until Ribbon 2.2.2 because it doesn't pass `loadBalancerKey` in `ClientRequest` to `serverLocator(loadBalancerKey)` in `LoadBalancerCommand`.
> 
> From Ribbon 2.2.3 (currently 2.2.4 is the latest), it is possible by Ribbon's [PR](https://github.com/Netflix/ribbon/pull/309) now.
> This PR is about supporting to pass `loadBalancerKey` from Zuul to Ribbon.
> 
> **Description**
> 
> This PR doesn't change any behavior of Zuul unless there is `loadBalancerKey` in `RequestContext` of Zuul.
> If `RibbonRoutingFilter` finds `loadBalancerKey` from `RequestContext`, it will pass it into `RibbonCommandContext`. And `ContextAwareRequest` will include this value from `RibbonCommandContext` and it will be passed to `LoadBalancerCommand` finally.
> 
> **How to Use**
> 
> User can add `loadBalancerKey` into `RequestContrext` in his own pre-filter like below.
> 
> ```
> RequestContext.getCurrentContext().set(LOAD_BALANCER_KEY, "any value you want");
> ```
> 
> Because user can access much information of HTTP request like headers and query params in pre-filter, user can use any information of HTTP request to determine `loadBalancerKey`.
> 
> The value for `loadBalancerKey` could be any `Object` and it will be passed when `IRule` is called. User can use this value to determine the address for special routing.
> 
> **Note**
> 
> There is another subclass of `AbstractLoadBalancerAwareClient` - `FeignLoadBalancer`. I think that we can support the same feature for Feign with the similar way. But there is no easy way for user to add his own logic for adding `loadBalancerKey`. So we need to consider more about how to support it in Feign.

Thanks,but it doesn't work when I use RetryableRibbonLoadBalancingHttpClient, which pass serviceId instead of  loadBalancerKey

Issue #2421, opened 2017-11-05
Eureka Server takes around 10 minutes to update the service-urls since refreshed


fahimfarookme:
I'm deploying eureka and config server on an on-premise cloud. I follow a config first approach where the defaultZone is configured with config-server in config-repo (because DNS is not an option for us). However this mandates a strict deployment order i.e. config server must be up and running before starting eureka. In our case we do not grantee a deployment order due to some infra policies. So the only option available is to invoke the /refresh endpoint once the config-server is started. Also as I tested, the property `eureka.client.eureka-service-url-poll-interval-seconds` is only applicable for DNS.

- Tried invoking /refresh endpoint of eureka / microservices, but the defaultZone is not updated.
- This is because Spring cloud defines `EurekaClient ` bean in refresh scope, however it injects `EurekaClientConfig` bean which is not in refresh scope.

```
@Bean(destroyMethod = "shutdown")
@ConditionalOnMissingBean(value = EurekaClient.class, search = SearchStrategy.CURRENT)
@org.springframework.cloud.context.config.annotation.RefreshScope
@Lazy
public EurekaClient eurekaClient(ApplicationInfoManager manager, EurekaClientConfig config, EurekaInstanceConfig instance) {
	manager.getInfo(); // force initialization
	return new CloudEurekaClient(manager, config, this.optionalArgs,
			this.context);
}
```

Since the `EurekaClient` is already on refresh scope, I believe it would be beneficial if the `EurekaInstanceConfig ` bean could also be defined in refresh scope so that /refresh could update the service-url list.

ryanjbaxter:
I think this makes sense.  Want to try submitting a PR for it?

fahimfarookme:
@ryanjbaxter absolutely!!! Let me file a PR

fahimfarookme:
@ryanjbaxter Further analysis revealed that even-though the `EurekaClient` bean is not defined in refresh scope, the service-url is getting updated in the Eureka Server Dashboard when I hit the /refresh endpoint, **however it takes around 10 minutes** for it to reflect the new service-urls and that's where I was misled. To summarize my finding;

[...]

PR #2455, opened 2017-11-17
Update Eureka Server's service-urls upon EnvironmentChangeEvent.

fahimfarookme:
Fixes gh-2421

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -35,7 +35,9 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.cloud.client.actuator.HasFeatures;
+import org.springframework.cloud.context.environment.EnvironmentChangeEvent;
 import org.springframework.cloud.netflix.eureka.EurekaConstants;
+import org.springframework.context.ApplicationListener;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
 import org.springframework.context.annotation.Configuration;
@@ -67,6 +69,7 @@ import com.sun.jersey.spi.container.servlet.ServletContainer;
 /**
  * @author Gunnar Hillert
  * @author Biju Kunjummen
+ * @author Fahim Farook
  */
 @Configuration
 @Import(EurekaServerInitializerConfiguration.class)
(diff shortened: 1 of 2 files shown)


fahimfarookme:
@ryanjbaxter I would like to request a rebuild, because the test failure is from `FormZuulServletProxyApplicationTests` and most unlikely due to my check-in. I've run all the tests on my branch however could not observe any test failures. Is it possible to share the surefire outputs?

spencergibb:
I would only update if specific keys are changed

fahimfarookme:
Thanks @ryanjbaxter @spencergibb.

Yeah, that makes sense and hope change to either `availability-zone`,`region` or `service-urls` should trigger the update.

fahimfarookme:
@spencergibb @ryanjbaxter Checked-in and rebased. 

I've avoided `String.matches()` / regex due to readability and performance. Also `changedKeys.contains("eureka.client.region")` is checked first since it's a O(1) operation. Kindly have a look.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
```diff
@@ -169,9 +172,69 @@ public PeerAwareInstanceRegistry peerAwareInstanceRegistry(
...
+			}
+		}
+		
+		/*
+		 * Check whether specific properties have changed.
+		 */
+		protected boolean shouldUpdate(final Set<String> changedKeys) {
+			assert changedKeys != null;
+			
+			if (clientConfig.shouldUseDnsForFetchingServiceUrls()) {
```
spencergibb: Can you explain? Should go as a comment.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
```diff
@@ -169,9 +172,69 @@ public PeerAwareInstanceRegistry peerAwareInstanceRegistry(
...
+				return false;
+			}
+			
+			if (changedKeys.contains("eureka.client.region")) {
+				return true;
+			}
+			
+			for (final String key : changedKeys) {
+				assert key != null;
```
spencergibb: Why throw an exception? Use an if.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
```diff
@@ -169,9 +172,69 @@ public PeerAwareInstanceRegistry peerAwareInstanceRegistry(
...
+				// considered only documented relaxed keys.
+				if (key.startsWith("eureka.client.service-url.") ||
+					key.startsWith("eureka.client.serviceUrl.") ||
+					key.startsWith("eureka.client.availability-zones.") ||
```
spencergibb: In boot 2.0 use only the - case names.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
fahimfarookme: This check is there because if the the `property eureka.client.use-dns-for-fetching-service-urls` is set to true, then no point of updating. I've documented it in the Java doc. Will add as a comment here as well.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
fahimfarookme: Looks like the the current master is still not migrated to boot 2.0? Anyways let me keep only the - case names.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
fahimfarookme: I used an `assert` here only to specify the assumption that property keys cannot be null, hence I'm not doing a null check before calling `key.startsWith()`. Since assertions are disabled in production, hope this should fine?

fahimfarookme:
I've addressed the review comments. Will rebase once you reviewed it.

fahimfarookme:
@spencergibb I've addressed the review comments. Please have a look. Will rebase once you reviewed.

fahimfarookme:
@spencergibb I've addressed the review comments. Please have a look. Will rebase once you reviewed.

fahimfarookme:
@spencergibb @ryanjbaxter Appreciate if you could review and let me know if anything needs to be addressed.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
```diff
@@ -169,9 +172,68 @@ public PeerAwareInstanceRegistry peerAwareInstanceRegistry(
...
+		protected boolean shouldUpdate(final Set<String> changedKeys) {
+			assert changedKeys != null;
+			
+			// if eureka.client.use-dns-for-fetching-service-urls is true, then
+			// service-url will not be fetched from environment.
+			if (clientConfig.shouldUseDnsForFetchingServiceUrls()) {
+				return false;
+			}
+			
+			if (changedKeys.contains("eureka.client.region")) {
```
ryanjbaxter: Just a small nit here, this `if` could be combined in the `for` loop below.  Small optimization more than anything else.

fahimfarookme:
@ryanjbaxter I've rebased it. Kindly let me know if anything is pending from my-side to proceed with merging.

ryanjbaxter:
@fahimfarookme thanks.  The only outstanding thing from my side was my comment https://github.com/spring-cloud/spring-cloud-netflix/pull/2455#discussion_r153247584.  I think this might make sense to do

fahimfarookme:
Thanks @ryanjbaxter for the review. [I did a quick micro-benchmark](https://github.com/fahimfarookme/benchmark-spring-cloud-netflix-2455) with JMH since it looks like the submitted PR could perform better than the [suggestion](https://github.com/spring-cloud/spring-cloud-netflix/pull/2455#discussion_r153247584).

In the benchmark results below:
* BenchmarkRunner.**submitted** - is my PR
* BenchmarkRunner.**suggested** - is what you have suggested. i.e.
   
```
for (final String key : changedKeys) {
	if (key.startsWith("eureka.client.service-url.") || 
		key.startsWith("eureka.client.availability-zones.") || 
		changedKeys.contains("eureka.client.region")) {
		return true;
	}
}
```

* BenchmarkRunner.**possibleSuggestion** - I also tried another version along the lines of what you've suggested. i.e.

```
for (final String key : changedKeys) {
	if (key.startsWith("eureka.client.service-url.") || 
		key.startsWith("eureka.client.availability-zones.") || 
		"eureka.client.region".equals(key)) {
		return true;
	}
}
```

Here's the JMH results.

### 1. Throughput (higher the better)

| Benchmark | Mode | Cnt | Score | Error | Units |
| --- | --- | --- | --- | ---| --- |
| BenchmarkRunner.possibleSuggestion | thrpt | 20 | 5267629.307 | ±  571685.517 | ops/s |
| BenchmarkRunner.submitted | thrpt | 20 | 8492321.724 | ± 2838354.868 | ops/s |
| BenchmarkRunner.suggested | thrpt | 20 | 3019842.359 | ±  796130.173 | ops/s |

### 2. Average time (lower the better)

| Benchmark | Mode | Cnt | Score | Error | Units |
| --- | --- | --- | --- | ---| --- |
| BenchmarkRunner.possibleSuggestion | avgt | 20 | ≈ 10⁻⁷ | | s/op |
| BenchmarkRunner.submitted | avgt | 20 | ≈ 10⁻⁷ | | s/op |
| BenchmarkRunner.suggested | avgt | 20 | ≈ 10⁻⁶ | | s/op |

Based on the [results](https://github.com/fahimfarookme/benchmark-spring-cloud-netflix-2455/blob/master/jmh.log) above the submitted PR is already the optimized version. Your thoughts please...

PR #2466, opened 2017-11-23
StringBuilder instead of + stitching string, used to improve performance

holy12345:
HI：
What I understand is that there is no difference between concatenating strings once with + or StringBuilder because the java compiler optimizes us (converting + to StringBuilder) but if you use + to concatenate strings within a for loop, there will be a loop StringBuilder will be created once. This will take up more JVM resources。
See the compiled code through jad
Use + sign：
    for(int j = 0; j < i; j++)
        {
            String u = args1[j];
            if(u.contains("@"))
                filteredUrls = (new StringBuilder()).append(filteredUrls).append(u.substring(0, u.indexOf("//") + 2)).append(u.substring(u.indexOf("@") + 1, u.length())).append(",").toString();
            else
                filteredUrls = (new StringBuilder()).append(filteredUrls).append(u).append(",").toString();
        }

Use StringBuilder ：
      for(int j = 0; j < i; j++)
        {
            String u = args1[j];
            if(u.contains("@"))
            {
                builder.append(u.substring(0, u.indexOf("//") + 2)).append(u.substring(u.indexOf("@") + 1, u.length())).append(",");
            } else
            {
                filteredUrls = (new StringBuilder()).append(filteredUrls).append(u).append(",").toString();
                builder.append(u).append(",");
            }
        }

So I submitted this PR, hoping to improve the performance of this part of the code ，hope you can pass this PR：）

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
@@ -288,13 +288,14 @@ public class EurekaController {
 	private String scrubBasicAuth(String urlList){
 		String[] urls=urlList.split(",");
 		String filteredUrls="";
+		StringBuilder sb = new StringBuilder();
 		for(String u : urls){
 			if(u.contains("@")){
-				filteredUrls+=u.substring(0,u.indexOf("//")+2)+u.substring(u.indexOf("@")+1,u.length())+",";
+				sb.append(u.substring(0,u.indexOf("//")+2)).append(u.substring(u.indexOf("@")+1,u.length())).append(",");
 			}else{
-				filteredUrls+=u+",";
+				sb.append(u).append(",");
 			}
 		}
-		return filteredUrls.substring(0,filteredUrls.length()-1);
+		return sb.substring(0,sb.length()-1);
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
```diff
@@ -288,13 +288,14 @@ protected void filterReplicas(Map<String, Object> model, StatusInfo statusInfo)
 	private String scrubBasicAuth(String urlList){
 		String[] urls=urlList.split(",");
 		String filteredUrls="";
+		StringBuilder sb = new StringBuilder();
```
ryanjbaxter: Remove the `String` variable and change the `StringBuilder` to `filteredUrls`

ryanjbaxter:
Please learn how to [format code on GitHub](https://help.github.com/articles/creating-and-highlighting-code-blocks/).

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
holy12345: Thank you for your reply, I will modify it according to your suggestion : )

PR #2467, opened 2017-11-25
Improve the document

holy12345:
Improve the documentation for the Eureka chapter

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -29,15 +29,13 @@ When a client registers with Eureka, it provides meta-data about itself
 such as host and port, health indicator URL, home page etc.  Eureka
 receives heartbeat messages from each instance belonging to a service.
 If the heartbeat fails over a configurable timetable, the instance is
-normally removed from the registry, if self-preservation is turned-off.
+normally removed from the registry.
 
 Example eureka client:
 
 [source,java,indent=0]
 ----
-@Configuration
-@ComponentScan
-@EnableAutoConfiguration
+@SpringBootApplication
 @RestController
 public class Application {
 


docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -29,7 +29,7 @@ When a client registers with Eureka, it provides meta-data about itself
 such as host and port, health indicator URL, home page etc.  Eureka
 receives heartbeat messages from each instance belonging to a service.
 If the heartbeat fails over a configurable timetable, the instance is
-normally removed from the registry.
+normally removed from the registry(assuming you turn off Eureka's self-protection feature).
```
fahimfarookme: , if self-preservation is turned-off.

holy12345:
@spencergibb Hello, thanks for your reply. I temporarily restore this paragraph. At the same time I @ Configuration, @ ComponentScan, @ EnableAutoConfiguration note replaced by @ SpinBootApplication. What do you think of this

holy12345:
@spencergibb In addition, there are some places in the document using @ Configuration, @ ComponentScan, @ EnableAutoConfiguration, and in some places using @SpringBootApplication. I want to modify it uniformly into @SpringBootApplication. If you agree, I will be unified modification in this pr, thank you

holy12345:
@fahimfarookme Thank you for your reviewed.If I say so, do you think it is OK?
if self-preservation is turned-off or self-preservation is turned-on and renewed not less the percentage configured for the value (not below than 85% in 15 minutes)

ryanjbaxter:
I think you can replace `@Configuration`, `@ComponentScan`, `@EnableAutoConfiguration`with `@springbootapplication`.

holy12345:
@ryanjbaxter Thank you for your reply, I will modify it according to your suggestion.

PR #2469, opened 2017-11-26
Associated with this PR # 2466

holy12345:
I closed the pr # 2466 and recreated one.
The reason I submit this pr is as follows：
My understanding is that there is no difference between using the "+" concatenated string and StringBuilder (because the java compiler helped us to optimize it). However, if the two are very different within the loop. Using "+" inside the loop concatenates strings results in a new StringBuilder object being created each time, thus wasting unnecessary jvm memory。Use jad to see the compiled code.
Use "+" to concatenate strings：

```
for(int j = 0; j < i; j++)
{
String u = args1[j];
if(u.contains("@"))
filteredUrls = (new StringBuilder()).append(filteredUrls).append(u.substring(0, u.indexOf("//") + 2)).append(u.substring(u.indexOf("@") + 1, u.length())).append(",").toString();
else
filteredUrls = (new StringBuilder()).append(filteredUrls).append(u).append(",").toString();
}
```

Use StringBuilder ：
```
for(int j = 0; j < i; j++)
{
String u = args1[j];
if(u.contains("@"))
{
builder.append(u.substring(0, u.indexOf("//") + 2)).append(u.substring(u.indexOf("@") + 1, u.length())).append(",");
} else
{
filteredUrls = (new StringBuilder()).append(filteredUrls).append(u).append(",").toString();
builder.append(u).append(",");
}
}
```

So I submitted this PR, hoping to improve the performance of this part of the code ，hope you can pass this PR：）

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
@@ -287,12 +287,12 @@ public class EurekaController {
 
 	private String scrubBasicAuth(String urlList){
 		String[] urls=urlList.split(",");
-		String filteredUrls="";
+		StringBuilder filteredUrls = new StringBuilder();
 		for(String u : urls){
 			if(u.contains("@")){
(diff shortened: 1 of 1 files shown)


holy12345:
branch is bad

Issue #2479, opened 2017-11-28
Hystrix default timeout property being ignored


rworsnop:
I just tried to update from Dalston.SR4 to Edgware.RELEASE.

I am setting  the `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds` property but, since the update to Edgware, the timeout is always 2 seconds. This setting comes from `AbstractRibbonCommand.getSetter` (since commit 146513ae).

I can work around the problem by setting the timeout for a specific command key. I.e., `hystrix.command.<HystrixCommandKey>.execution.isolation.thread.timeoutInMilliseconds`.

rworsnop:
Looks like this didn't make it into 1.4.1 or 1.4.2. Any updates on the actual release for this?

ryanjbaxter:
@rworsnop there have been a couple issues opened around this problem.  There was a change that made it into 1.4.2.RELEASE https://github.com/spring-cloud/spring-cloud-netflix/pull/2633/files.  Let us know if this addresses your issue

ldiasrs:
I got the same problem, but workaround that @rworsnop mentioned works fine. Thanks for that ;)

stormluke:
I got the same problem, `hystrix.threadpool.default.*` props is not working, but `hystrix.threadpool.<HystrixThreadPoolKey>.*` works.

PR #2481, opened 2017-11-29
Allow customization of hystrix timeouts in zuul ribbon command.

spencergibb:
fixes gh-2479

I wish there was a better way to do this.


rud:
I think maybe the documentation updates should be on a different branch / PR?

spencergibb:
At this point, with hystrix on maintenance mode, I'm going to close this

PR #2483, opened 2017-11-29
Allowing RibbonLoadbalancedRetryPolicy to update ServerStats for circuit tripping exceptions

pway99:
This pull request is for issue https://github.com/spring-cloud/spring-cloud-netflix/issues/1878 on the 1.4.x branch and is associated with https://github.com/spring-cloud/spring-cloud-netflix/pull/2390

Files Changed:
- AbstractRibbonCommand. Add `boolean isExecuteWithLoadBalancer()` which by default returns false.  Modify `run()` to call `executeWithLoadBalancer(...)` when RibbonCommand ` isExecuteWithLoadBalancer()` returns true and request is not retryable.

- RetryableRibbonLoadBalancingHttpClient. RetryPolicy is changed to InterceptingRetryPolicy

- RetryableOkHttpLoadBalancingClient. RetryPolicy is changed to InterceptingRetryPolicy

- RestClientRibbonCommand. Override `isExecuteWithLoadBalancer()` to return true.

- RibbonLoadBalancedRetryPolicy. Update `registerThrowable(...)` so the LoadBalancer ServerStats status are updated for circuit tripping exceptions

- RibbonLoadBalancedRetryPolicyFactoryTests. Add unit test for circuit tripping exceptions 

- RibbonLoadBalancingHttpClientTests. Update to account for the LoadBalancer being used for the initial server request

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
@@ -29,7 +29,7 @@ import org.springframework.cloud.client.loadbalancer.LoadBalancedRetryPolicy;
 import org.springframework.cloud.client.loadbalancer.LoadBalancedRetryPolicyFactory;
 import org.springframework.cloud.client.loadbalancer.RetryableStatusCodeException;
 import org.springframework.cloud.client.loadbalancer.ServiceInstanceChooser;
-import org.springframework.cloud.netflix.feign.ribbon.FeignRetryPolicy;
+import org.springframework.cloud.client.loadbalancer.InterceptorRetryPolicy;
 import org.springframework.cloud.netflix.ribbon.RibbonLoadBalancerClient;
 import org.springframework.cloud.netflix.ribbon.ServerIntrospector;
 import org.springframework.http.HttpRequest;
@@ -154,7 +154,7 @@ public class RetryableOkHttpLoadBalancingClient extends OkHttpLoadBalancingClien
 		return new RequestSpecificRetryHandler(false, false, RetryHandler.DEFAULT, null);
 	}
 
-	static class RetryPolicy extends FeignRetryPolicy {
+	static class RetryPolicy extends InterceptorRetryPolicy {
 		public RetryPolicy(HttpRequest request, LoadBalancedRetryPolicy policy, ServiceInstanceChooser serviceInstanceChooser, String serviceName) {
 			super(request, policy, serviceInstanceChooser, serviceName);
 		}
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
@@ -32,6 +32,7 @@ import org.springframework.cloud.client.loadbalancer.ServiceInstanceChooser;
 import org.springframework.cloud.client.loadbalancer.InterceptorRetryPolicy;
 import org.springframework.cloud.netflix.ribbon.RibbonLoadBalancerClient;
 import org.springframework.cloud.netflix.ribbon.ServerIntrospector;
+import org.springframework.cloud.netflix.ribbon.support.ContextAwareRequest;
 import org.springframework.http.HttpRequest;
 import org.springframework.retry.RetryCallback;
 import org.springframework.retry.RetryContext;
@@ -87,6 +88,16 @@ public class RetryableOkHttpLoadBalancingClient extends OkHttpLoadBalancingClien
 		this.loadBalancedRetryListenerFactory = loadBalancedRetryListenerFactory;
 	}
 
+	@Override
+	public boolean isClientRetryable(ContextAwareRequest request) {
+		return request!= null && isRequestRetryable(request);
+	}
+	
+	private boolean isRequestRetryable(ContextAwareRequest request) {
+		return request.getContext() == null ? true :
+			BooleanUtils.toBooleanDefaultIfNull(request.getContext().getRetryable(), true);
+	}
+	
 	private OkHttpRibbonResponse executeWithRetry(OkHttpRibbonRequest request, LoadBalancedRetryPolicy retryPolicy,
 												  RetryCallback<OkHttpRibbonResponse, Exception> callback) throws Exception {
 		RetryTemplate retryTemplate = new RetryTemplate();
@@ -96,8 +107,7 @@ public class RetryableOkHttpLoadBalancingClient extends OkHttpLoadBalancingClien
 		if (retryListeners != null && retryListeners.length != 0) {
 			retryTemplate.setListeners(retryListeners);
 		}
-		boolean retryable = request.getContext() == null ? true :
-				BooleanUtils.toBooleanDefaultIfNull(request.getContext().getRetryable(), true);
+		boolean retryable = isRequestRetryable(request);
 		retryTemplate.setRetryPolicy(retryPolicy == null || !retryable ? new NeverRetryPolicy()
 				: new RetryPolicy(request, retryPolicy, this, this.getClientName()));
 		return retryTemplate.execute(callback);
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
@@ -144,13 +144,22 @@ public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwar
 		// @formatter:on
 	}
 
+	public boolean isExecuteWithLoadBalancer() {
+		return false;
+	}
+	
 	@Override
 	protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
-		RS response = this.client.executeWithLoadBalancer(request, config);
-
+		RS response;
+		if (!isExecuteWithLoadBalancer()
+				&& request!= null &&  request.isRetriable()) {
+			response = this.client.execute(request, config);
+		} else {
+			response = this.client.executeWithLoadBalancer(request, config);
+		}
 		context.set("ribbonResponse", response);
 
 		// Explicitly close the HttpResponse if the Hystrix command timed out to
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
@@ -21,6 +21,8 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.cloud.netflix.ribbon.RibbonClientConfiguration;
 import org.springframework.cloud.netflix.ribbon.RibbonHttpResponse;
+import org.springframework.cloud.netflix.ribbon.support.AbstractLoadBalancingClient;
+import org.springframework.cloud.netflix.ribbon.support.ContextAwareRequest;
 import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
 import org.springframework.cloud.netflix.zuul.filters.route.RibbonCommand;
 import org.springframework.cloud.netflix.zuul.filters.route.RibbonCommandContext;
@@ -144,18 +146,17 @@ public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwar
 		// @formatter:on
 	}
 
-	public boolean isExecuteWithLoadBalancer() {
-		return false;
-	}
-	
 	@Override
 	protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
 		RS response;
-		if (!isExecuteWithLoadBalancer()
-				&& request!= null &&  request.isRetriable()) {
+		
+		boolean retryableClient = this.client instanceof AbstractLoadBalancingClient
+				&& ((AbstractLoadBalancingClient)this.client).isClientRetryable((ContextAwareRequest)request);
+		
+		if (retryableClient) {
 			response = this.client.execute(request, config);
 		} else {
 			response = this.client.executeWithLoadBalancer(request, config);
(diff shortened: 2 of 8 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -109,13 +109,22 @@ protected static Setter getSetter(final String commandKey,
...
+	
 	@Override
 	protected ClientHttpResponse run() throws Exception {
 		final RequestContext context = RequestContext.getCurrentContext();
 
 		RQ request = createRequest();
-		RS response = this.client.executeWithLoadBalancer(request, config);
-
+		RS response;
+		if (!isExecuteWithLoadBalancer()
```
ryanjbaxter: I am not understanding why only the `RestClientRibbonCommand` would call `execute` directly, can you explain?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
```diff
@@ -109,13 +109,22 @@ protected static Setter getSetter(final String commandKey,
...
+				&& request!= null &&  request.isRetriable()) {
```
ryanjbaxter: appears to be an extra space after the second `&&`

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
```diff
@@ -134,7 +134,7 @@ public RequestSpecificRetryHandler getRequestSpecificRetryHandler(OkHttpRibbonRe
 		return new RequestSpecificRetryHandler(false, false, RetryHandler.DEFAULT, null);
 	}
 
-	static class RetryPolicy extends FeignRetryPolicy {
+	static class RetryPolicy extends InterceptorRetryPolicy {
```
ryanjbaxter: why this change?

Also this is going to get a littler murky in the 2.0.x branch where I made some changes to this due to moving ribbon to its own module

ryanjbaxter:
Also please merge in the latest changes

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
spencergibb: I think this was weird since it wasn't specific to feign.

ryanjbaxter: It was weird, but I did it to reuse code instead of copying it everywhere.  I didnt want to rename the class either because it was already out in the wild.  I didnt like it to begin with.  Since I was reusing the code from Feign I am nervous that making this change would cause some issues.  But my memory could be completely wrong as well...

pway99: The InterceptorRetryPolicy is used for this fix because it will choose and set a 'RibbonServer' instance on the LoadBalancerContext if its ServiceInstance is null.  This allows the RibbonLoadBalancedRetryPolicy to use the loadBalancerContext.serviceInstance for updating ServerStats on circuit tripping exceptions.

prior to this fix the LoadBalancerCommand would choose and set the ServiceInstance; however, that instance is not a RibbonServer and it could not be used for obtaining ServerStats from the LoadBalancer.

pway99:
Hello @ryanjbaxter and @spencergibb,
Looks like there are a number of changes to the 1.4.x branch I will work on updating the code and addressing any associated issues.  Thanks again for taking time to review this PR.

ryanjbaxter:
Thanks.  If you can get that done and answer the questions above we might be able to get it in the next release.  

Sorry for the delay in reviewing this!

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/RetryableOkHttpLoadBalancingClient.java
pway99: it appears as though changes to ribbon and or commons  have made the InterceptorRetryPolicy no longer required.

pway99:
Hello @ryanjbaxter. It appears as though updates to the ribbon and or commons projects have reduced the number of changes required for this fix.  I have not yet identified the specific changes which result in the RibbonLoadBalancedRetryPolicy having access to a RibbonServer from the LoadBalancedRetryContext on the first retry.  Do you have any suggestions on where I should look?  Thanks again for making time to review this issue.

ryanjbaxter:
@pway99 are you talking about here https://github.com/spring-cloud/spring-cloud-netflix/pull/2483/files#diff-c20d68364dc403a6dc6e9551f84a08c8R134?

pway99:
@ryanjbaxter yes that is what I am referring to.  After updating this branch with the latest 1.4.x changes a number of previously passing unit tests started failing.  I was only able to fix them by removing changes from the AbstractRibbonCommand so every request is executed with the load balancer.  Then since the LoadBalancerCommand is choosing the first service instance the InterceptorRetryPolicy is no longer required

ryanjbaxter:
Was it these changes that made the difference?

pway99:
It does not look as if that commit is responsible. 
I just discovered what appears to be a bug in the RetryableRibbonLoadBalancingHttpClient which was causing most of the tests to fail.  (there are 2 dififerent failures which I am now investigating)

If you checkout: 
https://github.com/spring-cloud/spring-cloud-netflix/pull/2483/commits/508142271325c7f4f750d8f90a7fe26f76b06ae6 

Then modify RetryableRibbonLoadBalancingHttpClient line 133 so that getSecureRequest call updates the newRequest vs the original
     '''newRequest = getSecureRequest(newRequest, configOverride);'''

https://github.com/spring-cloud/spring-cloud-netflix/blob/a399cca7ce2b253e6a5d681d54b88e849e1bdc3d/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java#L133

ryanjbaxter:
Let us know what you find

pway99:
Hello @ryanjbaxter.  here is a quick update of what I have found so far.

I have created a new branch with the original changes and a few additional changes worth investigating.  

This is an interesting change because the request and newRequest reference the same instance; however, if the service is not null and the newRequest is updated with a new URI the host and port of the resulting request will be null.
https://github.com/pway99/spring-cloud-netflix/blob/3cdfadcd99a7a302a640593a0b7f5bea0e28ee4c/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java#L133

This is another interesting fix also associated with updating the request with a new URI.  However in this case the resulting path is more similar to the uri's decodedPath. 
https://github.com/pway99/spring-cloud-netflix/blob/3cdfadcd99a7a302a640593a0b7f5bea0e28ee4c/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClientTests.java#L454
~        

I am now working on a load test these changes to confirm their behavior and compare with the branch associated with this pull request

thanks again for you help

pway99:
Hello @ryanjbaxter and @spencergibb I have given this work another review and have made some modifications to narrow the scope and handle configurations where spring-retry is not on the classpath. 

1.   The AbstractRibbonCommand chooses to execute with or without the loadbalancer based on logic contained within the AbstractLoadBalancingClient implementation. 
2.   The Retryable 'apache' and 'ok' client implementations logic for determining if a request is retryable  is then shared by both the executeWithRetry and isClientRetriable methods.  
3.   The Retryable 'apache' and 'ok' client implementations utilize a InterceptorRetryPolicy which is necessary since the serviceInstance for the first request will be null and must be obtained from the chooser.   

** Note allowing the AbstractRibbonCommand execute requests with out the loadbalancer when the client is retryable and implements an InterceptorRetryPolicy provides the RibbonLoadBalancedRetryPolicy with a RibbonServer which can be used for updating server stats.

spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClientTests.java
```diff
@@ -451,7 +451,7 @@ public void testDoubleEncodingWithRetry() throws Exception {
 			public boolean matches(Object argument) {
 				if(argument instanceof HttpUriRequest) {
 					HttpUriRequest arg = (HttpUriRequest)argument;
-					return arg.getURI().equals(uri);
+					return arg.getURI().getPath().equals(uri.getPath());
```
pway99: I am concerned for this modification and am interested if you have any suggestions on why it is necessary.  Note that it is caused by the ribbon client updating the request with a new URI

ryanjbaxter: Actually I think I messed up the URI in this test, see https://github.com/spring-cloud/spring-cloud-netflix/pull/2669

pway99: Thanks!  I will look forward to updating after #2669 is merged

ryanjbaxter:
This looks pretty good to me.

So for a request to be retryable it has to have Spring Retry on the classpath and the request needs to be marked retryable, right?  Have you done testing to figure out in what cases the request is not considered retryable?  I know by default on GET requests are considered retryable and when using Zuul you can make a route as not retryable, is there anything else I am missing?

pway99:
Yes you are correct that a request is retryable when Spring Retry is on the class path and the Zuul route is retriable.  I think you are missing the case where Spring Retry is not on the classpath and the request running through a retryable Zuul route.  I have tested configurations with and without Spring Retry and where the routes retryable flag is true or false.  

I am glad you have mentioned the request type which is affected by the 'ribbon.OkToRetryOnAllOperations' flag.  I just tested  a POST request where Spring Retry is on the classpath, the zuul route is retryable and the OkToRetryAllOperations flag is false.  The code changes in this PR will result in the request being handled by the RetryableRibbonLoadBalancingHttpClient which will in fact be retryable for circuit related exceptions.  This is an interesting case because the OkToRetryAllOperations flag within the RibbonLoadBalancerContext instance is false.  Should the AbstractRibbonCommand executing these request types request with the loadbalancer command?  I am curious if you have any insight regarding this use case and will spend a bit more time investigating.
Thanks again for your assistance this is certainly a trick fix.

pway99:
> I assume it wouldnt retry in this case sinc Spring Retry is not on the classpath, correct?
I do not think that is correct for requests running through a retryable zuul route.  When Spring Retry is not on the classpath the request will be managed by the zuul LoadBalancerCommand which will also retry requests per its retryPolicy.  

https://github.com/Netflix/ribbon/blob/f8bb8ea3e651ca6c368a1a20e219a4b1209528ec/ribbon-loadbalancer/src/main/java/com/netflix/loadbalancer/reactive/LoadBalancerCommand.java#L312
https://github.com/Netflix/ribbon/blob/f8bb8ea3e651ca6c368a1a20e219a4b1209528ec/ribbon-loadbalancer/src/main/java/com/netflix/loadbalancer/reactive/LoadBalancerCommand.java#L232

> IMO if Spring Retry is on the classpath, ribbon.OkToRetryOnAllOperations is false and zuul.retryable is true I would expect the request not be retried. I would think all three requirements would need to be true in order for it to be retried.

I agree with your opinion, and after a bit more review I have found an error with my concern that these request types would be retried. 
   - The InterceptorRetryPolicy will check its policy which is a RibbonLoadBalancedRetryPolicy
       https://github.com/spring-cloud/spring-cloud-commons/blob/8dcbd35e98ae0e7ae937cbad287cfde453579ecc/spring-cloud-commons/src/main/java/org/springframework/cloud/client/loadbalancer/InterceptorRetryPolicy.java#L56
   - Then the RibbonLoadBalancedRetryPolicy will then check for the request type and OkToRetryAllOperations
       https://github.com/spring-cloud/spring-cloud-netflix/blob/84f56ec3b205c617d6c9c2b56046bc054c0eb477/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java#L88
      https://github.com/spring-cloud/spring-cloud-netflix/blob/84f56ec3b205c617d6c9c2b56046bc054c0eb477/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java#L75

ryanjbaxter:
> I do not think that is correct for requests running through a retryable zuul route. When Spring Retry is not on the classpath the request will be managed by the zuul LoadBalancerCommand which will also retry requests per its retryPolicy.

I am pretty sure the request is going to handled by an implementation of `AbstractLoadBalancingClient`.  And since Spring Retry is not on the classpath, that implementation wont be retryable.

pway99:
Thanks Ryan, you are correct.  I did not notice that the `RequestSpecificRetryHandler` provided by the `AbstractLoadBalancingClient` will always return false from its `isRetriableException(Throwable e, boolean sameServer)` method. That will prevent the `LoadBalancerCommand` from retrying the request.

https://github.com/spring-cloud/spring-cloud-netflix/blob/dd978169dac1626e7a7829b5c317a4cecc63ab3a/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonLoadBalancingHttpClient.java#L95
https://github.com/Netflix/ribbon/blob/f8bb8ea3e651ca6c368a1a20e219a4b1209528ec/ribbon-core/src/main/java/com/netflix/client/RequestSpecificRetryHandler.java#L67

ryanjbaxter:
Cool, glad we are on the same page.  So where does the PR stand, is there more work that needs to be done?

pway99:
I not aware of any additional code changes required for this PR.  
That said there are two minor 'code-style' issues which could be updated.  One being the static logger variable in the `RibbonLoadBalancedRetryPolicy` is not upper case and the other is the `isClientRetryable` method on the `AbstractLoadBalancingClient` is on a single line.

ryanjbaxter:
Sure go ahead and make those changes, we always appreciate a little code cleanup :)

pway99:
Sounds good. Thanks!

pway99:
@ryanjbaxter the test failure does not appear to be related to this PR.  Are you able to rebuild?

> Failed to transfer file: http://repo.spring.io/libs-snapshot-local/org/jruby/jruby-core/9.0.0.0/jruby-core-9.0.0.0.jar. Return code is: 409 , ReasonPhrase:Conflict. -> [Help 1]

ryanjbaxter:
All set, thanks for your patience with this!

pway99:
Its been fun, Thanks!

PR #2494, opened 2017-12-01
fix zuul filters package url in Edgware

eacdy:
fix zuul filters package url in Edgware


spencergibb:
Can you put all of the changes in one PR?

spencergibb:
Actuall, this is not for edgware, we merge forward, no need to make a master PR

Issue #2503, opened 2017-12-02
sslHostnameValidationEnabled=false is actually enabling hostname validation


qcastel:
Hello Team

When investigating another issue I raised : https://github.com/spring-cloud/spring-cloud-netflix/issues/2498
I found that the zuul.sslHostnameValidationEnabled option is doing the opposite of what it should do:

sslHostnameValidationEnabled=true will disable ssh validation
sslHostnameValidationEnabled=false will enable ssh validation

https://github.com/spring-cloud/spring-cloud-netflix/blob/4e8d58869e800ab5491635e046b240e3eb235b36/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java#L145

and the prototype is:

```
	@Override
	public HttpClientConnectionManager newConnectionManager(boolean disableSslValidation,
			int maxTotalConnections, int maxConnectionsPerRoute, long timeToLive,
			TimeUnit timeUnit, RegistryBuilder registryBuilder) {
```

I will try to contribute this trivial fix

qcastel:
I fixed it in my environment by overriding DefaultApacheHttpClientConnectionManagerFactory and fixing the way ssl is enabling hostname validation. 

https://github.com/qcastel/forgerockmtlswithzuul/blob/dfa0f833622bff41ebbc9429104c985a2f719c9c/forgerock-zuul/src/main/java/com/forgerock/example/mtls/zuul/ssl/ZuulApacheHttpClientConnectionManagerFactory.java#L90

PR #2504, opened 2017-12-02
Fixes gh-2503

qcastel:
Obvious fix

https://github.com/spring-cloud/spring-cloud-netflix/issues/2503

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -73,6 +73,7 @@ import static org.springframework.cloud.netflix.zuul.filters.support.FilterConst
  * @author Spencer Gibb
  * @author Dave Syer
  * @author Bilal Alp
+ * @author Quentin Castel
  */
 public class SimpleHostRoutingFilter extends ZuulFilter {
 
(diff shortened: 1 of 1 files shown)


qcastel:
@spencergibb Would like to, however your current tests framework doesn't provide a mocking around certificates, etc. 

The current tests around this option are very minimal:

```
	@Test
	public void validationOfSslHostnamesCanBeDisabledViaProperty() {
		addEnvironment(this.context, "zuul.sslHostnameValidationEnabled=false");
		setupContext();
		assertFalse("Hostname verification should be disabled via property",
				getFilter().isSslHostnameValidationEnabled());
	}
```

Also, It's the first time I fork and contribute in this repo. I haven't managed to get it compiled yet. Any guideline available for contributors for making the project compile with intellij?

qcastel:
Please Note that hostname validation is therefore disable by default. And it's even worth as this option is not only disabling the hostname validation but also SSL (see #2505).
Therefore, this should be considered as a security issue, a severe one IMHO.

qcastel:
If you are interested in reproducing this, I created a sample project https://github.com/qcastel/forgerockmtlswithzuul which was originally to investigate on testing mutual TLS with a zuul proxy in front of a micro-service.

ryanjbaxter:
I dont think you need to do to that extent.  I think we just need to vaildate that the correct `SSLContext` is setup in the connection manager.  If you look in `SimpleHostRoutingFilterTests` there are some reflection utilities you might be able to use to introspect the connection manager.

ryanjbaxter:
ping @qcastel

ryanjbaxter:
@qcastel are you interested in updating this pr?

qcastel:
@ryanjbaxter I have the feeling I won't find the time for it. I apologies for not being able to contribute as I would.
If someone wants to take over that issue, don't hesitate to take it.

PR #2514, opened 2017-12-05
Adjust the order of judgment

holy12345:
Hello, submit the public relations background is this:
In the ribbon `sample-client.ribbon.OkToRetryOnAllOperations`, this parameter indicates whether it is possible to retry all operations for this client(I think include the `GET` request).
When a retry occurs, we should really only judge if the parameter is true (because the meaning of this parameter is whether all operations should be retries), maybe I understand the mistake, I hope to get your answer.
So I submitted this PR, thank you, best wishes.


holy12345:
@ryanjbaxter Look forward to your reply, thank you very much.

PR #2518, opened 2017-12-06
Fix #2513

### LINKED ISSUE #2513: SimpleHostRoutingFilter does not respect requestEntity
Checkout getRequestBody() methods of SimpleHostRoutingFilter and RibbonRoutingFilter:

RibbonRoutingRoutingFilter:
```
protected InputStream getRequestBody(HttpServletRequest request) {
    InputStream requestEntity = null;

    try {

        requestEntity = (InputStream) RequestContext.getCurrentContext().get(REQUEST_ENTITY_KEY);

	if (requestEntity == null) {

	    requestEntity = request.getInputStream();

        }
    }
    catch (IOException ex) {

        log.error("Error during getRequestBody", ex);

    }

    return requestEntity;

}
```

And SimpleHostRoutingFilter:

```
private InputStream getRequestBody(HttpServletRequest request) {
    InputStream requestEntity = null;

    try {
        requestEntity = request.getInputStream();
    }
    catch (IOException ex) {
		// no requestBody is ok.
    }
    return requestEntity;
}
```

So, if one tires to implement one of the examples of modifying the request body by setting requestEntity in Zuul context, it will have no effect when using SimpleHostRoutingFilter as it takes body directly from request, ignoring the value set in context. Also, SimpleHostRoutingFilter.getRequestBody is private (opposed to RibbonRoutingFilter.getRequestBody() which is protected), so one can't even override it.

holy12345:
Hello, this PR fixes this problem, modify the point as follows:
1, First get from the `RequestContext`, if not get, we get from the request
2, change the method from `private` to `protected`

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -73,6 +73,7 @@ import static org.springframework.cloud.netflix.zuul.filters.support.FilterConst
  * @author Spencer Gibb
  * @author Dave Syer
  * @author Bilal Alp
+ * @author Gang Li
  */
 public class SimpleHostRoutingFilter extends ZuulFilter {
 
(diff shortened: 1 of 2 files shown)


holy12345:
@ryanjbaxter Hello, may I ask you. In addition to the code for the modified test added, and I added an additional extra with the of some of the test code.  why the report shows the test coverage is lower than before? (For the modified class compared to the previous test coverage has been greatly improved)

ryanjbaxter:
Why are you making these changes?

holy12345:
@ryanjbaxter Hello, thanks for your reply. This PR mainly provides two functions: the first to solve the problem of # 2513, the second increase of some test code. Due to my lack of rigor led to multiple commit (please forgive me, I will pay attention). Can you merge this PR? If you can not merge this PR because I did not do something right, then please tell me that I will amend it in time. best wishes.

spencergibb:
We can squash commits, don't worry.

holy12345:
@spencergibb Ok, thanks for the answer. Then what else do I need to do for this PR? If there is no problem hope you can merge it. best wishes ：)

PR #2547, opened 2017-12-13
Update zuul-core to 1.3.1 version

Aloren:


ryanjbaxter:
@Aloren have you taken a look at the test failures?  They seem like they might be related.

spencergibb:
It's an ordering problem. That test doesn't fail in isolation.

spencergibb:
The tests still need to be fixed, because I replicated it locally.

spencergibb:
Next email :-) #2565

ryanjbaxter:
I reverted the commit.  With M6 looming I would rather wait on this.

@Aloren if you could look at the failing tests and submit it again that would be great.

Aloren:
I ran tests on my local machine and they are green, so I thought that they were not stable from the beginning. I will look into it and try to fix.

ryanjbaxter:
They failed on my machine and on our Jenkins build as well  https://jenkins.spring.io/job/spring-cloud-netflix-master-ci/4650/

PR #2551, opened 2017-12-14
Added util method to upgrade the connection to a secure one.

TYsewyn:
- Adds support for websocket URIs
- Fixes bug where a non-secure scheme other than http would be changed to https
- Deprecated `updateToHttpsIfNeeded` in favor of `updateToSecureConnectionIfNeeded`


ryanjbaxter:
Was this fixing another issue that we already had open?

spencergibb:
It's in concert with the gateway websockets loadbalancing.

PR #2552, opened 2017-12-15
fix spring proxy warning

ileler:
```
[restartedMain] INFO  o.s.aop.framework.CglibAopProxy - Method [com.netflix.client.config.IClientConfig org.springframework.cloud.netflix.feign.ribbon.LoadBalancerFeignClient.getClientConfig(feign.Request$Options,java.lang.String)] is package-visible across different ClassLoaders and cannot get proxied via CGLIB: Declare this method as public or protected if you need to support invocations through the proxy.
```


ryanjbaxter:
Please learn how to [format code on GitHub](https://help.github.com/articles/creating-and-highlighting-code-blocks/).
Can you elaborate on when you are seeing this error?

ileler:
when i startup my project...It's not always the same warning.I haven't found a necessary path at the moment.Just find this warning and want to deal with it

spencergibb:
It's not a warning, it's `INFO`. We'd rather not open these methods just to get rid of a log line.

PR #2554, opened 2017-12-15
Increase spring retry listener function

holy12345:
This PR is related to this function https://github.com/spring-cloud/spring-cloud-commons/pull/293


holy12345:
Hi @ryanjbaxter 
I think the reason for the failure of the `CircleCI `test is that the class `LoadBalancedRetryListenerFactory `is in the `spring-cloud-commons` dependency and is not currently in `spring-cloud-netflix-core`, So I think it does not matter.

Issue #2524, opened 2017-12-07
eureka console,Instance jump address not consistent


An0nymous0:
Spring Cloud Version:2.0 M5
Spring boot Version:Finchley.M7
-------
question:
Instance.url jump address is not [eureka.instance.ip-address] + [server.port]?
-------
![image](https://user-images.githubusercontent.com/6836030/33710190-c44fb682-db7a-11e7-8e83-37ae8727a578.png)
![image](https://user-images.githubusercontent.com/6836030/33749878-6cd5c7fc-dc0b-11e7-89b2-5350a925dc8f.png)

client config
-------
eureka.instance.ip-address=127.0.0.1 

prefer-ip-address= true 

instance-id=${eureka.instance.ip-address}:${spring.application.name}:${server.port}

ryanjbaxter:
Seems like it is behaving the way you configured it.  You could put any ip you wanted in instance-id and that will not necessarily match the URL to the actual service.

An0nymous0:
@ryanjbaxter StatusPageUrl IP and port acquiescence value is what I did not find in the source

ryanjbaxter:
What do you mean you didnt find it in the source?

An0nymous0:
@ryanjbaxter yes,I want to know his sources.

ryanjbaxter:
Are you saying 62a8f72 caused the issue?

The code you pointed to in `EurekaInstanceConfigBean` did not change between Finchley.M4 and Finchley.M5.
https://github.com/spring-cloud/spring-cloud-netflix/blob/v2.0.0.M4/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaInstanceConfigBean.java#L286
https://github.com/spring-cloud/spring-cloud-netflix/blob/v2.0.0.M5/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaInstanceConfigBean.java#L286

An0nymous0:
They changed a lot

https://github.com/spring-cloud/spring-cloud-netflix/blob/v2.0.0.M4/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java

[...]

PR #2557, opened 2017-12-15
Fix incorrect managementPort, statusPageUrl and healthCheckUrl when management.server.port not set

brbsk:
Hello,

this PR is an attempt to fix incorrect urls being reported to Eureka since 2.0.0.M5.

It changes the following condition:
```java
(managementPort == null && isRandom(serverPort))
```
which always evaluated to false due to `serverPort` having fallback to a default, as in:
```java
int serverPort = Integer.valueOf(propertyResolver.getProperty("server.port", propertyResolver.getProperty("port", "8080")));
```
and hence always being != `RANDOM_PORT` (unless one sets the `server.port` or `port` to 0 in props but I don't really see a use case for that and how it could affect the Eureka registration part 🤔).

It also assumes that `healthCheckUrl` and `statusPageUrl` should not be set explicitly when `management.server.port` is not set (based on the note at the end of [this paragraph](https://docs.spring.io/spring-boot/docs/2.0.0.M7/reference/html/production-ready-monitoring.html#production-ready-customizing-management-server-context-path) of Spring Boot docs).

I hope it is useful. Any comments will be appreciated.

BR,
Barbara


pivotal-issuemaster:
@brbsk Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2557)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2557) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@brbsk Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2557)!

ryanjbaxter:
@brbsk is this related to #2524?

brbsk:
I think it might be, however I originally noticed the issue somewhere else.

I am using just the `eureka.instance.nonSecurePort` and `eureka.instance.ipAddress` options, no `management.server.port` option set. I noticed that since M5 the nonSecurePort option doesn't get applied to `managementPort` as it gets set too early (specifically `management.port` key in `metadata` map in `InstanceInfo`; the same occurs for `statusPageUrl` and `healthCheckUrl`).

The issue reveals itself in an Eureka aware Turbine - it loses capability of connecting to an instance registered in such a way. For an M4 instance having same options set (`eureka.instance.nonSecurePort` and `eureka.instance.ipAddress`), Turbine has no issue.

ryanjbaxter:
This seems OK to me @spencergibb what do you think?

brbsk:
Hello @spencergibb does this seem OK to you?

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
```diff
@@ -215,17 +216,16 @@ public void statusPageUrl_and_healthCheckUrl_do_not_contain_server_context_path(
...
 
 		setupContext(RefreshAutoConfiguration.class);
 		EurekaInstanceConfigBean instance = this.context
 				.getBean(EurekaInstanceConfigBean.class);
-		assertTrue("Wrong status page: " + instance.getStatusPageUrl(),
-				instance.getStatusPageUrl().endsWith(":8989/management/info"));
-		assertTrue("Wrong health check: " + instance.getHealthCheckUrl(),
-				instance.getHealthCheckUrl().endsWith(":8989/management/health"));
+
+		assertThat(instance.getStatusPageUrl()).isNull();
```
spencergibb: I'm confused why these aren't set? Just because the port isn't set, doesn't mean status and health check urls don't exist.

brbsk: It's not like they don't exist at all, they only don't get overriden here. They later get built by `InstanceInfo.Builder` upon registration (please see the Builder's setters).

This change only fixes broken communication with Turbine brought by M5, by restoring previous behaviour for the managementPort-not-set case. If other behaviour is needed I think it's a separate issue.

brbsk:
Hi, are we going to proceed with this (accept/decline/?)?

spencergibb:
I still need to be convinced the urls removed is the right thing to do.

brbsk:
Do you mean you need more input from me or you need more time for your own analysis, to get convinced?

brbsk:
Can we count on the issue being fixed in 2.0.0.M6, with this PR or any other?

spencergibb:
Not at the moment.

brbsk:
Is there anything I can do to make this possible?

PR #2559, opened 2017-12-16
Refactor the getScanner method of the FeignClientsRegistrar class

holy12345:
After Spring4.1 version we can use a very simple way to check if a scanned class is an `Annotation`. So I refactored this method to make it look more concise.

Note:
Please refer to the detailed information: https://jira.spring.io/browse/SPR-11711


holy12345:
Hello @ryanjbaxter 
Trouble you check this PR, any one of your answers are a great encouragement to me, thanks again!

PR #2563, opened 2017-12-19
OAuth2 auto configuration support for Eureka Client.

daniellavoie:
Provides support for OAuth2 on Eureka Client.

**Disclaimer**: Support is only provided when Eureka Client is not using Jersey.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -85,14 +85,35 @@ See {github-code}/spring-cloud-netflix-eureka-client/src/main/java/org/springfra
 
 To disable the Eureka Discovery Client you can set `eureka.client.enabled` to `false`.
 
-=== Authenticating with the Eureka Server
+=== Basic authentication with the Eureka Server
 
 HTTP basic authentication will be automatically added to your eureka
 client if one of the `eureka.client.serviceUrl.defaultZone` URLs has
 credentials embedded in it (curl style, like
-`http://user:password@localhost:8761/eureka`). For more complex needs
-you can create a `@Bean` of type `DiscoveryClientOptionalArgs` and
-inject `ClientFilter` instances into it, all of which will be applied
+`http://user:password@localhost:8761/eureka`).
+
+=== OAuth2 client support
+
+OAuth 2 support will be auto configured when `org.springframework.security.oauth:spring-security-oauth2` 
+is available in your classpath and you provide a `@Bean` of type `EurekaOAuth2ResourceDetails` within 
+your context. OAuth2 resource details can by configured with the following properties: 
+
+.application.yml
+----
+ eureka:
+  client:
+    oauth2:
+      client_secret: client-secret
+      client_id: user
+      access_token_uri: oauth2-token-uri
+----
+
+IMPORTANT: OAuth2 client support is only supported with Rest Template. Jersey dependencies must be excluded. See <<EurekaClient without Jersey>> for more information.
+
+=== Custom authentication
+
+For more complex needs you can create a `@Bean` of type `DiscoveryClientOptionalArgs` and
+inject `ClientFilter` or `RestTemplace` instances into it, all of which will be applied
 to the calls from the client to the server.
 
 NOTE: Because of a limitation in Eureka it isn't possible to support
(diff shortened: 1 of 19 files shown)


pom.xml
```diff
@@ -27,6 +27,7 @@
 		<spring-cloud-commons.version>2.0.0.BUILD-SNAPSHOT</spring-cloud-commons.version>
 		<spring-cloud-config.version>2.0.0.BUILD-SNAPSHOT</spring-cloud-config.version>
 		<spring-cloud-stream.version>Elmhurst.BUILD-SNAPSHOT</spring-cloud-stream.version>
+		<spring-security-oauth2.version>2.2.1.RELEASE</spring-security-oauth2.version>
```
spencergibb: This isn't managed by boot?

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -85,14 +85,35 @@ See {github-code}/spring-cloud-netflix-eureka-client/src/main/java/org/springfra
...
+      client_secret: client-secret
+      client_id: user
+      access_token_uri: oauth2-token-uri
+----
+
+IMPORTANT: OAuth2 client support is only supported with Rest Template. Jersey dependencies must be excluded. See <<EurekaClient without Jersey>> for more information.
+
+=== Custom authentication
+
+For more complex needs you can create a `@Bean` of type `DiscoveryClientOptionalArgs` and
```
spencergibb: Mention this is for jersey specifically.

daniellavoie: Well not exactly true, if you use the rest template you can still extend the `DiscoveryClientOptionalArgs` (that's what we do on SCS)

pom.xml
daniellavoie: Still a WIP : https://github.com/spring-projects/spring-security-oauth/issues/1240

docs/src/main/asciidoc/spring-cloud-netflix.adoc
spencergibb: Thanks, that's right

pom.xml
spencergibb: :+1:

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/BasicEurekaRestTemplateFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.http.client.support.BasicAuthorizationInterceptor;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * @author Daniel Lavoie
+ */
+public class BasicEurekaRestTemplateFactory implements EurekaRestTemplateFactory {
+	@Override
+	public RestTemplate newRestTemplate(String serviceUrl) {
+		RestTemplate restTemplate = new RestTemplate();
```
ryanjbaxter: Shouldnt we use `RestTemplateBuilder`?

daniellavoie: I don't have any strong opinion on that. Pushed a new version with the builder.

pom.xml
daniellavoie: I don't know if there are plans to make this managed on boot. @rwinch any opinion on that? I don't need autoconfiguration, just the dependencies. We are talking about boot 2.0.

rwinch: @daniellavoie Thanks for reaching out! There are no plans for the old OAuth project version to be managed by Spring Boot.

daniellavoie: Had a chat with Rob, best option is to migrate `OAuth2RestTemplate` to `WebClient` using Spring Security 5. I'll reword that PR.

daniellavoie: Ok, it won't be as easy. Spring Security 5 doesn't support automatic token retrieval for resource servers. The old `OAuth2RestTemplate` handled that for us. @rwinch plans on integrating that for Security 5.1. We have 3 options:

* Wait for Security 5.1
* Integrate the old non-managed lib until migration to Security 5.1
* Write token retrieval logic with WebClient (could be base work for Security 5.1).

I think the most reasonable option is to wait for Security 5.1

@ryanjbaxter @spencergibb Any opinion?

ryanjbaxter: I dont see any reason why we need this PR right now so I dont see why we cant wait.

PR #2564, opened 2017-12-19
Add check to prevent StackOverflowError

holy12345:
Hi @spencergibb @ryanjbaxter 
The problem solved by this PR is as follows:
When a `fallback `or `fallbackFactory `is defined in `@FeignClient`, defining a type that is an interface can result in recursive creation at spring start-up and resulting in `java.lang.stackoverflowerror`. So I think we should do type checking when loading `@FeignClient` annotation.

Your reply is a great encouragement to me, thanks again.

### DIFF
--- a/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/FeignClientsRegistrarTests.java
+++ b/spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/feign/FeignClientsRegistrarTests.java
@@ -20,13 +20,19 @@ package org.springframework.cloud.netflix.feign;
 import java.util.Collections;
 
 import org.junit.Test;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Configuration;
 import org.springframework.mock.env.MockEnvironment;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
 
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 
 /**
  * @author Spencer Gibb
+ * @author Gang Li
  */
 public class FeignClientsRegistrarTests {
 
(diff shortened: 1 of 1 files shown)


holy12345:
Hi @spencergibb @ryanjbaxter 
First of all thank you for your reply.Allow me to explain
First of all, I found that the current test code has covered the PR I submitted this time, the specific test code is `FeignHttpClientUrlTests` . By the results I saw this PR code has been covered by test.
In addition this `StackOverflowError ` situation is very good to reproduce, let me use a simple example to illustrate:

Note that DemoService is an interface, we can also be replaced by other interfaces
```
@FeignClient(name = "demo", url = "http://localhost:8080", fallback = DemoService.class)
public interface DemoService {

    @RequestMapping(value = "/feign")
    String test();
}
```
```
@EnableFeignClients
@SpringBootApplication
public class FeignRibbonApplication {

	public static void main(String[] args) {

		SpringApplication.run(FeignRibbonApplication.class, args);
	}
}
```
When you start the application you will find spring recursively create `fallback `or `fallbackFactory `parameters defined  and lead to `StackOverflowError`

Looking forward to your reply, best wishes!

holy12345:
Hi @spencergibb @ryanjbaxter 
Today in order to verify my reasoning I conducted a debug trace Finally found why there recursive call and cause `StackOverflowError`
Let me give a detailed explanation of why this happens.

```
@FeignClient(name = "demo", url = "http://localhost:8080", fallback = DemoService.class)
public interface DemoService {

    @RequestMapping(value = "/feign")
    String test();
}
```
Through the above code we can see that the contents of the fallback is the interface itself, which means that this situation will lead to the emergence of this error (if it is other interfaces will not be this error).
`FactoryBeanRegistrySupport#doGetObjectFromFactoryBean()` method is called during spring initialization of the bean，the most important code in this method is as follows: `object = factory.getObject();`.

`object = factory.getObject ();` calls the `FeignClientFactoryBean#getObject()` method.
This method will eventually be called(`HystrixTargeter#target()`):

```
	@Override
	public <T> T target(FeignClientFactoryBean factory, Feign.Builder feign, FeignContext context,
						Target.HardCodedTarget<T> target) {
		Class<?> fallback = factory.getFallback();
		if (fallback != void.class) {
			return targetWithFallback(factory.getName(), context, target, builder, fallback);
		}
		Class<?> fallbackFactory = factory.getFallbackFactory();
		if (fallbackFactory != void.class) {
			return targetWithFallbackFactory(factory.getName(), context, target, builder, fallbackFactory);
		}

		return feign.target(target);
	}
```

The `getFromContext`method is as follows
```
	private <T> T getFromContext(String fallbackMechanism, String feignClientName, FeignContext context,
								 Class<?> beanType, Class<T> targetType) {
		Object fallbackInstance = context.getInstance(feignClientName, beanType);
}
```
The problem lies here, `context.getInstance (feignClientName, beanType);` will call the `FactoryBeanRegistrySupport#doGetObjectFromFactoryBean()` method.
This formed a recursive call and eventually causes the `StackOverflowError `to appear.

Conclusion: This problem caused by a user's writing error, although it does not allow the application to start up, but I think it is not friendly (It is even costly). So I think we should check in advance to prevent this from happening.
In addition the existing test cases already covered the code I modified.
Looking forward to your reply, best wishes. Thanks again

ryanjbaxter:
I guess this makes sense if we want to protect the user from themselves.  It doesnt make sense to specify the Feign client as the fallback and I am not sure why anyone would do that in the first place.  I dont see the harm in adding the check, but I think we can just check that the fallback class isnt the same as the Feign Client.  It is not the fact that the class is an interface that is causing the stack overflow problem.

holy12345:
Hi @ryanjbaxter 
Thank you for your reply, your suggestion sounds great. I think we just want to prevent this from happening (Like you said the users may not do this)
But what I want to say is whether the scope of our examination can be broader(In fact, `fallback ` or `fallbackFactory `definition of the content is not allowed to interface type.)
For the reasons I just said I did a more extensive check.Do you agree with this view? thanks again ：）

ryanjbaxter:
I dont think that there is anything wrong with your implementation it works out because all Feign Clients are interfaces so it prevents the recursion.  

As @spencergibb said we still need a test for it.

holy12345:
Hi @ryanjbaxter 
According to your suggestion I added a test code, please check. Thank you for your patience guidance. best wishes : )

PR #2574, opened 2017-12-22
Set proper status and health eureka urls if prefer ip address is set (1.4.x)

Aloren:


ryanjbaxter:
@Aloren there is a test failure related to the ip address can you take a look?

Aloren:
should be working now. the codebase was changed quite a bit between branches :)

Issue #2586, opened 2017-12-27
Ribbon: `java.net.UnknownHostException` should be considered as connection failure error


brenuart:
My feeling is a `UnknownHostException` should be treated as a connection failure (no connection occurred after all...).

This could be handled by adding `java.net.UnknownHostException` to the list of _connectionRelated_ exceptions in `RequestSpecificRetryHandler`.

holy12345:
Hi @brenuart 
My understanding spring cloud retry mechanism is the use of `spring-retry` completed `RequestSpecificRetryHandler`  this class will be used in retry?

brenuart:
SpringRetry is opt-in and not suitable for those using Netflix's legacy `RestClient` (retry is then handled by Ribbon itself).

holy12345:
@brenuart 
Thank you for your reply, I just read the source code. I think what you mean is that you want to make the following changes:
In the `RequestSpecificRetryHandler ` class
```
  protected List<Class<? extends Throwable>> connectionRelated = 
            Lists.<Class<? extends Throwable>>newArrayList(SocketException.class, UnknownHostException.class);
```
This is right?
In addition, I would like to say `RequestSpecificRetryHandler ` is a `ribbon ` project, if we need to add we should submit PR in `Netflix/Ribbon` ?

spencergibb:
PR welcome @brenuart. Since we're not focusing on RestClient much

brenuart:
@spencergibb I'll prepare the PR. Should we understand the _RestClient approach_ is deprecated and the new "official" setup is HttpClient + SpringRetry?

spencergibb:
Those are the defaults now.

spencergibb:
`RibbonClient` has been deprecated by Netflix since we started integration

brenuart:
@spencergibb No idea why some checks failed - the "unknown classes" issue does not seem to be related at all to my changes...
Anyway, could you please have a look at my PR comments?

spencergibb:
I'm off this week and next, so I will later, or maybe someone else will review prior

brenuart:
Happy New Year then… and see ya next year ;-)

spencergibb:
Closing in favor of #2614

PR #2587, opened 2017-12-27
Add UnknownHostException into the list of connection related errors

brenuart:
Attempt to fix https://github.com/spring-cloud/spring-cloud-netflix/issues/2586

@spencergibb The idea is to add the `java.net.UnknownHostException` to the list of connection related errors maintained by the `RequestSpecificRetryHandler`. This can be done by subclassing as shown in this PR.

However, I noticed the `RequestSpecificRetryHandler` is also used by the Apache and OkHttp versions... Does it mean they also need this fix? Do they already retry on host unknown error via another mechanism ?


ryanjbaxter:
Can you add some tests?  

For the Apache and OK HTTP clients we use Spring Retry so it should not effect those implementations.

brenuart:
Closed in favour of #2614

Issue #2353, opened 2017-10-10
Cannot get response content while using retry in zuul


lly5044:
If the response from server has the status code configured in ribbon.retryableStatusCodes, the response content will be changed and I cannot get the origin response content from zuul which is usually necessary for my application.
Will this be fixed in the future?

lly5044:
i got a fix may help solve the problem

in RetryableRibbonLoadBalancingHttpClient

<pre>
  <code>
@Override
public RibbonApacheHttpResponse execute(final RibbonApacheHttpRequest request, final IClientConfig configOverride) throws Exception {
    final RequestConfig.Builder builder = RequestConfig.custom();
    IClientConfig config = configOverride != null ? configOverride : this.config;
    builder.setConnectTimeout(config.get(
            CommonClientConfigKey.ConnectTimeout, this.connectTimeout));
    builder.setSocketTimeout(config.get(
            CommonClientConfigKey.ReadTimeout, this.readTimeout));
    builder.setRedirectsEnabled(config.get(
            CommonClientConfigKey.FollowRedirects, this.followRedirects));

    final RequestConfig requestConfig = builder.build();
    final LoadBalancedRetryPolicy retryPolicy = loadBalancedRetryPolicyFactory.create(this.getClientName(), this);
    RetryCallback retryCallback = new RetryCallback() {
        @Override
        public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws Exception {
            //on retries the policy will choose the server and set it in the context
            //extract the server and update the request being made
            RibbonApacheHttpRequest newRequest = request;
            if(context instanceof LoadBalancedRetryContext) {
                ServiceInstance service = ((LoadBalancedRetryContext)context).getServiceInstance();
                if(service != null) {
                    //Reconstruct the request URI using the host and port set in the retry context
                    newRequest = newRequest.withNewUri(new URI(service.getUri().getScheme(),
                            newRequest.getURI().getUserInfo(), service.getHost(), service.getPort(),
                            newRequest.getURI().getPath(), newRequest.getURI().getQuery(),
                            newRequest.getURI().getFragment()));
                }
            }
            if (isSecure(configOverride)) {
                final URI secureUri = UriComponentsBuilder.fromUri(newRequest.getUri())
                        .scheme("https").build().toUri();
                newRequest = newRequest.withNewUri(secureUri);
            }
            HttpUriRequest httpUriRequest = newRequest.toRequest(requestConfig);
[...]

PR #2592, opened 2017-12-28
fix losing content while retrying

lly5044:
#2353

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonRecoveryCallback.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonRecoveryCallback.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2013-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.cloud.netflix.ribbon;
 
 import org.springframework.cloud.client.loadbalancer.RetryableStatusCodeException;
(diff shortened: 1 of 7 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
```diff
@@ -96,7 +101,7 @@ public RibbonApacheHttpResponse execute(final RibbonApacheHttpRequest request, f
 
 		final RequestConfig requestConfig = builder.build();
 		final LoadBalancedRetryPolicy retryPolicy = loadBalancedRetryPolicyFactory.create(this.getClientName(), this);
-		RetryCallback retryCallback = new RetryCallback() {
+		RetryCallback<RibbonApacheHttpResponse, Exception> retryCallback = new RetryCallback<RibbonApacheHttpResponse, Exception>() {
```
ryanjbaxter: missing a space here

lly5044:
I have to new a response object in RecoveryCallback and they have different constructors. I could not doing this without using reflection. 
Or I could set the response object into RetryableStatusCodeException instead of the raw reponse I set  now.
@ryanjbaxter

spencergibb:
I'm a bit concerned with the supporting the complexity of this.

ryanjbaxter:
@lly5044 here is what I was thinking regarding extracting the `RecoveryCallback` into its own class
```
public abstract class MyCallback<T, R> implements RecoveryCallback<T> {

	protected abstract T createResponse(R response, URI uri);

	@Override
	public T recover(RetryContext context) throws Exception {
		Throwable lastThrowable = context.getLastThrowable();
		if (lastThrowable != null && lastThrowable instanceof RetryableStatusCodeException) {
			RetryableStatusCodeException ex = (RetryableStatusCodeException) lastThrowable;
			return createResponse((R) ex.getResponse(), ex.getUri());
		}
		throw new RetryException("Could not recover", lastThrowable);
	}
}
```

Hope that makes things a bit clearer.

To @spencergibb's comment about complexity.  That was part of what I was trying to do by suggesting this is in hopes of making the code less complex and cleaner.

lly5044:
I see, Thanks for your advice @ryanjbaxter

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonRecoveryCallback.java
```diff
@@ -0,0 +1,23 @@
+package org.springframework.cloud.netflix.ribbon;
```
ryanjbaxter: Need copyright header

lly5044:
any update on your discussion? @ryanjbaxter

ryanjbaxter:
@lly5044 sorry for the delay.  I have actually been thinking about this change some more and how we can simplify not only this change but the whole retry functionality in Spring Cloud Netflix.  What I think I am going to do is incorporate your changes into what I want to do to simplify things, so I am most likely going to close this PR.  I hope to open a PR today or tomorrow with my changes and be sure to CC you on the changes so you can take a look.

ryanjbaxter:
Closing in favor of #2685

PR #2598, opened 2018-01-02
V1.3.6.release http method

minsuoqingcheng:
add http method support, which can configure request method in ZuulRoute


ryanjbaxter:
I am not sure what you are doing in this PR, but you should not be changing versions.

PR #2605, opened 2018-01-02
Feature/zuul trace proxy request helper

tcawley:


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/ZuulProperties.java
```diff
@@ -838,4 +852,23 @@ public String toString() {
 				.append("}").toString();
 	}
 
+	public class Enabled {
+		private boolean enabled = false;
```
spencergibb: Why this class?

tcawley: Hi Spencer,

I could easily just put the `enabled` flag within the local `TraceProxyRequestHelper` property class, but I was trying to suggest and open an opportunity for other zuul properties to move more towards the spring-boot-actuator convention where enabling/disabling properties looks like `prefix.<some-property>.enabled` vs. `zuul.<some-property>-enabled`.  I figured that other properties could simply make use of the `Enabled` class, if desired.

I could also just make my property `zuul.trace-proxy-request-helper-enabled`, too, to be more consistent with all the other zuul properties, but was following @ryanjbaxter's property name suggestion.

Please let me know which way you'd prefer to go and I'll be happy to update my PR.

spencergibb:
No need for a separate class

tcawley:
Hi Spencer,

Just to be clear, are you saying no need for a separate `Enabled` class or the entire separate `TraceProxyRequestHelper` class?  That is, which property do you prefer to see:
`zuul.trace-proxy-request-helper-enabled` or `zuul.trace-proxy-request-helper.enabled`?

spencergibb:
Enabled class

ryanjbaxter:
Could you add some documentation for the new property?

spencergibb:
Hmm, I don't think this is something we should support. Can't you just create a 
```
@Bean
public ProxyRequestHelper proxyRequestHelper() {}
```
that isn't a `TraceProxyRequestHelper`?

tcawley:
Holy <expletive here>!  Of course!  I apologize immensely for wasting your time with this.  The simple solution was there all along.  It's not like I've been using Spring for ages or nothing.  Sometimes my brain just goes walkabout.

PR #2612, opened 2018-01-04
Fix HystrixFix StreamTask#sendMetrics() batch all metrics to one message

igorKryvenko:


spring-cloud-netflix-hystrix-stream/src/main/java/org/springframework/cloud/netflix/hystrix/stream/HystrixStreamTask.java
```diff
@@ -19,9 +19,11 @@
 import java.io.IOException;
 import java.io.StringWriter;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.concurrent.LinkedBlockingQueue;
 
+import org.apache.commons.codec.binary.StringUtils;
```
spencergibb: Do not rely on Apache.

igorKryvenko: Already fixed it, and update PR

spencergibb:
Please reformat to tabs, the diff is unreadable otherwise

PR #2613, opened 2018-01-04
Fix HystrixFix StreamTask#sendMetrics() batch all metrics to one message

igorKryvenko:


igorKryvenko:
I don't think so.
Another solution is using Gson to create one json string.

spencergibb:
Is it even compatible with turbine stream, much less backwards compatible? It needs to be backwards compatible since very likely there might be mixed versions. I would assume there would need to be a corresponding change to turbine stream

igorKryvenko:
Do u mean backward compatible with previous versions of turbine stream?

Issue #2585, opened 2017-12-27
RequestSpecificRetryHandler retries same server for SERVER_THROTTLED


brenuart:
Netflix's `RequestSpecificRetryHandler` allows to retry on *same* server for `SERVER_THROTTLED` error.

When the `okToRetryOnAllErrors` property is `true`, the retry handler allows to retry on the same server unconditionally without taking into account the actual error. This is useless when the server responded with SERVER_THROTTLED.

It looks like the _if_ statement at https://github.com/Netflix/ribbon/blob/master/ribbon-core/src/main/java/com/netflix/client/RequestSpecificRetryHandler.java#L58-L65 should be considered first before the other conditions.

I doubt Netflix is willing to change this behavior but we could override it in SpringCloud.
What do you think ?

holy12345:
Hi @brenuart 
My understanding of spring cloud not only can set the number of retries the same server but also can set the number of retries of different servers.
You mean that we should not unconditionally retry the same server, but according to the specific situation to decide whether to continue to retry the same server?(For example, according to an exception to decide whether to retry the same server)

brenuart:
Ribbon can indeed be configured to retry the same server a couple of times before trying the next one.
If the server replies with SERVER_THROTTLED, there is no point in retrying the same server - whatever the number of retries that is allowed... With the default configuration, you will get the following scenario:
- (first call) server1: SERVER_THROTTLED
- (retry same) server1: SERVER_THROTTLED
- (retry) server2: OK

The second call is useless...

holy12345:
@brenuart 
Thank you for your reply, I would like to ask whether to continue retrying the same server when an exception occurs. Can this choice be up to the user? (such as provide a property to allow users to configure to decide which the situation does not need to retry on the same server).Instead of just for `SERVER_THROTTLED`  situation

ryanjbaxter:
Is there a "server throttled" status code?  I cant seem to find any information on it, besides its use in Ribbon.

brenuart:
It is indeed a Netflix/Ribbon specific feature that is hardcoded in their RestClient. The client returns `ClientException.ErrorType.SERVER_THROTTLED` when it gets a `503 (Service Unavailable)` from the downstream server.

When `okToRetryOnAllErrors` is true, the current implementation of `RequestSpecificRetryHandler` is such that the same server will be retried on `503` - which is obviously pointless. See the code below for more insight:

```
    @Override
[...]

PR #2614, opened 2018-01-04
Retry for “SERVER_THROTTLED” and “Unknown Host” exceptions

brenuart:
#2585: Override RequestSpecificRetryHandler to first handle ClientException.ErrorType.SERVER_THROTTLED type of error and respond to not retry on the same server. Delegate to legacy behavior in other cases.

#2586: Add “java.net.UnknownHostException” to the list of connection related issues

Sorry, but I could not find any existing test case where I could add tests for these changes :(


ryanjbaxter:
I would prefer if you submit a PR for each issue

spencergibb:

Issue #1571, opened 2016-12-22
Eureka: DOWN status because of /pause not propagated to registry when "eureka.client.healthcheck.enabled: true" (Brixton.SR7)


brenuart:
The `/pause` management endpoint stops the application context.
`EurekaDiscoveryClientConfiguration` sets the InstanceStatus to DOWN in reaction to the context stop event as follows:
```
this.applicationInfoManager.setInstanceStatus(InstanceStatus.DOWN);
```

This status change triggers an `onDemandeUpdate()` on the InstanceInfoReplicator to propagate the new status back to the Eureka registry. That operation starts with a call to `discoveryClient.refreshInstanceInfo()` followed by `discoveryClient.register()` (the latter is supposed to update the status on the registry).

Unfortunately, the call to `refreshInstanceInfo()` overrides the new status with the value returned by the HealthCheckHandler if one is provided:
```
    void refreshInstanceInfo() {
        applicationInfoManager.refreshDataCenterInfoIfRequired();
        applicationInfoManager.refreshLeaseInfoIfRequired();

        InstanceStatus status;
        try {
            status = getHealthCheckHandler().getStatus(instanceInfo.getStatus());
        } catch (Exception e) {
            logger.warn("Exception from healthcheckHandler.getStatus, setting status to DOWN", e);
            status = InstanceStatus.DOWN;
        }

        if (null != status) {
            applicationInfoManager.setInstanceStatus(status);
        }
    }
```

To summarize:
- /pause stops the application context
- EurekaDiscoveryClientConfiguration.stop() is invoked and changes the InstanceStatus to DOWN
- the InstanceInfoReplicator.onDemandeUpdate() method is invoked as listener of the InstanceInfo status
- the HealthCheckHandler is invoked and returns an OK status
- the OK status overrides the DOWN status set because of /pause
- when the `register()` method is called, the status is back to OK and nothing is updated on the registry

This happens *ONLY* if Eureka is configured with `eureka.client.healthcheck.enabled: true`.
Happens on Brixton.SR7

ryanjbaxter:
Thinking outload here.....
What if the `HealthCheckHandler` took into account the app has been "paused" and returned `DOWN`?

brenuart:
That's basically the workaround I'm busy with atm. Two approaches:
1/ Customize the `HealthCheckHandler` to take the application context status into account and return DOWN unconditionally if it is stopped.
[...]

PR #2615, opened 2018-01-04
Only return health status for eureka if started. Fix for #1571

brenuart:
See https://github.com/spring-cloud/spring-cloud-netflix/issues/1571#issuecomment-351638405 for more information about the rational behind the changes.


ryanjbaxter:
This seems OK to me.  I would like to hear what @spencergibb thinks.  

Any way you can think of testing this?

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaHealthCheckHandler.java
```diff
@@ -61,6 +66,11 @@
 
 	private ApplicationContext applicationContext;
 
+	/**
+	 * {@code true} when the context stop sequence has been initiated.
+	 */
+	private boolean stopping = false;
```
spencergibb: Stopping is a bad name. Running is better

OlgaMaciaszek:
Merged with https://github.com/spring-cloud/spring-cloud-netflix/pull/3999.

Issue #2606, opened 2018-01-03
Zuul, Ribbon and Hystrix timeout confusion


matt-shaw:
Hi,

I'm slightly confused by the which timeout settings work and don't work with Zuul, Ribbon and Hytrix.

We are using Spring Cloud Edgeware release. We are just using all the default values. We are connecting Zuul to our services using Ribbon connected to Eureka. Some of our services are a little slow ~10 secs ish. How do I increase the various timeouts that seem to be triggered to allow this normal use case to not timeout??

I've set these;
ribbon:
ReadTimeout: 10000
ConnectTimeout: 10000

which stopped some of the timeouts but I'm still left with the Hystrix timeouts.

Many thanks

Matt

ryanjbaxter:
Take a look at this section in the documentation

http://cloud.spring.io/spring-cloud-static/Edgware.RELEASE/single/spring-cloud.html#_hystrix_timeouts_and_ribbon_clients

When using Zuul we automatically wrap each proxied request in a hystrix command.  So you need to configure the hystrix timeouts to be slightly larger than the ribbon timeouts.

Something like
```
hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds: 11000
ribbon:
  ConnectTimeout: 10000
  ReadTimeout: 10000
```

holy12345:
As far as I know, the current `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds ` parameter in the `Edgeware ` version does not work (for specific reasons and changes see PR https://github.com/spring-cloud/spring-cloud-netflix/pull/2581)

matt-shaw:
I think I tried what Ryan suggested but it didn't work.  I'll double check when I'm back in the office.

I thought when using Hystrix with Zuul is sets the isolation strategy to SEMAPHORE and so that timeout config doesn't have any effect or have I misunderstood?

Are you both saying that the config suggested should work but due to that bug it doesn't currently work in Edgeware?  I assume the workaround is to turn off timeouts in Hystrix?

ryanjbaxter:
You are right, you can switch the isolation strategy used by Zuul by using `zuul.ribbonIsolationStrategy=THREAD`.  I a unsure if the bug @holy12345 mentioned effects Zuul as I have not looked into it yet.  (It is on my list of things to look at)

holy12345:
[...]

PR #2581, opened 2017-12-25
Fix #2570

holy12345:
This PR is fixed #2570

PR #2616, opened 2018-01-05
Fix 2606

holy12345:
@ryanjbaxter 
I closed the previous PR(#2581) and resubmitted a PR, which solved the #2606 issue.
On the basis of the previous I added test code.

Targeted at your question in #2581(`Wouldnt this all only apply if the hystrix isolation strategy is thread and not semaphore?`)
I think regardless of the isolation strategy is the thread or semaphore, this parameter(`hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds`) can meet the timeout setting.(I have completed this test locally)

Also, I found that you added this issue to the `1.4.1 milestone` ,this PR against the branch to `master` .If you think there is no problem with my changes and want this PR against the branch to `1.4.x`, then I will resubmit a PR against the branch to `1.4.x` (I think this would be less of a workload for your merge code)

Look forward to your reply, thanks again : )

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
@@ -51,6 +51,14 @@ public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwar
 	protected ZuulFallbackProvider zuulFallbackProvider;
 	protected IClientConfig config;
 
+	/**
+	 * Indicates that the users has not configured the
+	 * hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds parameter.
+	 * Because if the timeout value is 0 then it can be considered meaningless,
+	 * so we can use 0 to determine whether this attribute configurationed.
+	 */
+	private static final int NO_TIMEOUT_PROPERTY = 0;
+
 	public AbstractRibbonCommand(LBC client, RibbonCommandContext context,
 			ZuulProperties zuulProperties) {
 		this("default", client, context, zuulProperties);
@@ -91,8 +99,13 @@ public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwar
 								.andCommandKey(HystrixCommandKey.Factory.asKey(commandKey));
 
 		final HystrixCommandProperties.Setter setter = HystrixCommandProperties.Setter()
-				.withExecutionIsolationStrategy(zuulProperties.getRibbonIsolationStrategy()).withExecutionTimeoutInMilliseconds(
-						RibbonClientConfiguration.DEFAULT_CONNECT_TIMEOUT + RibbonClientConfiguration.DEFAULT_READ_TIMEOUT);
+			.withExecutionIsolationStrategy(zuulProperties.getRibbonIsolationStrategy());
+		final DynamicIntProperty timeout = DynamicPropertyFactory.getInstance().getIntProperty(
+			"hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds", NO_TIMEOUT_PROPERTY);
+		if (timeout.get() == NO_TIMEOUT_PROPERTY) {
+			setter.withExecutionTimeoutInMilliseconds(
+				RibbonClientConfiguration.DEFAULT_CONNECT_TIMEOUT + RibbonClientConfiguration.DEFAULT_READ_TIMEOUT);
+		}
 		if (zuulProperties.getRibbonIsolationStrategy() == ExecutionIsolationStrategy.SEMAPHORE){
 			final String name = ZuulConstants.ZUUL_EUREKA + commandKey + ".semaphore.maxSemaphores";
 			// we want to default to semaphore-isolation since this wraps
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
@@ -55,9 +55,17 @@ public abstract class AbstractRibbonCommand<LBC extends AbstractLoadBalancerAwar
 	 * Indicates that the users has not configured the
 	 * hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds parameter.
 	 * Because if the timeout value is 0 then it can be considered meaningless,
-	 * so we can use 0 to determine whether this attribute configurationed.
+	 * so we can use 0 to determine whether this attribute configured.
 	 */
 	private static final int NO_TIMEOUT_PROPERTY = 0;
+
+	/**
+	 * Indicates that the users has not configured the
+	 * hystrix.command.commandKey.execution.isolation.thread.timeoutInMilliseconds parameter.
+	 * Because if the timeout value is 0 then it can be considered meaningless,
+	 * so we can use 0 to determine whether this attribute configured.
+	 */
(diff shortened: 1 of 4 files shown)


holy12345:
Hello @ryanjbaxter 
Please look at this PR, I am willing to listen to your suggestions. Perhaps my writing is not elegant enough, any of your reply is a great encouragement to me.

ryanjbaxter:
I dont really understand why you closed the previous PR.  Closing PRs and reopening new ones can lead to the history of the change being lost.  I would like to avoid doing this in the future.

The property has the word "thread" in it so on the surface it would appear that this property would only be used if the strategy thread was used.  From what I can tell from the Hystrix documentation the only way to set the timeout when using semaphore isolation is with using the Hystrix Java API.
That being said I found this thread in the Hystrix project which applies to the topic https://github.com/Netflix/Hystrix/issues/300.  If you look at the PR it appears there is a property `hystrix.command.default.execution.timeoutInMilliseconds` which can be used to set the timeout regardless of the isolation strategy.  Seems like that is the property to use.

Since this appears to be a regression in Edgware (works in Dalston) we should submit the fix against the 1.4.x stream.

spencergibb:
I'm confused why we need to reference default properties at all

holy12345:
@ryanjbaxter 
I closed the previous PR, to bring you inconvenience, I apologize, I will avoid such cases later.
I am very grateful for your patience guidance, I spent some time today to understand this problem.

This PR(https://github.com/Netflix/Hystrix/pull/664) does add this parameter(`hystrix.command.default.execution.timeoutInMilliseconds`), I also learned through this PR that this change will be reflected in the `1.4 milestone` or later release.
Now we integrate Hystrix version is `hystrix-core-1.5.12`  but in this version does not have this parameter(`hystrix.command.default.execution.timeoutInMilliseconds`).

I saw the following code in `HystrixCommandProperties`

```
 //this property name is now misleading. //TODO figure out a good way to deprecate this property name
this.executionTimeoutInMilliseconds = getProperty(propertyPrefix, key, "execution.isolation.thread.timeoutInMilliseconds", builder.getExecutionIsolationThreadTimeoutInMilliseconds(), default_executionTimeoutInMilliseconds);
```

By javadoc and my local validation I got the following conclusion:

- In the `1.5.12` version we did not `execute.timeoutInMilliseconds` this parameter.

- Currently we are still using execution.isolation.thread.timeoutInMilliseconds to represent the timeout (regardless of whether the isolation strategy is semaphore or thread, as javadoc said, the name of this parameter is confusing)

- If user do not set `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds` or `hystrix.command.commandKey.execution.isolation.thread.timeoutInMilliseconds` then we will set the timeout to `RibbonClientConfiguration.DEFAULT_CONNECT_TIMEOUT + RibbonClientConfiguration.DEFAULT_READ_TIMEOUT`

The above is my understanding, looking forward to your reply, thank you very much.

spencergibb:
It certainly looks like it is there to me https://github.com/Netflix/Hystrix/pull/664/files#diff-fdd79dd8a910636566c5483010d5fa1eR136

holy12345:
@spencergibb 
thank you for your reply,As you said, this PR(https://github.com/Netflix/Hystrix/pull/664) does increase this parameter(`hystrix.command.default.execution.timeoutInMilliseconds`).This change should also be reflected in version `1.4` or later.

But in the current version(`1.5.12`) of this property does not exist.Find any method that does not set this property(`hystrix.command.default.execution.timeoutInMilliseconds`) in the `HystrixCommandProperties ` class.

spencergibb:
What do you mean it doesn't exist? It's there

holy12345:
@spencergibb 

My point is that we can only set the timeout time by the `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds parameter` (no matter if the isolation strategy is semaphore or thread)
I understand that this PR adds a parameter `hystrix.command.default.execution.timeoutInMilliseconds` instead of `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds,` to represent the timeout period (regardless of whether the isolation strategy is semaphore or thread).

Logical processing of parameters(`hystrix.command.default.execution.timeoutInMilliseconds`) in PR(https://github.com/Netflix/Hystrix/pull/664)

```
 //thread-specific timeout is deprecated.  If the documented property is not used, still respect the deprecated value
HystrixProperty<Integer> deprecatedThreadSpecificExecutionTimeoutInMilliseconds = getProperty(propertyPrefix, key, "execution.isolation.thread.timeoutInMilliseconds", builder.getExecutionTimeoutInMilliseconds(), default_executionTimeoutInMilliseconds);
        this.executionTimeoutInMilliseconds = getProperty(propertyPrefix, key, "execution.timeoutInMilliseconds", builder.getExecutionTimeoutInMilliseconds(), deprecatedThreadSpecificExecutionTimeoutInMilliseconds.get());
```
But in the current version (`1.5.12`) this function does not exist.

spencergibb:
Right, it was reverted https://github.com/mattrjacobs/Hystrix/commit/a2c5112356e69fead98ae28f9cf5ab47e7543967#diff-fdd79dd8a910636566c5483010d5fa1e

Why are we adding this property then? This `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds` is the right one even badly named.

holy12345:
@spencergibb 
In addition, for your question(`I'm confused why we need to reference default properties at all`)
My idea is this:
I set two defaults
```
/**
 * Indicates that the users has not configured the
 * hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds parameter.
 * Because if the timeout value is 0 then it can be considered meaningless,
 * so we can use 0 to determine whether this attribute configured.
 */
 private static final int NO_TIMEOUT_PROPERTY = 0;

/**
 * Indicates that the users has not configured the
 * hystrix.command.commandKey.execution.isolation.thread.timeoutInMilliseconds parameter.
 * Because if the timeout value is 0 then it can be considered meaningless,
 * so we can use 0 to determine whether this attribute configured.
 */
 private static final int NO_OVERRIDE_TIMEOUT_PROPERTY = 0;
```
The meaning of these two parameters on behalf of the user whether to set the timeout parameter.
Because Hystrix has its own set of logic to get timeouts, my idea is not to interfere with Hystrix's logic of getting the timeout.

spencergibb:
Can we take a step back and help me understand what exactly the problem is? As I see it, trying to set sane defaults caused more problems than it was worth. Maybe we should just revert that change. @ryanjbaxter? @holy12345 why don't you hold off with PRs until we decide what the right solution should be.

holy12345:
@spencergibb 
Because before ryanjbaxter said 
`If you look at the PR it appears there is a property hystrix.command.default.execution.timeoutInMilliseconds which can be used to set the timeout regardless of the isolation strategy. Seems like that is the property to use.`

My explanation before is to explain this property we can not use, maybe because my English is not good, leading to not say very clearly. I apologize.

spencergibb:
Thanks for your help, but I think we need to take care of this one.

holy12345:
@spencergibb 
Thank you very much for your and ryanjbaxter patience answer : )
Maybe I did not do well enough, but I will always pay attention to this issue, I hope I can fix it.
Thanks again

ryanjbaxter:
Sorry I completely misunderstood what the underlying problem was.  @spencergibb maybe the right approach is to check the ribbon settings for connect and read timeout and use those if set.  Otherwise use the defaults.  Thoughts?

holy12345:
Hi @ryanjbaxter 
thank you for your reply : )

My idea is that we only need to judge whether the user set these two parameters:
Set default (for all)
`hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds`
Set up specific services
`hystrix.command.commdKey.execution.isolation.thread.timeoutInMilliseconds`
If the above two parameters both are not set then we give Hystrix a default timeout.

If one or both of these two parameters are set by the user, then we do nothing (following Hystrix's internal logic for getting the timeout)

Look forward to listening to you for better advice, thanks again.

PR #2618, opened 2018-01-05
Fix HystrixStreamTask#sendMetrics() batch all metrics to one message

igorKryvenko:


igorKryvenko:
I closed the previous PR([#2613](https://github.com/spring-cloud/spring-cloud-netflix/pull/2613)) and resubmitted a new PR.
I investigated backward compatibles with previous versions turbine stream.
In all stage branches it will be compatible, because aggregator in turbine stream can parse it as json array.

spencergibb:
Again, I don't see the point of this unless turbine stream changes also, but also remains able to ingest old messages as well

spencergibb:
ribbon is no longer supported

PR #2619, opened 2018-01-05
hotfix/add-log-ribbon

igorKryvenko:


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
```diff
@@ -60,7 +63,7 @@ public RibbonLoadBalancedRetryPolicy(String serviceId, RibbonLoadBalancerContext
 				try {
 					retryableStatusCodes.add(Integer.valueOf(code.trim()));
 				} catch (NumberFormatException e) {
-					//TODO log
+					log.warn("Can't add status code: " + code, e);
```
ryanjbaxter: I think we can use a more accurate warning message here.  We cant add the status code because the code could not be converted to an integer.

PR #2624, opened 2018-01-08
Only strip the prefix if it actually part of the path

### LINKED ISSUE #2377: Zuul: Malformed route mapping if service ID starts with zuul prefix
We are currently facing an issue with the zuul route mapping:

zuul application.yml
```
zuul:
  prefix: /test
  strip-prefix: true
```

Test-App application.yml
```
spring:
  application:
    name: testserviceA
```

Now something happens and zuul creates a malformed routing. If I request the routes endpoint I see this:

HTTP/1.1 200 
Content-Type: application/vnd.spring-boot.actuator.v1+json;charset=UTF-8
Date: Wed, 18 Oct 2017 09:04:34 GMT
Transfer-Encoding: chunked

{
    "/test/**testserviceaservicea**/**": "testservicea"
}

ryanjbaxter:
Fixes #2377

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
@@ -138,7 +138,7 @@ public class SimpleRouteLocator implements RouteLocator, Ordered {
 		}
 		String targetPath = path;
 		String prefix = this.properties.getPrefix();
-		if (path.startsWith(prefix) && this.properties.isStripPrefix()) {
+		if (path.startsWith(prefix + "/") && this.properties.isStripPrefix()) {
 			targetPath = path.substring(prefix.length());
 		}
 		if (route.isStripPrefix()) {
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -138,7 +138,7 @@ protected Route getRoute(ZuulRoute route, String path) {
 		}
 		String targetPath = path;
 		String prefix = this.properties.getPrefix();
-		if (path.startsWith(prefix) && this.properties.isStripPrefix()) {
+		if (path.startsWith(prefix + "/") && this.properties.isStripPrefix()) {
```
spencergibb: What if `prefix` already ends with a `/`?

ryanjbaxter: If prefix ended with `/`, than we would creating a path that looked like `/test//testservicea/**`.  Seems like that is invalid to begin with.

spencergibb: ok. Do we prevent that from happening somewhere? If not, we should log an issue.

spencergibb: Otherwise, LGTM

ryanjbaxter: done #2625

PR #2633, opened 2018-01-10
Improve Zuul Hystrix Timeouts

### LINKED ISSUE #2606: Zuul, Ribbon and Hystrix timeout confusion

ryanjbaxter:
Instead of allways setting the Zuul Hystrix timeouts to be the default Ribbon connect and read timeouts we now use the following logic.
* If `hystrix.command.commandKey.execution.isolation.thread.timeoutInMilliseconds` is set use that timeout
* Else if `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds` is set use that timeout
* Else if the Ribbon connect and read timeouts are set use the combination of those
* Else use the default Ribbon connect and read timeouts

Fixes #2606

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/support/AbstractRibbonCommand.java
@@ -17,6 +17,8 @@
 
 package org.springframework.cloud.netflix.zuul.filters.route.support;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.springframework.cloud.netflix.ribbon.RibbonClientConfiguration;
 import org.springframework.cloud.netflix.ribbon.RibbonHttpResponse;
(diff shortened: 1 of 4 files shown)


ayulit:
@ryanjbaxter 
Could you tell me, please, when merge commit https://github.com/spring-cloud/spring-cloud-netflix/commit/8d18dfbc3d3f594e4e58e4ecf277624ee5b6e00d into Spring Cloud **Finchley** should be expected?

ryanjbaxter:
It will be included in the next Finchley Milestone.  Details on when that will be can be found here https://github.com/spring-cloud/spring-cloud-release/milestones

PR #2656, opened 2018-01-17
Add commandProperties in HystrixCommands

making:
that makes configuring properties such as timeout easy.

For now, additional code like bellow seems to be required to change timeout

``` java
HystrixCommandGroupKey groupKey = HystrixCommandGroupKey.Factory.asKey("foo");
HystrixCommandKey commandKey = HystrixCommandKey.Factory.asKey("bar");
HystrixCommandProperties.Setter commandProperties = HystrixCommandProperties.Setter().withExecutionTimeoutInMilliseconds(timeout);
Setter setter = Setter.withGroupKey(groupKey)
		.andCommandKey(commandKey)
		.andCommandPropertiesDefaults(commandProperties);

Mono<Bar> bar = HystrixCommands.from(...)
				.setter(setter)
				.fallback(fallback)
				.toMono()
```


ryanjbaxter:
Not sure I understand why this is necessary, why can't you use the Hystrix timeout properties?

spencergibb:
You can. It's a convenience to use the programmatic way. Similar to `@HystrixProperties` used by `@HystrixCommand`.

PR #2640, opened 2018-01-12
Prevent RibbonRoutingFilter throws exceptions when Http status codes are non-standard

holy12345:
This PR fixes #2639

PR #2671, opened 2018-01-20
Prevent RibbonRoutingFilter throws exceptions when Http status codes are non-standard(Against the 1.4.x branches)

### LINKED ISSUE #2639: Zuul: RibbonRoutingFilter throws error for non standard http status codes
We are using Spring Boot 1.5.9-RELEASE and spring-cloud-starter-zuul. One of the backend services we forward requests to provide non standard http status codes like 517. 

The current version of zuul contains a version of the RibbonRoutingFilter which executes on the ClientHttpResponse object following call: `.getStatusCode().value()` 

This causes an IllegalArgumentException. Since the filter requires the int value of the status code you could replace these calls by `.getRawStatusCode()` which do not throw an exception

holy12345:
This PR fixes #2639 and against the 1.4.x branch.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-netflix</artifactId>
-		<version>1.4.2.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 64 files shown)


ryanjbaxter:
Original PR #2640

PR #2685, opened 2018-01-25
Return content of last request when retrying failed requests

### LINKED ISSUE #2353: Cannot get response content while using retry in zuul

ryanjbaxter:
Some code taken from #2592.  Fixes #2353.


ryanjbaxter:
@lly5044 would appreciate your review

lly5044:
I didn't do a test , but I think you haven't extracted the content of response before you close the connection.
Could you get the content later after you retry failed?
In HttpClientStatusCodeException And RibbonResponseStatusCodeException 
@ryanjbaxter

ryanjbaxter:
@lly5044 thanks for catching that!

ryanjbaxter:
@lly5044 is you could take a look at the related PR in Commons as well that would be great https://github.com/spring-cloud/spring-cloud-commons/pull/309

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/okhttp/OkHttpStatusCodeException.java
```diff
@@ -0,0 +1,45 @@
...
+	private Response response;
+
+	public OkHttpStatusCodeException(String serviceId, Response response, URI uri) throws IOException {
+		super(serviceId, response.code(), response, uri);
+		this.response = new Response.Builder().code(response.code()).message(response.message()).protocol(response.protocol())
+				.request(response.request()).headers(response.headers()).handshake(response.handshake())
+				.cacheResponse(response.cacheResponse()).networkResponse(response.networkResponse())
+				.priorResponse(response.priorResponse()).sentRequestAtMillis(response.sentRequestAtMillis())
+				.body(response.peekBody(Integer.MAX_VALUE)).build();
+		response.close();
```
spencergibb: Moving this here seems very strange to me.

ryanjbaxter: Since the exception(s) were actually reading in the response, it seemed to make sense to me that it would be responsible for closing the response as well.  I suppose we can move it outside of the constructor, it just seemed disconnected that way.

ryanjbaxter: wrapping all this logic about copying over the response body in the exception also seemed to simplify the code in load balancing clients.  if we were to move the close call back to the load balancing clients than we would also have to move the code that copies the response body back there as well and it just added to something that was already complex.  since the fallback was using the response from the exception it just seemed to make the most sense to me to keep all that logic in the exception itself.

spencergibb:
Ok, maybe the class shouldn't be an exception, but something that returns one? Having that kind of logic in an exception feels wrong to me.

ryanjbaxter:
@spencergibb think this is a good compromise, let me know what you think

PR #2710, opened 2018-02-05
Set secure health check url

### LINKED ISSUE #2673: Eureka SSL config and separate management port registering with wrong secureHealthCheckUrl
Hi,

we are using SSL configuration for our services and we have differing server.port and management.port.

Looking at how the instances are registered, the secureHealthCheckUrl seems to be wrong. It uses the server.port instead of the management.port. This only happens with ssl configuration. The problem does not occur without ssl.

```xml
<instance>
    <!-- ... -->
    <port enabled="false">10007</port>
    <securePort enabled="true">10007</securePort>
    <!-- ... -->
    <metadata>
          <management.port>11007</management.port>
    </metadata>
    <homePageUrl>http://localhost:10007/</homePageUrl>
    <statusPageUrl>http://localhost:11007/info</statusPageUrl>
    <healthCheckUrl>http://localhost:11007/health</healthCheckUrl>
    <secureHealthCheckUrl>https://localhost:10007/health</secureHealthCheckUrl>
    <!-- ... -->
</instance>
```
configuration:
```
server.port=10007
server.ssl.enabled=true
server.ssl.key-alias= ...
server.ssl.key-store= ...

management.port=11007
management.security.enabled=false
management.ssl.enabled=true
management.ssl.key-alias= ...
management.ssl.key-store= ...

eureka.instance.metadata-map.management.port=${management.port}
eureka.instance.nonSecurePortEnabled=false
eureka.instance.securePortEnabled=true
eureka.instance.securePort=${server.port}
```
The only workaround that works for me ist registering manually for each service the secure-health-check-url with hardcoded port, since SPEL expressions are not evaluated.

eureka-client-1.7.0
Spring-cloud version Edgware.RELEASE (also Dalston.SRx tested...)
spring-boot version 1.5.9.Release

ryanjbaxter:
Fixes #2673

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java
@@ -24,11 +24,16 @@ public class DefaultManagementMetadataProvider implements ManagementMetadataProv
 			return null;
 		}
         String healthCheckUrl = getHealthCheckUrl(instance, serverPort, serverContextPath,
-                managementContextPath, managementPort);
+                managementContextPath, managementPort, false);
         String statusPageUrl = getStatusPageUrl(instance, serverPort, serverContextPath,
                 managementContextPath, managementPort);
 
-        return new ManagementMetadata(healthCheckUrl, statusPageUrl, managementPort == null ? serverPort : managementPort);
+        ManagementMetadata metadata = new ManagementMetadata(healthCheckUrl, statusPageUrl, managementPort == null ? serverPort : managementPort);
+        if(instance.isSecurePortEnabled()) {
+            metadata.setSecureHealthCheckUrl(getHealthCheckUrl(instance, serverPort, serverContextPath,
+                    managementContextPath, managementPort, true));
+        }
+        return metadata;
 	}
 
 	private boolean isRandom(Integer port) {
@@ -36,10 +41,10 @@ public class DefaultManagementMetadataProvider implements ManagementMetadataProv
 	}
 
 	private String getHealthCheckUrl(EurekaInstanceConfigBean instance, int serverPort, String serverContextPath,
-                                     String managementContextPath, Integer managementPort) {
+                                     String managementContextPath, Integer managementPort, boolean isSecure) {
         String healthCheckUrlPath = instance.getHealthCheckUrlPath();
         String healthCheckUrl = getUrl(instance, serverPort, serverContextPath, managementContextPath,
-                managementPort, healthCheckUrlPath);
+                managementPort, healthCheckUrlPath, isSecure);
         log.debug("Constructed eureka meta-data healthcheckUrl: " + healthCheckUrl);
         return healthCheckUrl;
     }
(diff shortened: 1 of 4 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java
```diff
@@ -24,22 +24,27 @@ public ManagementMetadata get(EurekaInstanceConfigBean instance, int serverPort,
...
-                managementContextPath, managementPort);
+                managementContextPath, managementPort, false);
         String statusPageUrl = getStatusPageUrl(instance, serverPort, serverContextPath,
                 managementContextPath, managementPort);
 
-        return new ManagementMetadata(healthCheckUrl, statusPageUrl, managementPort == null ? serverPort : managementPort);
+        ManagementMetadata metadata = new ManagementMetadata(healthCheckUrl, statusPageUrl, managementPort == null ? serverPort : managementPort);
+        if(instance.isSecurePortEnabled()) {
+            metadata.setSecureHealthCheckUrl(getHealthCheckUrl(instance, serverPort, serverContextPath,
+                    managementContextPath, managementPort, true));
```
spencergibb: Why the new parameter? Isn't it available via `instance.isSecurePortEnabled()`?

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java
```diff
@@ -48,19 +53,19 @@ public String getStatusPageUrl(EurekaInstanceConfigBean instance, int serverPort
...
     private String getUrl(EurekaInstanceConfigBean instance, int serverPort,
                           String serverContextPath, String managementContextPath,
-                          Integer managementPort, String urlPath) {
+                          Integer managementPort, String urlPath, boolean isSecure) {
         managementContextPath = refineManagementContextPath(serverContextPath, managementContextPath, managementPort);
         if (managementPort == null) {
             managementPort = serverPort;
         }
-        String scheme = instance.getSecurePortEnabled() ? "https" : "http";
+        String scheme = isSecure ? "https" : "http";
```
spencergibb: Why this change?

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java
ryanjbaxter: Yes but if I leave the code as is then both the health check url and the status page url will also be secure.  And I think we wanted to leave the existing functionality as is but just set the correct secure health check url.  So to do this and reuse the code I added the parameter so when we call getHealthCheckUrl and getStatusPageUrl we can explicitly make sure it is not secure even though the secure port is enabled.

spencergibb: I see.

Issue #2724, opened 2018-02-11
#2483 causing NPE when no servers available for the load balancer


lchayoun:
The NPE is caused by the change in `AbstractRibbonCommand#run` method.
Before this change `this.client.executeWithLoadBalancer(request, config)` was always invoked, handling the case that no servers are available through `LoadBalancerContext#getServerFromLoadBalancer` method.

Stack Trace:
```
Caused by: java.lang.NullPointerException: null
	at org.springframework.cloud.netflix.ribbon.RibbonLoadBalancerClient$RibbonServer.getHost(RibbonLoadBalancerClient.java:169)
	at org.springframework.cloud.netflix.ribbon.apache.RetryableRibbonLoadBalancingHttpClient$1.doWithRetry(RetryableRibbonLoadBalancingHttpClient.java:129)
	at org.springframework.cloud.netflix.ribbon.apache.RetryableRibbonLoadBalancingHttpClient$1.doWithRetry(RetryableRibbonLoadBalancingHttpClient.java:119)
	at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:287)
	at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:180)
	at org.springframework.cloud.netflix.ribbon.apache.RetryableRibbonLoadBalancingHttpClient.executeWithRetry(RetryableRibbonLoadBalancingHttpClient.java:178)
	at org.springframework.cloud.netflix.ribbon.apache.RetryableRibbonLoadBalancingHttpClient.execute(RetryableRibbonLoadBalancingHttpClient.java:152)
	at org.springframework.cloud.netflix.ribbon.apache.RetryableRibbonLoadBalancingHttpClient.execute(RetryableRibbonLoadBalancingHttpClient.java:58)
	at org.springframework.cloud.netflix.zuul.filters.route.support.AbstractRibbonCommand.run(AbstractRibbonCommand.java:185)
	at org.springframework.cloud.netflix.zuul.filters.route.support.AbstractRibbonCommand.run(AbstractRibbonCommand.java:52)
	at com.netflix.hystrix.HystrixCommand$2.call(HystrixCommand.java:302)
	at com.netflix.hystrix.HystrixCommand$2.call(HystrixCommand.java:298)
	at rx.internal.operators.OnSubscribeDefer.call(OnSubscribeDefer.java:46)
	... 148 common frames omitted
```

ezraroi:
[...]

PR #2727, opened 2018-02-14
Fix npe when no servers available

pway99:
This PR is for issue #2724 where an NPE is thrown when no servers are available.

The Retryable http and ok load balancing clients have been updated so that a netflix `ClientException` is thrown when either the server or its host is null.  This allows the `RibbonRoutingFilter` to handle the these exceptions and present a more useful error message.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
@@ -117,19 +117,15 @@ public class RetryableRibbonLoadBalancingHttpClient extends RibbonLoadBalancingH
 
 		final RequestConfig requestConfig = builder.build();
 		final LoadBalancedRetryPolicy retryPolicy = loadBalancedRetryPolicyFactory.create(this.getClientName(), this);
-		RetryCallback<RibbonApacheHttpResponse, IOException> retryCallback = new RetryCallback<RibbonApacheHttpResponse, IOException>() {
+		RetryCallback<RibbonApacheHttpResponse, Exception> retryCallback = new RetryCallback<RibbonApacheHttpResponse, Exception>() {
 			@Override
-			public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws IOException {
+			public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws Exception {
 				//on retries the policy will choose the server and set it in the context
 				//extract the server and update the request being made
 				RibbonApacheHttpRequest newRequest = request;
 				if(context instanceof LoadBalancedRetryContext) {
 					ServiceInstance service = ((LoadBalancedRetryContext)context).getServiceInstance();
-					try {
-						validateServiceInstance(service);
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
```diff
@@ -124,7 +126,13 @@ public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws IOExcep
 				RibbonApacheHttpRequest newRequest = request;
 				if(context instanceof LoadBalancedRetryContext) {
 					ServiceInstance service = ((LoadBalancedRetryContext)context).getServiceInstance();
-					if(service != null) {
+					if (service == null) {
```
ryanjbaxter: Can we move this code into its own method, perhaps in the super class?

pway99: yes.  I have moved the validation logic to the AbstractLoadBalancingClient.

pway99:
hello @ryanjbaxter, I also moved ServiceInstanceChooser implementation to the AbstractLoadBalancing client and am curious to know if you have any concerns about that change.  Thanks

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
```diff
@@ -124,14 +125,17 @@ public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws IOExcep
...
 				if(context instanceof LoadBalancedRetryContext) {
 					ServiceInstance service = ((LoadBalancedRetryContext)context).getServiceInstance();
-					if(service != null) {
-						//Reconstruct the request URI using the host and port set in the retry context
-						newRequest = newRequest.withNewUri(UriComponentsBuilder.newInstance().host(service.getHost())
-								.scheme(service.getUri().getScheme()).userInfo(newRequest.getURI().getUserInfo())
-								.port(service.getPort()).path(newRequest.getURI().getPath())
-								.query(newRequest.getURI().getQuery()).fragment(newRequest.getURI().getFragment())
-								.build().encode().toUri());
+					try {
```
ryanjbaxter: Why do you catch the exception in the Apache implementation and not in the OK implementation?

pway99: The `RetryCallback` within the Apache implementation is defined for an `IOException` where the OK implementation is less specific and handles an `Exception`.   Updating the Apache implementation so the callback is less specific would improve this change.

pway99: Hello @ryanjbaxter.  There are a few unused imports in both the Apache and OK retryable clients.  Would you mind if I pushed another update?

Issue #2443, opened 2017-11-14
NullPointerException in LoadBalancerFeignClient


akefirad:
The method `CachingSpringLoadBalancerFactory.create` has:
```java
public FeignLoadBalancer create(String clientName) {
  if (this.cache.containsKey(clientName)) {
    return this.cache.get(clientName);
  }

  // create a new client, cache it, and return it.
}

```
But having `cache` field of type `ConcurrentReferenceHashMap` means `this.cache.get(clientName)` regardless of `this.cache.containsKey(clientName)`. Why not:
```java
public FeignLoadBalancer create(String clientName) {
  FeignLoadBalancer cached = this.cache.get(clientName);
  if (cached != null) {
    return cached;
  }

  // create a new client, cache it, and return it.
}
```

spencergibb:
Can you describe how to recreate the problem?

akefirad:
The `create` method is being used in `LoadBalancerFeignClient`:
```java
@Override
public Response execute(Request request, Request.Options options) throws IOException {
  try {
    URI asUri = URI.create(request.url());
    String clientName = asUri.getHost();
    URI uriWithoutHost = cleanUrl(request.url(), clientName);
    FeignLoadBalancer.RibbonRequest ribbonRequest = new FeignLoadBalancer.RibbonRequest(
        this.delegate, request, uriWithoutHost);

    IClientConfig requestConfig = getClientConfig(options, clientName);
    return this.lbClientFactory.create(clientName).executeWithLoadBalancer(ribbonRequest,
        requestConfig).toResponse();
  }
  catch (ClientException e) {
    // Handle exception
  }
}
```
The `return` line might throw NPE (we're are experiencing it with v1.2.2. Nevertheless the code is the same). Did I answer your question?

spencergibb:
No. I'd like to recreate the problem, so I can write a test and then fix it.

akefirad:
I'm not sure how you want to right a test for this considering the fact that it involves race condition and GC activities. Anyway here's a dirty code we managed to reproduce the scenario once with `-Xmx50m` (please note that  it may or may not fail):
```java
        Map<String, Object> map = new ConcurrentReferenceHashMap<>();
        ExecutorService executor = Executors.newFixedThreadPool(2);

        Future<?> puts = executor.submit(() ->
                IntStream.range(1, Integer.MAX_VALUE)
                        .forEach(value -> map.put("key-" + value, new Object())));

[...]

PR #2734, opened 2018-02-21
Fix the race condition with the GC in CachingSpringLoadBalancerFactory

jebeaudet:
Fix for #2443


ryanjbaxter:
@jebeaudet Feign has been moved to its own project can you please submit a PR there https://github.com/spring-cloud/spring-cloud-openfeign

Issue #2763, opened 2018-03-07
FeignClientFactoryBean  could set Encoder/Decoder with name and type other than only with type from context ?


abnerbos:
Scenes：
     My project need fetch two RPC service from two system which  provide @Configuration class to auto import. Each @Configuration class already define itself  Encoder And Decoder , it's difference class.
Just as:
```
@Configuration
@EnableFeignClients({"xxx.client.spring"})
public class AClientSpringConfig {
    private MappingJackson2HttpMessageConverter jsonConverter;
    @Bean
    public Decoder feignDecoder() {
        ObjectFactory<HttpMessageConverters> objectFactory = () -> {
            return new HttpMessageConverters(false, Lists.newArrayList(new HttpMessageConverter[]{this.jsonConverter}));
        };
        return new ResponseEntityDecoder(new SpringDecoder(objectFactory));
    }
    @Bean
    public Encoder feignEncoder() {
        ObjectFactory<HttpMessageConverters> objectFactory = () -> {
            return new HttpMessageConverters(false, Lists.newArrayList(new HttpMessageConverter[]{this.jsonConverter}));
        };
        return new SpringEncoder(objectFactory);
    }
     @Bean
      public ErrorDecoder feignErrorDecoder() {
        return defaultFeignErrorDecoder();
      }
....
```

After import the two Configuration class, my project startup fail with the error says there is two Encoder bean AClientSpringConfig.feignEncoder and  BClientSpringConfig.feignEncoder .
```
Field XXService in XXX.XXService required a single bean, but 2 were found:
	- feignErrorDecoder: defined by method 'feignErrorDecoder' in XXX.AClientSpringConfig
	- feignErrorDecoder: defined by method 'feignErrorDecoder' in XXX.BClientSpringConfig
```

Matters:
Spent some time to read spring-cloud-netflix code, find the matter code in FeignClientFactoryBean.class:
```
protected Feign.Builder feign(FeignContext context) {
		FeignLoggerFactory loggerFactory = get(context, FeignLoggerFactory.class);
		Logger logger = loggerFactory.create(this.type);

		// @formatter:off
		Feign.Builder builder = get(context, Feign.Builder.class)
				// required values
				.logger(logger)
				.encoder(get(context, Encoder.class))
				.decoder(get(context, Decoder.class))
				.contract(get(context, Contract.class));
		// @formatter:on

		...

		return builder;
	}

	protected <T> T get(FeignContext context, Class<T> type) {
[...]

PR #2770, opened 2018-03-10
FeignContext add method getSuitableInstance to get instance with eurekaServiceName and classType

abnerbos:
for the issue https://github.com/spring-cloud/spring-cloud-netflix/issues/2763 , 
do get bean from context could with eurekaServiceName and beanType.

    /**
     * getSuitableInstance from context
     *
     * 1.without bean for the type return null
     * 2.only exist one return the bean
     * 3.choose the one which has @Primary annotation
     * 4.choose the one which name is FeignProducerName+type.getSimpleName
     * 5.on enough info for choose from multiple bean ,return null
     * @param name
     * @param type
     * @param <T>
     * @return
     */
    public <T> T getSuitableInstance(String name, Class<T> type)

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/feign/FeignClientFactoryBean.java
@@ -124,7 +124,7 @@ class FeignClientFactoryBean implements FactoryBean<Object>, InitializingBean,
 	}
 
 	protected <T> T get(FeignContext context, Class<T> type) {
-		T instance = context.getInstance(this.name, type);
+		T instance = context.getSuitableInstance(this.name, type);
 		if (instance == null) {
 			throw new IllegalStateException("No bean found of type " + type + " for "
 					+ this.name);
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@abnerbos Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2770)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2770) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@abnerbos Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2770)!

spencergibb:
This is the wrong approach. You should put decoders on specific configurations per feign client.

abnerbos:
The problem has been solved because each feign client Encoder/Decoder is in web auto scan path, so while other feign client use the type get  Encoder/Decoder  will get wrong, thank you for your answer.

And this solution really not elegant, but confused me is the way obtain the bean, maybe in the parent context, maybe in other feign client default config or each feign own config, It would be better if there was a clearer way.

Issue #2832, opened 2018-04-10
Zuul: no Content-Length in client response when both request and upstream are gzip


brenuart:
Zuul *never* includes the `Content-Length` header in the client response if upstream is gzipped.

See lines: https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java#L243-L248

However, I see no reason not to forward the Content-Length header if present on a gzipped upstream response if the client accepts Gzip too.
What you think ?

ryanjbaxter:
Yeah I agree not sure why we wouldnt include it

brenuart:
Want a PR for it ?

PR #2833, opened 2018-04-10
Always forward Content-Length header from origin when set 

brenuart:
unless response is compressed and client does not support it.

Fix for #2832


brenuart:
Sorry - but I have no idea why the test failed :-(
I don't see how it is related to the changes I have made...

spencergibb:
rebuilding

spencergibb:
looks better

ryanjbaxter:
Any way to add a test for this?

brenuart:
Working on it.

 

Unfortunately, it appears the existing `SendResponseFilterTests` are possibly making incorrect assumptions about the actual behaviour of Zuul filters.

For instance, I’m worried about the following line of code: https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-zuul/src/test/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilterTests.java#L203

 

If Zuul would behave this way, a Content-length header would always be included in all responses – which is not the case.

Moreover, I cannot find anywhere in the existing filters any traces of setting this property on the RequestContext.

Issue #2807, opened 2018-03-31
Zuul: HandlerMapping refreshed twice for every HeartbeatEvent


brenuart:
When Zuul is running as a _proxy_ (`@EnableZuulProxy`), the`ZuulHandlerMapping` is set dirty twice for every single discovery heartbeat event.

It happens a first time because of `ZuulServerAutoConfiguration#ZuulRefreshListener` and a second time by `ZuulProxyAutoConfiguration#ZuulDiscoveryRefreshListener`.

Both seems to have the same trigger conditions and to serve the same purpose.
Is this expected ?

----
_Tested on Edgware.SR3_

ryanjbaxter:
Nope, not as far as i know

brenuart:
Is this something you’d like a PR for?

spencergibb:
Yes!

brenuart:
The `Server` configuration triggers a refresh for the following events:
- `ContextRefreshedEvent`
- `RefreshScopeRefreshedEvent`
- `RoutesRefreshedEvent`
- `HeartBeatEvent` (but forgot the `ParentHeartbeatEvent`)

The `Proxy` configuration adds support for:
- `ParentHeartbeatEvent`
- `InstanceRegisteredEvent`

According to me, a refresh should occur for all these events for both the _Server_ and _Proxy_ configurations. A single RefreshListener could be used and created by the _Server_ configuration - with nothing special in the _Proxy_ configuration.

Does it sounds correct?

PR #2836, opened 2018-04-10
Refactor HandlerMapping refresh into a single listener as per #2807

brenuart:


brenuart:
Ok for me - RefreshListeners were not covered by any test case and I just removed one of the two instances - I though coverage would have improved... :(

spencergibb:
We're not super sensitive to codecov, more informational.

spencergibb:
Merged with some polish via 196abf2a24f30f1632f4032703c471602c874556

Issue #2805, opened 2018-03-30
Zuul fails to forward gzipped response with an empty body when the client does not Accept-Encoding: gzip


brenuart:
When the client does not accept gzip encoding, Zuul attempts to uncompress the response received from the downstream service (if compressed).

Unfortunately, it fails with a `500` if the compressed response does not have a body. The exception is thrown by the `SendResponseFilter#writeResponse()` method when it creates a new `GZipInputStream` on the response stream. Since the stream is empty, a `java.io.EOFException` is thrown instead of the only expected `ZipException`.

An easy way to reproduce the issue is to send a HEAD request for which there is no body in the response (as per the spec) as follows:

```
curl -i --head http://localhost:8888/service/info
```

Note there is no `Accept-Encoding: gzip` header in the request.

----
_Tested on Edgware.SR3_

brenuart:
Please note that the problem would not appear if the response had a `Content-Length: 0` header - in this case the SendResponseFilter won't try to un compress the response stream.
Unfortunately, we cannot rely on this head being always present - it is not mandatory and certainly not for HEAD requests which are not expected to have a body in their response.

spencergibb:
so the solution is, if empty response and no `Accept-Engoding: gzip`, don't create `GZipInputStream`?

brenuart:
Yes, I think so. That’s what we did actually. 
To be 100% correct, we try to read a byte from the response stream using a PushbackStream and if fails we set the Content-Length to zero on the RequestContext. This way the current impl does not try to Gunzip it.

spencergibb:
PR?

brenuart:
Sure thing. But in a week when back from holiday if you don’t mind ;-)

spencergibb:
NP, my holiday was last week :smile:.

brenuart:
As written in the PR, I managed to keep the original  behaviour in an efficient way (a single copy of a few bytes only).

Issue #2831, opened 2018-04-10
may be found a bug for httpclient connection leak


zhangxsh:
refer #2814 

I use a subclass of DefaultErrorAttributes  to deal with exceptions

```
 @Component("unifyErrorAttributes")
public class UnifyErrorAttributes extends DefaultErrorAttributes {
```

But I found that when exception generated,the http client connection can not be released.

here is normal routed request log,there are leased and released both exist,it's normal:

![image](https://user-images.githubusercontent.com/37093837/38555755-f252f072-3cf9-11e8-9c53-29d42729a0c0.png)

But when route failed,only the http connection leased action exists but no released action,bcz the coming exception disturbed the release action in future:

![qq 20180410200139](https://user-images.githubusercontent.com/37093837/38555948-93a0b400-3cfa-11e8-8551-5abfdaff6619.jpg)

the PostErrorHandlerFilter.java is a post filter,it throw an exception when route failed.

```
    @Override
        public Object filter() {
        HttpServletResponse response = RequestContext.getCurrentContext().getResponse();
        int cde = response.getStatus();
        String uri=getOriginalRequest().getRequestURI();
        HttpStatus status=HttpStatus.valueOf(cde);
        String req_detail= (String) 
        ThreadLocalCache.get(GatewayFilterConstants.CacheKey.HTTP_REQ_DETAIL);
        String body= (String) 
        ThreadLocalCache.get(GatewayFilterConstants.CacheKey.HTTP_REQ_BODY_DETAIL);


        if(status.isError()){//||status.is3xxRedirection()){
            runtimeURLCollection.errorUrls(uri,";"+cde+"[route response error]");
            throw new GateWayException(BasicErrorEnum.ERROR_1001,"status:"+cde);
        }
        return null;
    }
```

so, at last,all of the connections are not released,my app is down.'netstat' shows all TCP  in CLOSE_WAIT
my question:
1、is this a problem?
2、how to deal with that?

ryanjbaxter:
Is `PostErrorHandlerFilter` something you wrote?

zhangxsh:
BTW，PostErrorHandlerFilter is a very very simple post filter，after request routed，it check the response code，if not 2xx then throw an exception，if an exception throws during the time，the http connection will not be released

ryanjbaxter:
Why throw an exception?

brenuart:
My feeling is the issue is larger than what originally described.
[...]

PR #2838, opened 2018-04-11
Handle GZip response from downstream with an empty body.

brenuart:
Fix for  #2805.
- Do not complain when receiving an empty response stream from origin with `Content-Encoding: gzip` but instead don’t try to uncompress the stream and send an empty response too.
- Properly rewind the response stream from origin upon failure to initialize the GZIPInputStream when decompression is needed then send the response “asis” to the client but without the `Content-Encoding: gzip` and `Content-Length` headers.
- Additional tests to assert the behaviour of the filter with regards to Content-Length propagation and response uncompress when required.

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
@@ -17,11 +17,14 @@
 package org.springframework.cloud.netflix.zuul.filters.post;
 
 import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.SequenceInputStream;
 import java.util.List;
+import java.util.Objects;
 import java.util.zip.GZIPInputStream;
 
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -212,6 +186,47 @@ else if (context.getResponseGZipped() && isGzipRequested) {
...
+	protected InputStream handleGzipStream(InputStream in) throws Exception {
+		// Record bytes read during GZip initialization to allow to rewind the stream if needed
+		//
+		RecordingInputStream stream = new RecordingInputStream(in);
+		try {
+			return new GZIPInputStream(stream);
+		}
+		catch (java.util.zip.ZipException ex) {
+			
+			if (stream.getBytesRead()==0) {
```
ryanjbaxter: could we check this without throwing an exception?

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -212,6 +186,47 @@ else if (context.getResponseGZipped() && isGzipRequested) {
...
+				// stream was empty, return the original "empty" stream
+				return in;
+			}
+			else {
+				// reset the stream and assume an unencoded response
+    			log.warn(
```
ryanjbaxter: spacing

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
```diff
@@ -121,60 +124,32 @@ private void writeResponse() throws Exception {
...
 			}
-			boolean isGzipRequested = false;
-			final String requestEncoding = context.getRequest()
-					.getHeader(ZuulHeaders.ACCEPT_ENCODING);
-
-			if (requestEncoding != null
-					&& HTTPRequestUtils.getInstance().isGzipped(requestEncoding)) {
-				isGzipRequested = true;
+			else {
+    			is = context.getResponseDataStream();
```
ryanjbaxter: spacing

brenuart:
Since you are "reading" this code, I have another question not directly related to the current issue:

`writeResponse()` does it's best to guarantee that the response stream from origin is properly closed in all cases (cfr. release pooled connection). However, this does not cover exceptions thrown outside of the method, like during `addResponseHeader()` for instance.
As a solution we could wrap the entire `run()` method with a try/finally and move the close logic there.

But still, this would only guarantee that the response is closed *IF* it goes through the SendResponseFilter. What would happen if an exception is raised somewhere between say the RibbonRoutingFilter and SendResponseFilter filters? Zuul should normally direct the request to the "error" filters - but none of them since to care about closing the response like what the SendResponseFilter does. This may be what issue #2831 is all about...

A more robust approach would be:
- filters should explicitly register their resources that need to be disposed when the request is fully completed
- the RequestContext is `unset` by the ZuulServletFilter. We could use a custom RequestContext that would dispose registered resources when `unset()` is invoked.

I know Zuul is becoming obsolete now... Do you think this is worth opening another issue?

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
ryanjbaxter: OK I wasnt sure if you could use `InputStream.available` to do the same thing

ryanjbaxter:
Yes open another issue so we can discuss that separately

brenuart:
I will.

Issue #2887, opened 2018-04-26
Use ZuulServletFilter instead of ZuulServlet


candrews:
In my Spring Boot 2.0.1 / Spring Cloud Finchley.RC1 application using `@EnableZuulProxy`, I'd like to have some URLs handled by Spring MVC and others by Zuul, like this:
/actuator/** - > Spring Boot Actuator
/** -> Zuul

I also would like buffering to be disabled (as currently works with paths under `zuul.servlet-path`, which by default is `/zuul`) and a minimum of other things to be done before zuul to get the best performance and fewest compatibility problems (such as with large file uploads).

My initial approach was to set `zuul.servlet-path=/`, but then requests to `/actuator/health`, for example, would still get processed by zuul. Next, I also set `zuul.ignored-patterns=/actuator/**` but that results in a blank page being returned (status 200 with no body) - ZuulServlet ignores the request, but since it's a servlet not a filter, there's nothing else to process the request so nothing further happens (`No route found for uri: /actuator/health` is logged).

I don't think there is a way to do what I'm attempting to do right now.

One solution I found is to use ZuulServletFilter instead. Configure a bean like this:
```
@Bean
public FilterRegistrationBean<ZuulServletFilter> zuulServletFilter(){
	final FilterRegistrationBean<ZuulServletFilter> filterRegistration = new FilterRegistrationBean<>();
	filterRegistration.setFilter(new ZuulServletFilter());
	filterRegistration.setOrder(Ordered.HIGHEST_PRECEDENCE);
	filterRegistration.addInitParameter("buffer-requests", "false");
	return filterRegistration;
}
```
unset `zuul.servlet-path`  and set `zuul.ignored-patterns=/actuator/**` and it works as expected (requests to /actuator go to actuator, everything else is handled by the configured zuul routes).

I propose that ZuulServlet not be used, and ZuulServletFilter be used instead. I don't think there's a downside.

ryanjbaxter:
Maybe I am missing something, with the following configuration I was able to have Zuul route the request to `/mypath` correctly and also still access all the actuator endpoints.
```
zuul.ignored-patterns=/actuator/**
zuul.routes.google.path=/mypath/**
zuul.routes.google.url=http://httpbin.org
management.endpoints.web.exposure.include=*
```

spencergibb:
I think it specifically has to do with a default route `/**`

candrews:
> I think it specifically has to do with a default route /**

Correct.

[...]

PR #2888, opened 2018-04-26
Use ZuulServletFilter instead of ZuulServlet

candrews:
Using the filter instead of the servlet allow unhandled/ignored requests to pass through Zuul to be handled by something else, as opposed to the servlet which cannot do so.

See https://github.com/spring-cloud/spring-cloud-netflix/issues/2887

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/ZuulServerAutoConfiguration.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/ZuulServerAutoConfiguration.java
@@ -18,6 +18,7 @@
 package org.springframework.cloud.netflix.zuul;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Map;
 import java.util.List;
 
@@ -29,6 +30,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClas
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.boot.web.servlet.error.ErrorController;
 import org.springframework.cloud.client.actuator.HasFeatures;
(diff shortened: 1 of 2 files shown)


ryanjbaxter:
The build on CircleCi is not passing, was it passing locally for you?  If so I can try and build things locally.

candrews:
The errors in circle ci are all unrelated to my changes, as far as I can tell. I do not know why they're happening.

ryanjbaxter:
OK we have problems like this sometimes.  I will build it locally tomorrow.  Thanks.

ryanjbaxter:
Things are not building locally for me either.  Your branch has not merged in some of the latest changes from master so it doesnt even compile.  But even so, merging in those changes, I am still getting test failures.

```
Failed tests:   postWithForm(org.springframework.cloud.netflix.zuul.FormZuulServletProxyApplicationTests): expected:<200> but was:<404>
  postWithMultipartFile(org.springframework.cloud.netflix.zuul.FormZuulServletProxyApplicationTests): expected:<200> but was:<404>
  postWithMultipartForm(org.springframework.cloud.netflix.zuul.FormZuulServletProxyApplicationTests): expected:<200> but was:<404>
  postWithUTF8Form(org.springframework.cloud.netflix.zuul.FormZuulServletProxyApplicationTests): expected:<200> but was:<404>
```

ryanjbaxter:
I cant really explain why it is happening with your change but it looks like the call at this line is just blocking waiting forever 
https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/RibbonRoutingFilter.java#L159

spencergibb:
I don't think this is something we are going to merge going forward.

spencergibb:
So far, you are the only one asking for it and we aren't eager to change something we're aren't going to invest major effort in long term.

candrews:
> So far, you are the only one asking for it and we aren't eager to change something we're aren't going to invest major effort in long term.

People aren't asking for "Use ZuulSevletFilter instead of ZuulServlet" directly because they don't know that's what they want. Every person who has to modify their application by using the special `/zuul` path, or who encounters a problem with uploads and has to debug it, is indirectly asking for this change.

spencergibb:
I can bring it up with the team.

ryanjbaxter:
We lost the label, I will add it back and we will talk about it, sorry.

candrews:
> Can you make this opt in via a property `zuul.use-filter=true` or something?

We could do that... but this approach is faster/better in every use case. There's no reason to ever have it off. (If there is, please let me know as I'm clearly not aware of it :smile:  )

spencergibb:
It's not a minor change. The servlet has been the default for 4+ years in this project. I'd like it to be opt-in. If enough people use it, we can change the default in the future.

candrews:
Implemented `zuul.use-filter` as described.

spencergibb:
Closed with some polish via 703b4ce577a29409d284f47a4507d1f52b8d3697

Thanks for your first contribution to this project!

spencergibb:
cherry-picked to 2.0.x

PR #2901, opened 2018-05-02
fix the random port bug.

naah69:
## random port bug about spring-boot and spring-cloud-eureka
it work fine in spring-boot.
but it take a mistake in spring-cloud-eureka.

application.yml
```
server:
  port: ${random.int[11001,12000]}
```

we know it is a random port for web server.

server.port still be random after the config covert to class
and it run random every time when we use it.

## Explain
it runs three times in spring-cloud application.

#### *First. it runs with initialized web server in spring-boot.
it is started port.

log:`11917`
```
Tomcat initialized with port(s): 11917 (http)
```

#### *Second. it runs for get serverPort in spring-cloud-eureka-client
it is eureka registration center URL port.

EurekaClientAutoConfiguration.eurekaInstanceConfigBean():`11158`
```
int serverPort = Integer.valueOf(env.getProperty("server.port", env.getProperty("port", "8080")));
```

#### *Third. it runs for instanceId in spring-cloud-commons
it is eureka client instanceId.

IdUtils.getDefaultInstanceId():`11168`
```
String indexPart = resolver.getProperty("spring.application.instance_id",	resolver.getProperty("server.port"));
```

## Run Result
result log：
```
DiscoveryClient_BULLET/192.168.0.5:bullet:11168 - registration status: 204

Tomcat started on port(s): 11917 (http) with context path ''
```

eureka client instanceId:
```
UP (1) - 192.168.0.5:bullet:11168
```

eureka registration center URL port:
```
<a href="http://192.168.0.5:11584/actuator/info" target="_blank">192.168.0.5:bullet:11168</a>
```

## Influence

* we can't visited /info in eureka registration center.
* every application can't fetch real registry list.

## Solution
i rewrite the code in three project.

* spring-boot-autoconfigure
* spring-cloud-commons
* spring-cloud-netflix-eureka-client

i had pull request them.

* nayan3480232@163.com
* nayan3480232@outlook.com

# Thank you for reading all of this!

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -81,6 +81,7 @@ import static org.springframework.cloud.commons.util.IdUtils.getDefaultInstanceI
  * @author Matt Jenkins
  * @author Ryan Baxter
  * @author Daniel Lavoie
+ * @author Yan Na
  */
 @Configuration
 @EnableConfigurationProperties
@@ -146,7 +147,7 @@ public class EurekaClientAutoConfiguration {
 		EurekaInstanceConfigBean instance = new EurekaInstanceConfigBean(inetUtils);
 
 		instance.setNonSecurePort(serverPort);
-		instance.setInstanceId(getDefaultInstanceId(env));
+		instance.setInstanceId(getDefaultInstanceId(env,String.valueOf(serverPort)));
 		instance.setPreferIpAddress(preferIpAddress);
 		instance.setSecurePortEnabled(isSecurePortEnabled);
 		if (StringUtils.hasText(ipAddress)) {


spencergibb:
I highly doubt that boot will accept your change. This would have been a good one to discuss before submitting 3 separate pull requests. Setting a port to a random int from the random property source is a bad idea. If you set the server.port to 0, each runtime will get an unused port from the system, this is a better approach.

naah69:
server.port=0
but the serverId is same.
so wo can't access them ,when we run multi same application.

naah69:
this pull request also fix the same serviceId when we use server.port=0 and run multi same application.

naah69:
yes,they will customized.
but why not we make it be automatic?
the registration center also show them real port as serviceId

spencergibb:
Not all systems use port in the instanceId, it happens to be there and the instanceId is not the canonical source for the port. More than one port can be active in an app.

naah69:
when we use server.port=0
i runs two application.
image
image

but it show one application in  registration center and we can see it just one available by /eureka/apps
image
image

naah69:
we must give them different serviceId .them will work fine.
so if we always manual operation,it is inefficient.
we can run application freely，when we use flow control

spencergibb:
why do you want to run multiple servers on the same host?

naah69:
for dev and not all company have many server。
😁i just give a suggest for better develop

ExtremeYu:
that is what a meet

ExtremeYu:
@spencergibb Have the random port bug fixed？

naah69:
@ExtremeYu you can see my commit,althought it has been closed

ExtremeYu:
@naah69 i saw it. but when to release ?

naah69:
@ExtremeYu  no way,it has been closed ,they think that it is not important,are you chinese?

ExtremeYu:
@naah69 I followed you, don't you click my avatar and see wether  chinese am i or not ?

@spencergibb is the random port bug not important?

naah69:
@ExtremeYu  哈哈，感觉这名和头像就是出自我大天朝 这里不方便  加微信聊吧  17604510149

PR #2939, opened 2018-05-17
fix prefix header

MilovdZee:
When multiple Zuul routers are placed in series they all add x-forwarded-xxx headers. Most headers are seperated by comma's but the prefix header is just plainly concatenated resulting in useless values.

This patch adds the missing comma.

For example:

```
Header: 'x-forwarded-server' = '123.123.122.188, tst.abc.nl'
Header: 'x-forwarded-prefix' = '/rest/abc/v1/facades/recipes/search,/mobile-services/recipes'
Header: 'x-forwarded-port' = '80,8080'
Header: 'x-forwarded-host' = 'test.abc.nl,gateway:8080'
Header: 'x-forwarded-for' = '123.123.223.203, 123.123.67.20'
```

The comma in the x-forwarded-prefix value was missing

**Why?**
I have two Zuul proxies in series with rules like this:
```
zuul:
  routes:
    proxy1:
      path: /legacy/path/to/rest/search/**
      url: http://proxy2.abc.nl/new/path/to/rest/search
```
In the next proxy I have something like this:
```
zuul:
  routes:
    proxy2:
      path: /new/path/to/rest/search/**
      url: http://search.services.abc.nl/search
```
And the rest service in the search controller listens to /search?query=cheese
The search service must also return hateaos links and so needs to know the original link the request was sent to. In this case: https://api.abc.nl/legacy/path/to/rest/search?query=cheese
But when the prefixes are just concatenated that would result in: /legacy/path/to/rest/search/new/path/to/rest/search?query=cheese and so is completely useless.
I do need to know the original request.

Another thing is that I can't think of a situation that concatenating the prefixes is of any use... 
Issue with my pull request is that if somebody relies on this behaviour it would fail after the merge...

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
@@ -241,6 +241,7 @@ public class PreDecorationFilter extends ZuulFilter {
 				else {
 					newPrefixBuilder.append(prefix);
 				}
+				newPrefixBuilder.append(",");
 			}
 			newPrefixBuilder.append(route.getPrefix());
 			prefix = newPrefixBuilder.toString();
--- a/spring-cloud-netflix-zuul/src/test/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilterTests.java
+++ b/spring-cloud-netflix-zuul/src/test/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilterTests.java
@@ -273,7 +273,7 @@ public class PreDecorationFilterTests {
 		assertEquals("localhost", ctx.getZuulRequestHeaders().get("x-forwarded-host"));
 		assertEquals("80", ctx.getZuulRequestHeaders().get("x-forwarded-port"));
 		assertEquals("http", ctx.getZuulRequestHeaders().get("x-forwarded-proto"));
-		assertEquals("/prefix/api",
+		assertEquals("/prefix,/api",
 				ctx.getZuulRequestHeaders().get("x-forwarded-prefix"));
 		assertEquals("1.2.3.4, 5.6.7.8",
 				ctx.getZuulRequestHeaders().get("x-forwarded-for"));
@@ -314,7 +314,7 @@ public class PreDecorationFilterTests {
 		assertEquals("localhost", ctx.getZuulRequestHeaders().get("x-forwarded-host"));
 		assertEquals("80", ctx.getZuulRequestHeaders().get("x-forwarded-port"));
 		assertEquals("http", ctx.getZuulRequestHeaders().get("x-forwarded-proto"));
-		assertEquals("/context-path/api",
+		assertEquals("/context-path,/api",
 				ctx.getZuulRequestHeaders().get("x-forwarded-prefix"));
 		assertEquals("foo",
 				getHeader(ctx.getOriginResponseHeaders(), "x-zuul-serviceid"));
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@MilovdZee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2939)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2939) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@MilovdZee Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=2939)!

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
```diff
@@ -241,6 +241,7 @@ private void addProxyPrefix(RequestContext ctx, Route route) {
 				else {
 					newPrefixBuilder.append(prefix);
 				}
+				newPrefixBuilder.append(",");
```
ryanjbaxter: Wouldnt this result in a `x-forwarded-prefix` ending in a `,` if the headers wasnt already present?  I think a better approach would be to add the `,` to the end of the header value if it exists and is not already present in the code above.

MilovdZee: Nope, the line a little bit further in teh code always add the new prefix behind the added comma

spencergibb:
Can you explain why you think this is wrong? we made up `X-Forwarded-Prefix`

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/PreDecorationFilter.java
ryanjbaxter: Sure if the header already existed in the request, but if it didnt than it would just append it at the end.  

I wouldnt make an assumption that this situation is covered in a test somewhere

MilovdZee:
```
if (StringUtils.hasText(route.getPrefix())) {
  // So the route does have a prefix configured
  StringBuilder newPrefixBuilder = new StringBuilder();
  if (prefix != null) {
    // The prefix is also set
    if (prefix.endsWith("/") && route.getPrefix().startsWith("/")) {
      newPrefixBuilder.append(prefix, 0, prefix.length() - 1);
    }
    else {
      newPrefixBuilder.append(prefix);
    }
    // newPrefixBuilder now contains the prefix AND we know that the route does have a prefix
    newPrefixBuilder.append(",");
  }
  // Add the configures route prefix to the builder. Possibly containing an existing prefix appended with a comma
  newPrefixBuilder.append(route.getPrefix());
  prefix = newPrefixBuilder.toString();
}
```

MilovdZee:
The if-then-else:
```if (prefix.endsWith("/") && route.getPrefix().startsWith("/")) {```
Only makes sense if the prefixes are appended without the comma. What would be the use case for that??? When would appending the prefixes be useful? And besides that. It is not in sync with the behaviour of other x-forwarded headers.

MilovdZee:
Is there any progress? I see the label "for team discussion" but any estimate on when that will take place?

spencergibb:
I'll add it to the schedule

spencergibb:
Sorry for the delay here. I'm going to check to see if adding the comma breaks other spring projects integration with this header.

spencergibb:
Unfortunately, this breaks existing functionality in Spring Framework:

See https://github.com/spring-projects/spring-framework/blob/5e0cd9fb6074e73a7bd7909c56d9637d329ba0ed/spring-web/src/main/java/org/springframework/web/server/adapter/ForwardedHeaderTransformer.java#L129-L140

If you want to open an issue there and get that changed, we can revist here.

PR #2962, opened 2018-05-23
Log warning if getting a route throws an exception

chenkunyun:
fix https://github.com/spring-cloud/spring-cloud-netflix/issues/2956

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
@@ -68,7 +68,11 @@ public class SimpleRouteLocator implements RouteLocator, Ordered {
 		for (Entry<String, ZuulRoute> entry : getRoutesMap().entrySet()) {
 			ZuulRoute route = entry.getValue();
 			String path = route.getPath();
-			values.add(getRoute(route, path));
+			try {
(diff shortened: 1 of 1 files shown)


chenkunyun:
@ryanjbaxter 
Does the codecov work correctly? 
All the modification that I made to the code is just adding a try-catch, and the catch block can only be hit under the condition as I mentioned in https://github.com/spring-cloud/spring-cloud-netflix/issues/2956
Is there a place where I could write some test code to increase the code coverage to meet your standard(target 65.61%)?

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -68,7 +68,13 @@ public SimpleRouteLocator(String servletPath, ZuulProperties properties) {
 		for (Entry<String, ZuulRoute> entry : getRoutesMap().entrySet()) {
 			ZuulRoute route = entry.getValue();
 			String path = route.getPath();
-			values.add(getRoute(route, path));
+			try {
+				values.add(getRoute(route, path));
```
tdanylchuk: i guess it's better to split this lines into two 
```
ZuulRoute route  = getRoute(route, path)
values.add(route);
```
It would be more informative in case of exception.

spencergibb:
Merged with a test via 9287f446dd79da06f0eea02b08f21fff211bbdd1

PR #3001, opened 2018-06-10
Add support cluster query param in TurbineStream

yongsungyoon:
### Brief Description
Add support `cluster` parameter into Turbine Stream server.

### Background
Turbine Stream server doesn't support a `cluster` query parameter, unlike Turbine server until now. It means that we have to receive all Hystrix metrics at the same time from Turbine Stream server.  It is not a critical problem in the most environment. But in a production environment like us -  we have over hundreds of circuit breaks and dozens of thread pools - Turbine dashboard is frozen when we try to load the page because too many metrics are delivered to the browser at the same time and it makes our turbine dashboard unusable.

### Description
This PR adds the support of `cluster` parameter for Turbine Stream server.  So users can specify URL on their Turbine dashboard like below. 

```
http://my.turbine.sever:8989/turbine.stream?cluster=customers
```
If query parameter is not provided or `default` is provided, all metrics will be delivered and it is the default behavior of current release. 

The limitation is that user cannot configure how to extract cluster name. The implementation uses Eureka's serviceId as cluster name and these are not configurable. The limitation was described in the document. 

### Etc
The implementation has the limitation about how to define and extract cluster name from instances. 
Nevertheless, it might be very useful for users who are using Turbine Stream servers in the production environment like us. Moreover, the change doesn't have any side effect on user side because the default behavior without parameter is same as current release.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -726,6 +726,24 @@ You can then point the Hystrix Dashboard to the Turbine Stream Server instead of
 
 Spring Cloud provides a `spring-cloud-starter-netflix-turbine-stream` that has all the dependencies you need to get a Turbine Stream server running - just add the Stream binder of your choice, e.g. `spring-cloud-starter-stream-rabbit`. You need Java 8 to run the app because it is Netty-based.
 
+Turbine Stream server also supports the `cluster` parameter.
+Unlike Turbine server, Turbine Stream uses eureka serviceIds as cluster names and these are not configurable.
+
+If Turbine Stream server is running on port 8989 on `my.turbine.server` and you have two eureka serviceIds `customers` and `products`  in your environment, the following URLs will be available on your Turbine Stream server. `default` and empty cluster name will provide all metrics that Turbine Stream server receives.
+
+----
+http://my.turbine.sever:8989/turbine.stream?cluster=customers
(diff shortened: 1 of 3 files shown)


spencergibb:
Actually, `turbine-stream` has always been java 8. https://github.com/spring-cloud/spring-cloud-netflix/blob/1.4.x/spring-cloud-netflix-turbine-stream/pom.xml#L25-L26

ryanjbaxter:
Hmmm i wonder why CircleCI if failing then...
```
[ERROR] Failed to execute goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.14:check (enforce-java-7) on project spring-cloud-netflix-turbine-stream: Execution enforce-java-7 of goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.14:check failed. NullPointerException -> [Help 1]
```

spencergibb:
Looks like the .jdk8 files was accidentally removed here https://github.com/spring-cloud/spring-cloud-netflix/commit/bbac5a730f673f55f0ff86cb5f94fa21aecf135c

spencergibb:
@yongsungyoon please rebase against 1.4.x

yongsungyoon:
@spencergibb I've rebased the PR against 1.4.x.

ryanjbaxter:
@yongsungyoon I was not able to merge this forward into the 2.0.x branch, it was outside of my knowledge.  Would you mind submitting another PR against that branch with the correct changes?

Issue #3009, opened 2018-06-14
Adding support for IRule key in RibbonLoadBalancerClient


andersenleo:
Hi -

As far as I can tell it should be pretty straight forward (code-wise) to add support in `RibbonLoadBalancerClient` for propagating "key" for an `IRule` implementation. Currently the "key" concept seems to be missing. I'm *guessing* one of the reason is that the cloud-common `LoadBalancerClient` doesn't want to tie itself too hard into any implementation?

Our scenario is that we're implementing stateful client-side load-balancing where certain entities are "bound" to certain service instances and we're using the "key" to tell an `IRule` to find the service that owns that specific entity. We've recently started porting some of our micro-services from a homegrown Spring framework (where we do a manual setup of Ribbon and uses the `LoadBalancedCommand` to propagate the "key") to a Spring-Boot based solution. We would like to leverage as much as possible on what spring-cloud provides (such as discovery against Consul, for example).

The "dream scenario" for us would be to be able to do something like:

```java
WebClient client = WebClient.builder().baseUrl("http://STORE")
  .filter(lbFilter)
  .build();
client.post()
  .uri("/update/ID")
  .body(BodyInserters.fromObject("foo"))
  // Specify the key for the IRule
  .attribute("load-balancer-key", "ID")
  .retrieve()
  .bodyToMono(String.class)
  .single().block();
```

Where it's possible to pass on the "key" through the `ExchangeFilterFunction` for a load-balanced WebClient, or to simply call `LoadBalancerClient#execute("myService", "theKey", instance -> callWithRestTemplate(...))` (which is what the exchange-filter-function would be using behind the scene). 

So - why no access to the "key" in the `RibbonLoadBalancerClient`? And - would it make sense to work on a pull-request to get it in there? If we'd like to keep the `LoadBalancerClient` free from hard dependencies on the Ribbon model we could call it "hint" in the `LoadBalancer` API (similar to how `org.springframework.messaging.converter.AbstractMessageConverter` is done).

ryanjbaxter:
As far as the reason goes, you are probably right, we didnt want to tie `LoadBalancerClient` to Ribbon.  

Maybe instead of "hint" we can have a Map of metadata instead (allowing more flexibility).  Feel free to submit a PR with an idea of your proposal.

andersenleo:
[...]

PR #3012, opened 2018-06-15
Add 'hint' parameter to overloaded methods in RibbonLoadBalancerClient

andersenleo:
References #3009 

This is a prototype of how it would look. To maintain backward compatibility
with the `LoadBalancerClient` API the "hint" version of `choose` and `execute`
are overloaded with an extra parameter.

Next step would be to have a look at how the
`org.springframework.cloud.client.loadbalancer.reactive.LoadBalancerExchangeFilterFunction`
could be modified to take advantage of this change. The WebClient allows
attributes to be attached to a request - this could be used to pass along the
'hint' as part of the request:

```java
WebClient.create("http://STORE").get().uri("update").attribute(LoadBalancerClient.HINT, "123").retrieve()...
```


spencergibb:
Yeah, we're looking at a context object in the future, which would be a good place for this.

andersenleo:
@spencergibb interesting - I'm guessing that is being discussed in the scope if the commons `LoadBalancerClient`? 

I'm eager to help out if possible - as mentioned in #3009 we (my company) would be really interested in getting proper IRule support into the library in a not too far future :-) 

Is there anything written down on the reasoning and proposed design of the 'context'?

PR #3018, opened 2018-06-20
fixes permanently down after refresh

### LINKED ISSUE #2972: Eureka client permanently down after refresh
Issue most likely related to several, but I want emphasize the root problem here (IMHO).
The very last issue that I've found is https://github.com/spring-cloud/spring-cloud-netflix/issues/1857 .

**Pre-requirements:**
cloud version - `Edgware.SR3`
1. setup simplest eureka server
2. setup simplest boot app only with eureka-client dep 
with properties `eureka.instance.initialStatus=DOWN` and `eureka.client.healthcheck.enabled=true`

**Steps to reproduce:**
1. Start service 
- Service registers in eureka with `DOWN` status, since it is initial status.
- `InstanceInfoReplicator` updates status to `UP` as it now uses `EurekaHealthCheckHandler` which was registered by `EurekaAutoServiceRegistration`, but not going to `InstanceInfo` object to get current `DOWN` state
2. Calling /refresh endpoint
- `CloudEurekaClient` bean recreated and re-registers instance in Eureka with initial status `DOWN`
- `InstanceInfoReplicator` keeping re-registering instance with `DOWN` status as it uses empty `HealthCheckCallbackToHandlerBridge` which gets current `DOWN` state from  `InstanceInfo`, but not from `EurekaAutoServiceRegistration`, from which it should.

**Root cause:**
`EurekaHealthCheckHandler` uses `EurekaRegistration` bean as provider for `EurekaClient` object.
Since it's not being recreated it stores old `CloudEurekaClient` and `EurekaHealthCheckHandler` sets to inactive bean (which was dropped of spring context due to refresh scope) instead of newly created one.
It can be easily seen in debug mode, it shows different objects hashcodes in spring context and in `EurekaRegistration` bean.
It actually was the root cause of https://github.com/Netflix/eureka/issues/1025 since it tried to call `onDemandUpdate` on shut down client.

**Possible solutions:**
- Make `EurekaRegistration` bean refresh scoped and use bean `Provider` in `EurekaAutoServiceRegistration` to retrieve bean from context after refresh
- Rewrite `EurekaRegistration.getEurekaClient()` to get `CloudEurekaClient` bean from context, not cached one.

**Why does it actually that matter?**
This issue was possible to find only using `eureka.instance.initialStatus=DOWN` property, but in case of default `UP` you won't be able to see, that after refresh (manual/bus) service won't be `DOWN` anymore, even if some health indicators will go `DOWN`.

tdanylchuk:
root cause:
cached bean trying to re-register
resolution summary:
get re-created bean from context to re-register

Fixes #2972

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaRegistration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaRegistration.java
@@ -44,7 +44,7 @@ import com.netflix.discovery.EurekaClientConfig;
 /**
  * @author Spencer Gibb
  */
-public class EurekaRegistration implements Registration, Closeable {
+public class EurekaRegistration implements Registration {
 	private static final Log log = LogFactory.getLog(EurekaRegistration.class);
 
 	private final EurekaClient eurekaClient;
@@ -201,8 +201,4 @@ public class EurekaRegistration implements Registration, Closeable {
 		return this.instanceConfig.getSecurePort();
 	}
 
-	@Override
-	public void close() throws IOException {
-		this.eurekaClient.shutdown();
-	}
 }
--- a/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
+++ b/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
@@ -16,27 +16,32 @@
 
 package org.springframework.cloud.netflix.eureka;
 
-import java.io.IOException;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.junit.After;
 import org.junit.Test;
 import org.mockito.Mockito;
+
+import org.springframework.aop.framework.Advised;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.SearchStrategy;
 import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.autoconfigure.RefreshAutoConfiguration;
+import org.springframework.cloud.client.serviceregistry.AutoServiceRegistrationProperties;
 import org.springframework.cloud.commons.util.UtilAutoConfiguration;
+import org.springframework.cloud.context.refresh.ContextRefresher;
 import org.springframework.cloud.context.scope.GenericScope;
-import org.springframework.cloud.netflix.eureka.serviceregistry.EurekaRegistration;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.test.util.ReflectionTestUtils;
 
 import com.netflix.appinfo.ApplicationInfoManager;
+import com.netflix.appinfo.HealthCheckHandler;
 import com.netflix.discovery.EurekaClient;
 import com.netflix.discovery.EurekaClientConfig;
 import com.netflix.discovery.shared.transport.jersey.EurekaJerseyClient;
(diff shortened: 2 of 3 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
```diff
@@ -52,41 +54,44 @@
 
 	private EurekaServiceRegistry serviceRegistry;
 
-	private EurekaRegistration registration;
+	private ObjectProvider<EurekaRegistration> registrationProvider;
```
spencergibb: Did you try without object provider? Refresh scope creates a proxy which should handle getting the refreshed instance for you.

tdanylchuk:
I'll port fix to master soon

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaRegistration.java
```diff
@@ -201,8 +201,4 @@ public int getSecurePort() {
 		return this.instanceConfig.getSecurePort();
 	}
 
-	@Override
-	public void close() throws IOException {
```
spencergibb: I worry that this will have negative sure effects: it won't deregister on a normal shutdown.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
```diff
@@ -425,16 +466,6 @@ public void eurekaClientClosed() {
 		}
 	}
 
-	@Test
-	public void eurekaRegistrationClosed() throws IOException {
```
spencergibb: This test can't be removed as it is a regression

spencergibb:
Don't port, we merge forward if accepted.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaRegistration.java
tdanylchuk: since `DiscoveryClient` is bean itself with marked destroy method, I don't think we need close it twice.
BTW saw exception in logs related this before Egware.3 (when exception was thrown but not muted)

Anyway I've wrote test for this, you may take a look.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
tdanylchuk: test is being moved to ```shouldCloseDiscoveryClient``` since `EurekaRegistration` is not closing DiscoveryClient anymore

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
tdanylchuk: Khmm, will try, was thinking only about lazy init as proxy, but not refreshed bean

tdanylchuk: works, fine, reverted class change

tdanylchuk:
can't get why it's failing, locally works well and changes should not affect test functionality which is failing...

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaRegistration.java
ryanjbaxter: So where will we shutdown the client otherwise?

tdanylchuk: we don't , spring context will shutdown it for us
```
@Bean(destroyMethod = "shutdown")
@ConditionalOnMissingBean(value = EurekaClient.class, search = SearchStrategy.CURRENT)
public EurekaClient eurekaClient(ApplicationInfoManager manager, EurekaClientConfig config) {
	return new CloudEurekaClient(manager, config, this.optionalArgs, this.context);
}
```
Here is shutdown method `@Bean(destroyMethod = "shutdown")`

ryanjbaxter: got it thanks

tdanylchuk:
@ryanjbaxter can we port it to master branch?

ryanjbaxter:
I merged it forward, you might want to double check the merge because it wasnt pretty.  All the tests were passing for me locally.

tdanylchuk:
I see some unrelated changes to my PR at https://github.com/spring-cloud/spring-cloud-netflix/commit/13c2500d0ea77734223f3bcac4bdf0863a0fd3b5 - you want me to polish them?

ryanjbaxter:
The unrelated changes are coming from other PRs that were merged forward

PR #3019, opened 2018-06-21
Tiny things

holy12345:
Background

In `SendErrorFilter` source code have a comment that say `move SendErrorFilter in error package` so
I create a error package and move it.


holy12345:
CI test is failed, the error message is `EurekaHealthCheckTests.shouldRegisterService()` test failed. but I run the test code on my laptop is ok and I think the code what I modify is not related with the `EurekaHealthCheckTests` code.

Issue #3038, opened 2018-06-28
Ribbon PoolKeepAliveTime is never read


arkin:
Because the ribbon config entry PoolKeepAliveTime is read as a string instead of an integer/long, it never passes the check at https://github.com/spring-cloud/spring-cloud-netflix/blob/1.4.x/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/HttpClientRibbonConfiguration.java#L90
Due to this issue, one cannot override the default value of 900 seconds.

arkin:
Looks like the same issue exists on master, too, although the code was moved to another class.

spencergibb:
This was fixed in the 1.4.x line by #3039.

spencergibb:
Looks like there was a regression when property access was moved.

PR #3039, opened 2018-06-28
Fixes Ribbon Apache Client Configuration Effectively Ignoring PoolKeepAliveTime

arkin:
Due to type mismatching (String vs Long) the PoolKeepAliveTime config was being ignored.
Unfortunately, the default and key definition still have a type mismatch (long vs Integer) outside this project.
This fix makes the assumption that the default will stay within integer boundaries.
Fixes gh-3038.

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/HttpClientRibbonConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/HttpClientRibbonConfiguration.java
@@ -81,22 +81,23 @@ public class HttpClientRibbonConfiguration {
 			Integer timerRepeat = config.getPropertyAsInteger(
 					CommonClientConfigKey.ConnectionCleanerRepeatInterval,
 					DefaultClientConfigImpl.DEFAULT_CONNECTION_IDLE_TIMERTASK_REPEAT_IN_MSECS);
-			// Since we cannot change the type of CommonClientConfigKey.PoolKeepAliveTime from Integer to Long,
-			// we are working with integers.
-			// Unfortunately, the default is defined as long in the same package whereas the config key is using
-			// Integer.
-			Integer timeToLive = config.get(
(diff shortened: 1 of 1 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/HttpClientRibbonConfiguration.java
```diff
@@ -81,21 +81,22 @@ public HttpClientConnectionManager httpClientConnectionManager(
...
+			// we are working with integers.
+			// Unfortunately, the default is defined as long in the same package whereas the config key is using
+			// Integer.
+			Integer timeToLive = config.get(
+					CommonClientConfigKey.PoolKeepAliveTime,
+					(int)DefaultClientConfigImpl.DEFAULT_POOL_KEEP_ALIVE_TIME);
 			Object ttlUnitObj = config
 					.getProperty(CommonClientConfigKey.PoolKeepAliveTimeUnits);
 			TimeUnit ttlUnit = DefaultClientConfigImpl.DEFAULT_POOL_KEEP_ALIVE_TIME_UNITS;
-			if (timeToLiveObj instanceof Long) {
```
ryanjbaxter: Can we just check if it is a String and then convert the string to a Long?

arkin: Sure.

arkin:
@ryanjbaxter I implemented it the way I think you suggested. Are you ok with propagating exceptions? I thought a typo/error in the config should stop initialization. What do you think?

arkin:
@ryanjbaxter I noticed that you released 1.4.5 recently and it will probably be a while until the next release comes along that includes this. Is there a workaround you can think of that we can use until then?

ryanjbaxter:
Yes 1.4.5 is in the process of being released.  This will be merged forward into the 2.0.x stream and will be included in Finchley.SR1.  I would say we would probably have another Edgware release in the future just no sure when.

Issue #2965, opened 2018-05-23
Return HTTP status code 504 for Hystrix timeout in Zuul proxy


taasjord:
For Hystrix timeout, Zuul proxy always returns HTTP status 500. From the client side, this looks like an application error more than a proxy error. I suggest to use more specific HTTP status code 504 "Gateway Timeout".

This will allow the client to be more intelligent on how to handle errors, for example be careful about resubmitting a POST when getting 504.

Edit: it looks like rejected and short-circuit does return 503 "Service Unavailable" and timeouts from Ribbon client does in fact return 504, so it remains the Hystrix circuit timeout that now does return 500  Internal Server Error instead of 504 - and the 500 metrics are reported on the /error path rather than on actual path.

PR #3060, opened 2018-07-06
Return HTTP status code 504 for Hystrix timeout in Zuul proxy (issue 2965)

dev-priporov:
This is an enhancement for [this one](https://github.com/spring-cloud/spring-cloud-netflix/issues/2965)

Perhaps will be better to fix that inside _zuul-core_ because my solution isn't what I'm proud, but I tested that and it works.


spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendErrorFilter.java
```diff
@@ -100,10 +103,18 @@ public Object run() {
 
 	protected ExceptionHolder findZuulException(Throwable throwable) {
 		if (throwable.getCause() instanceof ZuulRuntimeException) {
+			Throwable cause = throwable.getCause().getCause().getCause();
```
ryanjbaxter: I am not to fond of this type of code as it could result in a NPE, I would want some checks here to protect us.

dev-priporov:
@ryanjbaxter 
fixed

Issue #3063, opened 2018-07-07
EurekaHealthIndicator not filter out service without instance


addozhang:
version: spring-cloud-netflix-eureka-client-1.4.4.RELEASE 

We are doing some performance on eureka by registering a lot of services. After services evicted, we found those evicted services still listed in `discoveryComposite.eureka` section of `/health` endpoint. But not in `discoveryComposite.discoveryClient`. It increases payload of `/health`.

Compared with `DiscoveryClientHealthIndicator`, `EurekaHealthIndicator` does not filter out those services without instances.

![image](https://user-images.githubusercontent.com/2224492/42407300-17838f28-81ec-11e8-9dbc-ac697b8a02fa.png)

In `EurekaHealthIndicator. getApplications()`, adding three lines will filter them out.

```java
	private Map<String, Object> getApplications() {
		Applications applications = this.eurekaClient.getApplications();
		if (applications == null) {
			return Collections.emptyMap();
		}
		Map<String, Object> result = new HashMap<>();
		for (Application application : applications.getRegisteredApplications()) {
			if (application.getInstances().isEmpty()) {
				continue;
			}
			result.put(application.getName(), application.getInstances().size());
		}
		return result;
	}
```

addozhang:
One more thing, Eureka client still holds those applications who has no any instances in `DiscoveryClient.localRegionApps`. The only way to delete them I found is setting `eureka.client.log-delta-diff` as true. But this approach will add one more full fetch request after delta request. 

In method `updateDelta()`, the logic of handling instance deleted is confused. Even not apps exists, it still adds app back, and does removing op again. 

I think the correct logic is if no app exists, do nothing. If exists, remove instance from it and **check if it has any instance up. If no, remove app.**

```java
if (ActionType.DELETED.equals(instance.getActionType())) {
    Application existingApp = applications.getRegisteredApplications(instance.getAppName());
    if (existingApp == null) {
        applications.addApplication(app);
    }
    logger.debug("Deleted instance {} to the existing apps ", instance.getId());
    applications.getRegisteredApplications(instance.getAppName()).removeInstance(instance);
}
```

ryanjbaxter:
@addozhang would you like to submit a PR for `EurekaHealthIndicator`?

You should open an issue in Netflix Eureka for the `DiscoveryClient` issue.

ryanjbaxter:
closed via 965fbf8

PR #3068, opened 2018-07-10
Filter out service without instances from health status

addozhang:
PR for #3063.


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaHealthIndicator.java
```diff
@@ -97,6 +97,9 @@ else if (lastFetch > clientConfig.getRegistryFetchIntervalSeconds() * 2000) {
 		}
 		Map<String, Object> result = new HashMap<>();
 		for (Application application : applications.getRegisteredApplications()) {
+			if (application.getInstances().isEmpty()) {
+				continue;
```
spencergibb: I'd prefer if not empty put rather than continue.

addozhang: You r right. thx

PR #3079, opened 2018-07-17
fixes #3053

### LINKED ISSUE #3053: Zuul retry and ServerStats
I came across some problems while testing Zuul retry logic. Acording to my tests ServerStats.successiveConnectionFailureCount is increased only in RibbonLoadBalancedRetryPolicy.updateServerInstanceStats. I've never seen ServerStats.clearSuccessiveConnectionFailureCount() being executed. The same goes for ServerStats.activeRequestsCount. It seems to me that the methods: noteResponse, noteRequestCompletion from LoadBalancerContext class are never executed. Could you please tell me if something is wrong with my config?

My config:
Springboot 1.5.13.RELEASE with Spring Cloud Edgware.SR3 and BestAvailableRule as a load balancing strategy.

I have spring-retry on my classpath.

```
zuul:
  health.enabled: true
  ribbonIsolationStrategy: SEMAPHORE
  retryable: true
  ignoredServices: '*'
  routes:
    consumer:
      path: /consumer/**
      serviceId: consumer
      stripPrefix: false

consumer:
  ribbon:
    ConnectTimeout: 5000
    ReadTimeout: 60000
    MaxAutoRetriesNextServer: 3
    MaxAutoRetries: 0
    OkToRetryOnAllOperations: true

hystrix:
  command:
    default:
      circuitBreaker:
        enabled: true
      execution:
        timeout:
          enabled: true

```

pway99:
fixes https://github.com/spring-cloud/spring-cloud-netflix/issues/3053
Retryable apache and okhttp clients utilize `RibbonStatsRecorder` for updating `ServerStats` upon successful requests

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
@@ -17,6 +17,8 @@ package org.springframework.cloud.netflix.ribbon.apache;
 
 import java.net.URI;
 import org.apache.commons.lang.BooleanUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpUriRequest;
@@ -63,6 +65,8 @@ public class RetryableRibbonLoadBalancingHttpClient extends RibbonLoadBalancingH
 	private LoadBalancedRetryListenerFactory loadBalancedRetryListenerFactory =
 		new LoadBalancedRetryListenerFactory.DefaultRetryListenerFactory();
 	private RibbonLoadBalancerContext ribbonLoadBalancerContext;
+	
+	private static final Log LOGGER = LogFactory.getLog(RetryableRibbonLoadBalancingHttpClient.class);
 
 	@Deprecated
 	//TODO remove in 2.0.x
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
```diff
@@ -132,6 +132,11 @@ public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws Excepti
 							.port(service.getPort()).path(newRequest.getURI().getPath())
 							.query(newRequest.getURI().getQuery()).fragment(newRequest.getURI().getFragment())
 							.build().encode().toUri());
+					
+					if (service instanceof RibbonServer) {
+						RibbonLoadBalancerContext rctx = new RibbonLoadBalancerContext(getLoadBalancer());
```
ryanjbaxter: Shouldnt we be using the bean here rather than creating a new instance?

pway99: yes using the bean is better.  I have updated the code accordingly.
thanks!

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java
```diff
@@ -132,6 +133,10 @@ public RibbonApacheHttpResponse doWithRetry(RetryContext context) throws Excepti
 							.port(service.getPort()).path(newRequest.getURI().getPath())
 							.query(newRequest.getURI().getQuery()).fragment(newRequest.getURI().getFragment())
 							.build().encode().toUri());
+					
+					if (service instanceof RibbonServer) {
+						statsRecorder = new RibbonStatsRecorder(ribbonLoadBalancerContext, ((RibbonServer)service).getServer());
```
ryanjbaxter: is it ok if `ribbonLoadBalancerContext` is null?  We might need a null check here because there is no guarantee the setter is called

PR #3096, opened 2018-07-21
Update RibbonLoadBalancedRetryPolicy.java

deepakagrawal121:
//reset same server since we are moving to a new server
sameServerCount = 0;

This reset will make retry to next server equals to retry on same server.

Let's have a example:

RetrySameServer:3
RetryNextServer:1

And we are done with retry on same server.

then condition if(sameServerCount >= lbContext.getRetryHandler().getMaxRetriesOnSameServer() && canRetry(context)) will set to true and sameServerCount = 0; will set to true.

And it will keep retying until sameServerCount does not match. Or If I set to same MaxRetryOnNextServer to 2 it will retry on next server 7 times.

### DIFF
--- a/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
+++ b/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonLoadBalancedRetryPolicy.java
@@ -117,8 +117,6 @@ public class RibbonLoadBalancedRetryPolicy implements LoadBalancedRetryPolicy {
 		//until we actually equal the same server count limit.  This will allow us to make the initial
 		//request plus the right number of retries.
 		if(sameServerCount >= lbContext.getRetryHandler().getMaxRetriesOnSameServer() && canRetry(context)) {
-			//reset same server since we are moving to a new server
-			sameServerCount = 0;
 			nextServerCount++;
 			if(!canRetryNextServer(context)) {
 				context.setExhaustedOnly();


pivotal-issuemaster:
@deepakagrawal121 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3096)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3096) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@deepakagrawal121 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3096)!

ryanjbaxter:
The build is failing with this change.

I also dont completely understand what the problem is that you are solving.  Can you elaborate?

A unit test to demonstrate the problem is also required.

ulmermark:
I was looking at this exact same code today and seeing strange behavior as well.

The number of retries seemed to be "double" what was expected

My scenario was running a single instance of a Microservice registered with Eureka

My request were going through Zuul and using a route based on a serviceId

ribbon.MaxAutoRetries= 1
ribbon.MaxAutoRetriesNextServer= 1
zuul.retryable=true

I set the microservice up to throw a 502 which I had said was a retryableStatusCode

The initial request would fail.  Then retry to the original server and set sameServerCount = 1, then retry again and 502  failure.  

sameServerCount >= MaxAutoRetries, so "switch" to next server, and setNextServerCount = 1, and sameServerCount = 0.

The retry again with failure, with sameServerCount = 0, then setSameServerCount = 1.

The next failure would check sameServerCouint++ to MaxAutoRetries, which would cause another retry,  When this retry failed, then sameServerCount and nextServerCount were finally larger than the retries configured.  Causing the retries to be exhausted.

This would result in 1 original call + 3 retries . instead of 1 + 2 retries. 

 If MaxAutoRetries was 2, you would then see 1 + 2 + 2 + 1 = 6 total tries.

spencergibb:
We're cleaning out the issue tracker and closing issues that we've not seen much demand to fix. Feel free to comment with additional justifications if you feel that this one should not have been closed.

Issue #3122, opened 2018-08-03
Allow to disable response decompression


tjuchniewicz:
When using Apache Http Client, response received from backed service is always decompressed. This is feature of Apache Http Client. Content is decompressed and `Content-Encoding` header is removed from response: [ResponseContentEncoding](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/protocol/ResponseContentEncoding.java#L125). Zuul processes decompressed content.

This make no sense for Zuul in all cases because we forward response to Zuul client and there is no reason to decompress and compress again if Zuul client requested compression. Maybe one wants to do some content based filtering and decompressed content is expected.

After Apache Http Client removes  `Content-Encoding` header all Zuul calls like:  [context.getResponseGZipped()](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java#L138) will always return `false`.

My proposal is to disable automatic decompression in Apache in: [RetryableRibbonLoadBalancingHttpClient ](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RetryableRibbonLoadBalancingHttpClient.java#L78):
using: 
```java
builder.setContentCompressionEnabled(false);
```
or use property to control this behavior.

tjuchniewicz:
@brenuart I see you may be interested in gzip issues.

ryanjbaxter:
I think this makes sense, I think having a property to control this makes sense.

tjuchniewicz:
I will prepare PR for this.

tmack8001:
This is similar to what I have done just wrapped up nicely in a property instead of a custom.http connection manager and related http clients.

PR #3134, opened 2018-08-12
Allow to disable compression in Apache HttpClient

tjuchniewicz:
Fixes gh-3122

### DIFF
--- a/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonProperties.java
+++ b/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonProperties.java
@@ -22,7 +22,11 @@ import static com.netflix.client.config.DefaultClientConfigImpl.DEFAULT_PORT;
 import static com.netflix.client.config.DefaultClientConfigImpl.DEFAULT_READ_TIMEOUT;
 
 public class RibbonProperties {
-    private final IClientConfig config;
+	private final IClientConfig config;
+
+	// move to DefaultClientConfigImpl in ribbon-core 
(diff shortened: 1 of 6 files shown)


spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/RibbonProperties.java
```diff
@@ -22,7 +22,11 @@
 import static com.netflix.client.config.DefaultClientConfigImpl.DEFAULT_READ_TIMEOUT;
 
 public class RibbonProperties {
-    private final IClientConfig config;
+	private final IClientConfig config;
```
ryanjbaxter: indentation is off

tjuchniewicz: Fixed. RibbonProperties use different indentation than other files.

tjuchniewicz:
I will prepare test.
Regarding documentation. I've implemented `CommonClientConfigKey.GZipPayload'  from `ribbon-core` so we should find documentation there... but I can't see description of this config key.

tjuchniewicz:
Unit tests added.

ryanjbaxter:
Should we also make a similar change in `RibbonLoadBalancingHttpClient` in the case that Spring Retry is not on the classpath?

tjuchniewicz:
`RibbonLoadBalancingHttpClient` is already updated.

spencergibb:
Squashed and merged with polish via 0f7c64325e009941b49d5b0136ae6b72efe8a184

tjuchniewicz:
@spencergibb Thank you.

PR #3136, opened 2018-08-13
Fix invalid URI error when Feign request URI does not end in /

ryanjbaxter:


ryanjbaxter:
Was also fixed in Spring Cloud OpenFeign with https://github.com/spring-cloud/spring-cloud-openfeign/commit/d57cea265eedf94451d6824c5e6cbaa89a8cf549

PR #3165, opened 2018-08-29
WIP (Do Not Merge): Adds support for netflix concurrency limits

spencergibb:
See https://github.com/Netflix/concurrency-limits/ for explanation of the library.

This adds support for spring mvc, webflux and (shortly) micrometer.

@twicksell @elandau @jkschneider

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,6 +148,7 @@
 		<!-- Not part of the reactor build: build and deploy separately -->
 		<!--module>spring-cloud-netflix-hystrix-contract</module-->
 		<module>spring-cloud-netflix-core</module>
+		<module>spring-cloud-netflix-concurrency-limits</module>
 		<module>spring-cloud-netflix-hystrix-dashboard</module>
 		<module>spring-cloud-netflix-hystrix-stream</module>
 		<module>spring-cloud-netflix-eureka-client</module>
--- a/spring-cloud-netflix-concurrency-limits/pom.xml
+++ b/spring-cloud-netflix-concurrency-limits/pom.xml
@@ -44,11 +44,16 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-webflux</artifactId>
-			<scope>test</scope>
+			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-test</artifactId>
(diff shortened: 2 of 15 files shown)


elandau:
@spencergibb Looks awesome!  

As an FYI, I'm not too happy with the contract for collecting metrics and the integration between the Limiter and Strategy classes.  I was planning on doing some refactoring in that space but will definitely make sure not to break this integration.  This would most likely only affect how the Limiter is constructed.

spencergibb:
If refactoring gets done before we go GA, it wouldn't be a problem

spring-cloud-netflix-concurrency-limits/src/main/java/org/springframework/cloud/netflix/concurrency/limits/reactive/ConcurrencyLimitsWebFilter.java
```diff
@@ -0,0 +1,52 @@
...
+
+	private final Limiter<ServerWebExchange> limiter;
+
+	public ConcurrencyLimitsWebFilter(Limiter<ServerWebExchange> limiter) {
+		this.limiter = limiter;
+	}
+
+	@Override
+	public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
+		Optional<Limiter.Listener> listener = limiter.acquire(exchange);
```
elandau: You may want to keep the code within the functional style so there's no need to call isPresent() multiple times.
```
return limiter.acquire(exchange) 
    .map(listener -> chain.filter(exchange)
        .doOnSuccess(v -> listener.onSuccess())
        .doOnError(throwable -> listener.onIgnore()))
    .orElseGet(() -> {
        exchange.getResponse().setStatusCode(HttpStatus.TOO_MANY_REQUESTS);
	//TODO: set body
	return exchange.getResponse().setComplete();
    });
```

cbornet:
Hi,

Sorry to comment on a closed issue. Please tell me if there is a better channel for this kind of discussion (SoF ?)
I just wanted to know your opinion for this with Webflux. From what I understand, Webflux applies backpressure on the TCP connexion that propagates it through its own congestion mechanism.
So does Netflix's concurrency-limit add any value in the Webflux context ?

spencergibb:
Gitter or stack overflow

cbornet:
Thanks. I posted the question on https://gitter.im/spring-cloud/spring-cloud

PR #3178, opened 2018-09-06
DO NOT MERGE - for discussion - Ordered discovery client with issues

OlgaMaciaszek:


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClient.java
```diff
@@ -42,11 +41,11 @@
 
 	public static final String DESCRIPTION = "Spring Cloud Eureka Discovery Client";
 
-	private final EurekaInstanceConfig config;
+	private final EurekaInstanceConfigBean config;
 
 	private final EurekaClient eurekaClient;
 
-	public EurekaDiscoveryClient(EurekaInstanceConfig config, EurekaClient eurekaClient) {
+	public EurekaDiscoveryClient(EurekaInstanceConfigBean config, EurekaClient eurekaClient) {
```
OlgaMaciaszek: If anyone has a different implementation of `EurekaInstanceConfig` in the classpath, this will cause exception, but we would have to change it this way to call `getOrder()` from `EurekaInstanceConfig` and not another property bean (like I did before), cause `EurekaInstanceConfig` is not ours and we can't add `getOrder()` method there; I could switch to a new interface extending `EurekaInstanceConfig`, but this still would not be backwards compatible.

spencergibb: Make `EurekaInstanceConfigBean` implemented `Ordered`, then do an instance of check so you don't have to rely on `EurekaInstanceConfigBean`

PR #3202, opened 2018-09-20
Use the appname and id from in mem instanceInfo object instead of raw config

qiangdavidliu:
As one example, when running in local dev with registration disabled, the instance id from the raw config may be null. This is guaranteed to not be null if read from the local InstanceInfo.


ryanjbaxter:
Could you submit this against the 2.0.x branch?

qiangdavidliu:
Will do.

ryanjbaxter:
Going to close this PR in favor of 3212.  We can merge that forward.

Issue #3175, opened 2018-09-04
healthCheckUrl and statusPageUrl are invalid when using sidecar.


wsvoorhees:
Bug report

From a customer:

The trouble I'm having is with the <healthCheckUrl> reported in Eureka. It is being reported as the port of the pod in kubernetes. When I attempt to specify the correct IP and port that is accessible from outside of the kubernetes cluster using the spring property sidecar.health-uri the sidecar will report the status as DOWN and still reports the <healthCheckUrl> in Eureka using the pod port rather than the URL I provided in the spring property.

This feels like a bug since I'm providing the correct health check URL, but maybe there is some understanding into how it is intended to be used.

I'm requesting the version numbers being used, but I believe it's finchley.

scottlindner:
I am setting the following:
eureka.instance.prefer-ip-address: true
sidecar.port: 1234
sidecar.hostname: 123.4.5.6
sidecar.ip-address: 123.4.5.6
And this works and reports health UP

But when I add:
sidecar.health-uri: 1234.4.5.6:1234/actuator/health
It will report health as DOWN and the entire sidecar will crash if I try hitting any other actuator ports.

From looking at the code, it looks like the property is being used by the sidecar to check health rather than what is being sent to eureka.

We may be hitting a use case issue for the sidecar. I'm trying to use it in a kubernetes cluster combined with applications deployed in Pivotal Cloud Foundry. The internal kube network isn't accessible to the apps in PCF so I'm overriding them in the sidecar to be the nodePort and nodeIP that is accessible to PCF. This works for the hostname and port, but not for the health-uri.

ryanjbaxter:
So are you trying to set the health check url that will be registered with the Eureka server?

ryanjbaxter:
Did you try this http://cloud.spring.io/spring-cloud-static/Finchley.SR1/single/spring-cloud.html#_status_page_and_health_indicator?

scottlindner:
Thanks for that reference. I gave it a try in a couple of different ways. It doesn't seem to do anything to affect this.

Let me better try to depict what I'm doing and why so you can get a better visual for it. The sidecar is running in a container that is on a different network than the services that will be accessing it including eureka. This is because I'm accessing services deployed in kubernetes from a Cloud Foundry application server outside of the kubernetes network.

The sidecar is running in a kubernetes pod at http://10.200.43.41:5678

[...]

PR #3213, opened 2018-09-26
Sidecar: add accept-all-certificates property in order to be able to use https healthchecks with self-signed certificates

Hyunk3l:
Hello guys!

Adding a new property I think can be interesting :)
Actually I had to do a workaround (https://github.com/spring-cloud/spring-cloud-netflix/issues/3175#issuecomment-420276337 thanks @spencergibb ) in a project I'm working on, but I think that [changing the way `RestTemplate` is created](https://www.baeldung.com/httpclient-ssl), we can make Sidecar work with https healthchecks, with self-signed certificates.

Don't know if it's useful for you. What do you think?

## Stuff I'd like to improve in the case the feature is OK
* Default value for `accept-all-certificates`. How can I set it to `false`?
* Add Unit test to `LocalApplicationHealthIndicator`


pivotal-issuemaster:
@Hyunk3l Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3213)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3213) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Hyunk3l Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3213)!

spencergibb:
Part of the problem is that Apache http client is not a required dependency.

ryanjbaxter:
Maybe you can do some of this configuration in a configuration class where you can figure out if the Apache Http Client is on the classpath.

Hyunk3l:
Hello @ryanjbaxter @spencergibb 
Moving this block https://github.com/spring-cloud/spring-cloud-netflix/pull/3213/files to a new "config" class named `RestTemplateConfiguration` is ok?
Furthermore, about checking Apache HttpClient in classpath, is ok to implement a try/catch block like this https://github.com/spring-cloud/spring-cloud-netflix/blob/2b3235e6c9e20e3ab768c81701bc2a403701d3db/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/PropertiesFactory.java#L51 ?

thanks :)

ryanjbaxter:
I would probably make the `HealthCheckHandler` creation in `SidecarConfiguration` conditional on the presence of the `ApacheHttpClient`.  Then have one for when the class is not present.  You can use `@ConditionalOnClass(ApacheHttpClient.class)`

spring-cloud-netflix-sidecar/src/main/java/org/springframework/cloud/netflix/sidecar/SidecarConfiguration.java
```diff
@@ -153,7 +160,24 @@ public HealthCheckHandler healthCheckHandler(
 				final LocalApplicationHealthIndicator healthIndicator) {
 			return new LocalApplicationHealthCheckHandler(healthIndicator);
 		}
+	}
 
+	@Bean
```
ryanjbaxter: This would need an `@ConditionalOnMissingClass`

Hyunk3l:
image

Having some issues while rebasing to `2.0.x`
I'll open a new PR, I think it's faster. Sorry for any inconvenience.

PR #3224, opened 2018-10-10
Sidecar: add accept-all-ssl-certificates property in order to be able to use https health checks with self-signed certificates

Hyunk3l:
Hello guys!

Adding a new property I think can be interesting :)
Actually I had to do a workaround (https://github.com/spring-cloud/spring-cloud-netflix/issues/3175#issuecomment-420276337 thanks @spencergibb ) in a project I'm working on, but I think that [changing the way `RestTemplate` is created](https://www.baeldung.com/httpclient-ssl), we can make Sidecar work with https healthchecks, with self-signed certificates.

Don't know if it's useful for you. What do you think?

## Stuff I'd like to improve in the case the feature is OK
* Default value for `accept-all-certificates`. How can I set it to `false`?
* Add Unit test to `LocalApplicationHealthIndicator`

PS: PR https://github.com/spring-cloud/spring-cloud-netflix/pull/3213 closed because of a problem while rebasing :/

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -2021,6 +2021,17 @@ info:
   url: https://github.com/spring-cloud-samples
 ----
 
+Using https health check with self-signed certificates:
+
+.application.yml
+[source,yaml]
+----
+sidecar:
+  port: 8000
+  health-uri: https://localhost:8000/health.json
+  accept-all-ssl-certificates: true
+----
+
 [[retrying-failed-requests]]
 == Retrying Failed Requests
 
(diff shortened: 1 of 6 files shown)


ryanjbaxter:
Sounds good, also need to add documentation

Hyunk3l:
@ryanjbaxter thanks! Done!

docs/src/main/asciidoc/spring-cloud-netflix.adoc
```diff
@@ -2021,6 +2021,17 @@ info:
   url: https://github.com/spring-cloud-samples
 ----
 
+Using https health check with self-signed certificates:
```
ryanjbaxter: instead of just calling out self-signed certificates I would say

"To enable the health check request to accept all certificates when using HTTPs set `sidecar.accept-all-ssl-certificates` to `true`.

Hyunk3l: Much better! Done, thanks!

PR #3264, opened 2018-11-06
dynamic load turbine.appConfig

huanwei:
Signed-off-by: huanwei <huan@harmonycloud.cn>

This PR is just a tiny workaround to make turbine dynamically load "turbine.appConfig" property so that we do not need to restart turbine instance when updating the service list.


ryanjbaxter:
Why isnt this handled using the `/refresh` endpoint?

ryanjbaxter:
When using spring cloud config server and spring cloud bus a refresh can be triggered automatically

huanwei:
@ryanjbaxter In my scenario, I do not need Springcloud configserver + springcloud bus, instead I just use Kubernetes configmap to maintain the configuration, turbine will listen the configuration change. I think this way also works for configserver+bus.

Issue #3256, opened 2018-10-29
HystrixConcurrencyStrategy not obtain from  HystrixPlugins#getConcurrencyStrategy()


ailinzhou:
https://github.com/spring-cloud/spring-cloud-netflix/blob/6859b7a83588f33c8a946de40cfbab0917de3d05/spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java#L45-L46

``HystrixConcurrencyStrategy`` has no relation with Spring Container, and if someone registered via HystrixPlugins, **the following rest in init method  will drop an existing strategy**.
I  also  confused ： if we should inject a ``HystrixConcurrencyStrategy``  in Spring Container。

ryanjbaxter:
`SecurityContextConcurrencyStrategy` is what we configure

ailinzhou:
It seems that Hystrix shared HystrixConcurrencyStrategy between all Hystrix Command. In our are application, we registed a ConcurrencyStrategy by HystrixPlugin wrapp the callable for Some purposes. the follow code made the wrapCallable logic lost. 
https://github.com/spring-cloud/spring-cloud-netflix/blob/6859b7a83588f33c8a946de40cfbab0917de3d05/spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java#L51-L68

So, is that possible in ``HystrixSecurityAutoConfiguration.java`` the HystrixConcurrencyStrategy in com.netflix.hystrix.strategy.HystrixPlugins#getConcurrencyStrategy**  also be considered.

ryanjbaxter:
Can you provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) that reproduces the problem? It should be available as a GitHub (or similar) project or attached to this issue as a zip file.

ailinzhou:
Thank you for your reply @ryanjbaxter .
A demo.zip file was attached. 
When I run com.test.demo.DemoApplicationTests#whitShareSecurityContext  with **hystrix.shareSecurityContext=true** in application.properties the test failed, And if set **hystrix.shareSecurityContext=false** test will pass.
[demo.zip](https://github.com/spring-cloud/spring-cloud-netflix/files/2541972/demo.zip)

ryanjbaxter:
[...]

PR #3276, opened 2018-11-13
Multiple HystrixConcurrencyStrategy detected for #3256

ailinzhou:
Add multiple HystrixConcurrencyStrategy check. #3256

### DIFF
--- a/spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
+++ b/spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.netflix.hystrix.security;
 
 import javax.annotation.PostConstruct;
 
+import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategyDefault;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
```diff
@@ -56,18 +57,36 @@ public void init() {
...
-				new SecurityContextConcurrencyStrategy(existingConcurrencyStrategy));
+				new SecurityContextConcurrencyStrategy(concurrencyStrategy));
 		HystrixPlugins.getInstance().registerEventNotifier(eventNotifier);
 		HystrixPlugins.getInstance().registerMetricsPublisher(metricsPublisher);
 		HystrixPlugins.getInstance().registerPropertiesStrategy(propertiesStrategy);
 		HystrixPlugins.getInstance().registerCommandExecutionHook(commandExecutionHook);
 	}
 
+	private HystrixConcurrencyStrategy getConcurrencyStrategy() {
+		HystrixConcurrencyStrategy registeredStrategy = HystrixPlugins.getInstance()
```
ryanjbaxter: The logic here is kind of confusing the way it is written.  For example it seems that if `existingConcurrencyStrategy` is not `null` we could just return that right away.  I also dont think that the `IllegalStateException` is necessary, we should just default to the `existingConcurencyStrategy` if there is both registered (my suggestion above would eliminate the need for this).

spring-cloud-netflix-hystrix/src/test/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfigurationTest.java
```diff
@@ -0,0 +1,97 @@
+package org.springframework.cloud.netflix.hystrix.security;
```
ryanjbaxter: Missing license header.

ryanjbaxter:
Please submit this against the 2.0.x branch

spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
ailinzhou: > For example it seems that if `existingConcurrencyStrategy` is not `null` we could just return that right away

Good idea. will `detectRegisteredConcurrencyStrategy()` be better ?



> I also dont think that the `IllegalStateException` is necessary, we should just default to the `existingConcurencyStrategy` if there is both registered

If both registered and they are not same, we really don‘t know which to select. No matter which one we choose, we lose the logic in another one. For `IllegalStateException`, this is the some style with Hystrix. (In Hystrix when register a `ConcurrencyStrategy` if there already has one, IllegalStateException was thrown)

spring-cloud-netflix-hystrix/src/test/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfigurationTest.java
ailinzhou: All right, I will add it.

ailinzhou:
Is this mean I should checkout a branch from 2.0.x branch commit code in it and submit a new PR to 2.0.x ? or only need to submit this commit to as a PR to 2.0.x ?

spencergibb:
It also means rebasing your branch on 2.0.x

ailinzhou:
Submit this PR to 2.0.x and rebase the commit to 2.0.x was done. Please see #3297

PR #3277, opened 2018-11-13
Hystrix CircuitBreaker Implementation

ryanjbaxter:
Implementation based on https://github.com/spring-cloud/spring-cloud-commons/pull/445


spring-cloud-netflix-core/src/test/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+/**
+ * @author Ryan Baxter
+ */
+public class HystrixCircuitBreakerTest {
+
+	@Test
+	public void run() {
+		HystrixCircuitBreakerConfigFactory configFactory = new HystrixCircuitBreakerConfigFactory.DefaultHystrixCircuitBreakerConfigFactory();
```
spencergibb: Why isn't this just part of the builder? The default could be there if not set. I'd also add `HystrixCommand.Setter` as part of the builder similar to what we did with [HystrixCommands](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-hystrix/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCommands.java).

ryanjbaxter: You mean why is it passed in the constructor as opposed to as a method in the builder?  It is passed in the builder constructor now.  

I suppose we could experiment with adding a method to the builder interface and using type params to pass it as a parameter in the method.  The config object is going to be different for different implementations.  I am not sure what the benefit is of that though.

spencergibb: If a builder doesn't have more methods than that in the interface, not sure how useful it is.

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+ * Builds Hystrix circuit breakers.
+ *
+ * @author Ryan Baxter
+ */
+public class HystrixCircuitBreakerBuilder implements CircuitBreakerBuilder {
+
+	private String id;
+	private HystrixCircuitBreakerConfigFactory configFactory;
+
+	public HystrixCircuitBreakerBuilder(HystrixCircuitBreakerConfigFactory configFactory) {
```
ryanjbaxter: @spencergibb how about this?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfigFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import com.netflix.hystrix.HystrixCommandProperties;
+
+/**
+ * Provides configuration for circuit breakers.
+ *
+ * @author Ryan Baxter
+ */
+public interface HystrixCircuitBreakerConfigFactory {
+
+	public HystrixCommandProperties.Setter get(String id);
```
ryanjbaxter: Not sure the `id` parameter here makes sense anymore since `HystrixCircuitBreakerConfigFactory` is passed in via the builder.  You could in theory create one factory and use it in all builders in which case it would make sense to keep the `id` parameter.  Thoughts?

spencergibb: Yeah, I guess I'm questioning the need for this factory. Like `HystrixCommands` you could just have builder methods for the two `Setter`s.

spencergibb:
Also, why is this against 2.0.x? Shouldn't it be master for Greenwich?

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfigFactory.java
ryanjbaxter: I think its needed to abstract the configuration differences between various circuit breaker libraries.  For example with Hystrix there is only 1 configuration class while with R4J there are two (circuit breaker and timer), the builder interface wouldnt work if we had methods for each configuration class required

ryanjbaxter:
I was thinking we could initially put it in a SR of finchley since it is a net new feature (shouldnt effect anything else)

spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/HystrixCircuitBreakerConfigFactory.java
spencergibb: Except this doesn't extend anything and is specific to hystrix.

ryanjbaxter: Yes but the builder does

ryanjbaxter: Just to make things clearer....

Different circuit breaker implementation may have different ways to configure things.  For example with Hystrix you can configure the circuit breaker in the timeout in on object, `HystrixCommandProperties`.  However in R4J circuit breaker and timeouts are two distinct pieces of functionality and have their own config objects (and their own modules for that matter).  So to configure the R4J implementation we would need two config objects, `TimeLimiterConfig`, and `CircuitBreakerConfig`.  Since we want `CircuitBreakerBuilder` to be a common interface which passes the "config" for the circuit breaker we would need something to encapsulate the configuration for the circuit breaker so that the `CircuitBreakerBuilder` interface can be the same across all implementations.  

For reference here is the R4J config factory https://github.com/spring-cloud-incubator/spring-cloud-r4j/blob/master/spring-cloud-r4j-core/src/main/java/org/springframework/cloud/r4j/core/R4JConfigFactory.java

Now an argument can be made to remove the `id` parameter as I mentioned above, but we could also make the argument that is makes sense to keep it as well https://github.com/spring-cloud/spring-cloud-netflix/pull/3277#discussion_r234062774

ryanjbaxter:
Everything is now against master.

I got rid of the builder concept and moved to a circuit breaker factory concept as that seems to make more sense in this scenario.  The config factory still exists and is passed in to the constructor of the circuit breaker factory bean.  Hope this makes more sense.

ryanjbaxter:
Code has been moved to incubator project https://github.com/spring-cloud-incubator/spring-cloud-circuitbreaker

PR #3297, opened 2018-12-02
Submit PR #3276 against the 2.0.x branch

ailinzhou:
Follow suggestion of #3276 
1. Submit PR against 2.0.x.
2. Add ASF license header.
3. Method name `getConcurrencyStrategy` to `detectRegisteredConcurrencyStrategy`

### DIFF
--- a/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
+++ b/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.netflix.hystrix.security;
 
 import javax.annotation.PostConstruct;
 
+import com.netflix.hystrix.strategy.concurrency.HystrixConcurrencyStrategyDefault;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.AllNestedConditions;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/hystrix/security/HystrixSecurityAutoConfiguration.java
```diff
@@ -56,18 +57,36 @@ public void init() {
...
+				.getConcurrencyStrategy();
+		if (existingConcurrencyStrategy == null) {
+			return registeredStrategy;
+		}
+		//Hystrix registered a default Strategy.
+		if (registeredStrategy instanceof HystrixConcurrencyStrategyDefault){
+			return existingConcurrencyStrategy;
+		}
+		//If registeredStrategy not the default and not some with existingConcurrencyStrategy throw a exception.
+		if (!existingConcurrencyStrategy.equals(registeredStrategy)){
```
ryanjbaxter: Why throw an exception in this case?  I think its fine to just return `existingConcurrencyStrategy`.

ailinzhou: if existingConcurrencyStrategy is not some with `registeredStrategy`, We can't simply use the `existingConcurrencyStrategy`, if we do that the  wrapper code logic in `registeredStrategy` will lost, obviously this will lead to a bug.
The exception for inform the user this case is not correct, and they must deal with case.
Do you have any other advice(inform the user this case is not correct, and they must deal with case)?

ryanjbaxter: Say that `existingConcurrencyStrategy` is not `null` (they provided a bean) and they also set a `registeredStrategy` why would returning `existingConcurrencyStrategy` be wrong?  Yes it doesnt make too much sense to have set both but I dont see how choosing `existingConcurrencyStrategy` as the default return value being a bug.

PR #3321, opened 2018-12-14
Decode query params in RequestContentDataExtractor

### LINKED ISSUE #3300: RequestContentDataExtractor#extractFromMultipartRequest duplicated url-parameter-value if content-type is multipart/form-data and value contains escaped characters
### Bugreport:

The value of a URL parameter gets duplicated if content-type is multipart/form-data and the value contains escaped characters.

`POST 'http://localhost/test?name=hello%20world'`
results in
`name = hello world,hello world`

This happens in [org.springframework.cloud.netflix.zuul.util.RequestContentDataExtractor#extractFromMultipartRequest](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/util/RequestContentDataExtractor.java#L68).
There two lists should be compared but the first _listOfAllParams_ contains unescaped values and the other _listOfOnlyQueryParams_ contains escaped values.

Example application: https://github.com/s3erjaeh/example_for_spring-cloud-netflix-issue 

### Environment: 
`org.springframework.cloud:spring-cloud-netflix-zuul:2.0.1.RELEASE`

included via 
- org.springframework.cloud; version: Finchley.SR1
- org.springframework.boot; version: 2.0.5.RELEASE

ryanjbaxter:
Fixes #3300

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/util/RequestContentDataExtractor.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/util/RequestContentDataExtractor.java
@@ -18,8 +18,9 @@ package org.springframework.cloud.netflix.zuul.util;
 
 import static java.util.Arrays.stream;
 import static java.util.Collections.emptyMap;
-import static org.springframework.util.StringUtils.*;
 import static org.springframework.util.StringUtils.isEmpty;
+import static org.springframework.util.StringUtils.tokenizeToStringArray;
+import static org.springframework.util.StringUtils.uriDecode;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/util/RequestContentDataExtractor.java
```diff
@@ -78,12 +79,16 @@
 					.collect(Collectors.toList());
 			List<String> listOfOnlyQueryParams = queryParamsGroupedByName.get(key);
 
-			if (listOfOnlyQueryParams != null
-					&& !listOfOnlyQueryParams.containsAll(listOfAllParams)) {
-				listOfAllParams.removeAll(listOfOnlyQueryParams);
-				for (String value : listOfAllParams) {
-					builder.add(key,
-							new HttpEntity<>(value, newHttpHeaders(request, key)));
+			if(listOfOnlyQueryParams != null) {
```
spencergibb: Can you maybe comment in words what this is doing? Even before the change I'm not quite sure.

ryanjbaxter: I mean I can explain what the code is doing....

`request.getParameterMap` returns query params and form data.  If there is form data (`listOfOnlyQueryParams` does not contain `listOfAllParams`) then we remove the query params from `listofAllParams` and add the form params to the Map.  What I am not clear on is what that map ends up being used for...

Issue #3359, opened 2019-01-21
Defect of hystrix stream queue configuration


Incarnation-p-lee:
I would like to config the send-Rate of hystrix.stream.queue with some yml configuration like below code.
```yml
hystrix:
  stream:
    queue:
      send-id: false
      #      gather-rate: 2500
      send-rate: 2000
```

But still not working as the client always send with 500ms. After some investigation of source code in below
https://github.com/spring-cloud/spring-cloud-netflix/blob/5ecd0842cec00914206742c1d95c4b0dcb8fc5a8/spring-cloud-netflix-hystrix-stream/src/main/java/org/springframework/cloud/netflix/hystrix/stream/HystrixStreamTask.java#L92-L118

It should use `sendRate` instead of `send-rate`, as it not come from properties bean, only obtain properties value.

From one user's view, these two format should both work well as expect. The `@Scheduled` should leverage the value from property bean instead of literal property.

I can prepare the PR for it later if no urgent for this issus.

ryanjbaxter:
A PR would be great!  Thanks!

ryanjbaxter:
@Incarnation-p-lee Why do you think it should be `send-rate` instead of `sendRate`?

Incarnation-p-lee:
@Incarnation-p-lee 
I don't think it should be `send-rate` or `sendRate`. I mean if user leverage this property in `application.yml`, both format should works well as expected. But for now only `sendRate` works.

ryanjbaxter:
Ah I see what you are saying now, thanks

PR #3361, opened 2019-01-23
issue-3359

dev-priporov:
fixed _send-rate_ in HystrixStreamTask.
Seems that there is the same defect for _gather-rate_ property. It's also fixed. 
#3359


ryanjbaxter:
See the discussion in #3359, this is not the fix we need to make

PR #3364, opened 2019-01-24
Fix defect of hystrix stream queue configuration, fix #3359.

### LINKED ISSUE #3359: Defect of hystrix stream queue configuration

Incarnation-p-lee:
* Support 2 config format send-rate and sendRate.
* If both given, send-rate will be ignore.
Fixes #3359
Signed-off-by: Pan Li <panli@microsoft.com>

### DIFF
--- a/spring-cloud-netflix-hystrix-stream/src/main/java/org/springframework/cloud/netflix/hystrix/stream/HystrixStreamTask.java
+++ b/spring-cloud-netflix-hystrix-stream/src/main/java/org/springframework/cloud/netflix/hystrix/stream/HystrixStreamTask.java
@@ -89,7 +89,7 @@ public class HystrixStreamTask implements ApplicationContextAware {
 	}
 
 	// TODO: use integration to split this up?
-	@Scheduled(fixedRateString = "${hystrix.stream.queue.sendRate:500}")
+	@Scheduled(fixedRateString = "${hystrix.stream.queue.sendRate:${hystrix.stream.queue.send-rate:500}}")
 	public void sendMetrics() {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Incarnation-p-lee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3364)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3364) to manually synchronize the status of this Pull Request.

ryanjbaxter:
can you submit this against the 2.0.x branch?

Incarnation-p-lee:
@ryanjbaxter 
Sure, my pleasure.

PR #3396, opened 2019-02-19
let the property:"management.endpoints.web.base-path" affect the default value of eureka's statusPageUrlPath and healthCheckUrlPath

aftersss:
Now when I use the property `management.endpoints.web.base-path` set to `/endpoint`, I want to affect the `statusPageUrl` and `healthCheckUrl` which is registered into eurekaServer, but actually they can not be affected. So now I have to add two more properties to fix this problem:

```
eureka.instance.health-check-url-path=${management.endpoints.web.base-path}/health
eureka.instance.status-page-url-path=${management.endpoints.web.base-path}/info
```

This pull request try to fix this problem, so the above two properties will not be required any more.


TYsewyn:
Some notes to this PR:
We shouldn't be using any fixed values like `/actuator`, and `/info` or `/health`.
The complete URI can be changed via properties: `server.servlet.context-path`, `management.server.servlet.context-path`, `management.endpoints.web.base-path` and `management.endpoints.web.path-mapping`.
AFAIK there are a couple of places where changes are needed to accomplish this change.

Issue #3406, opened 2019-03-02
Connection pool shut down on "zuul.host.*" property change


denysivano:
**Problem:**

When using Spring Boot + Spring Cloud Netflix Zuul and changing any `zuul.host.*` property dynamically, an exception occurred on any subsequent forwarding request:
```
2019-03-02 17:59:28.310  WARN 21726 --- [nio-8080-exec-5] o.s.c.n.z.filters.post.SendErrorFilter   : Error during filtering

com.netflix.zuul.exception.ZuulException: Forwarding error
	at org.springframework.cloud.netflix.zuul.filters.route.SimpleHostRoutingFilter.handleException(SimpleHostRoutingFilter.java:246) ~[spring-cloud-netflix-zuul-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	at org.springframework.cloud.netflix.zuul.filters.route.SimpleHostRoutingFilter.run(SimpleHostRoutingFilter.java:226) ~[spring-cloud-netflix-zuul-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	at com.netflix.zuul.ZuulFilter.runFilter(ZuulFilter.java:117) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.FilterProcessor.processZuulFilter(FilterProcessor.java:193) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.FilterProcessor.runFilters(FilterProcessor.java:157) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.FilterProcessor.route(FilterProcessor.java:118) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.ZuulRunner.route(ZuulRunner.java:96) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.http.ZuulServlet.route(ZuulServlet.java:116) ~[zuul-core-1.3.1.jar:1.3.1]
	at com.netflix.zuul.http.ZuulServlet.service(ZuulServlet.java:81) ~[zuul-core-1.3.1.jar:1.3.1]
	at org.springframework.web.servlet.mvc.ServletWrappingController.handleRequestInternal(ServletWrappingController.java:165) [spring-webmvc-5.1.5.RELEASE.jar:5.1.5.RELEASE]
[...]

PR #3407, opened 2019-03-02
Re-create connection manager on "zuul.host.*" property change

denysivano:
When changing `zuul.host.*` properties, it makes a sense to re-create `HttpClientConnectionManager` along with `HttpClient` to make `maxTotalConnections` and `maxConnectionsPerRoute` properties dynamically configurable.

It also prevents the use of shut down connection pool (fixes gh-3406).

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
@@ -133,15 +133,20 @@ public class SimpleHostRoutingFilter extends ZuulFilter
 
 			if (createNewClient) {
 				try {
-					SimpleHostRoutingFilter.this.httpClient.close();
+					this.httpClient.close();
 				}
 				catch (IOException ex) {
 					log.error("error closing client", ex);
 				}
 				// Re-create connection manager (may be shut down on HTTP client close)
-				SimpleHostRoutingFilter.this.connectionManager.shutdown();
-				SimpleHostRoutingFilter.this.connectionManager = newConnectionManager();
-				SimpleHostRoutingFilter.this.httpClient = newClient();
+				try {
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/route/SimpleHostRoutingFilter.java
```diff
@@ -137,6 +138,9 @@ public void onPropertyChange(EnvironmentChangeEvent event) {
 				catch (IOException ex) {
 					log.error("error closing client", ex);
 				}
+				// Re-create connection manager (may be shut down on HTTP client close)
+				SimpleHostRoutingFilter.this.connectionManager.shutdown();
```
spencergibb: This should move in the above try catch or in its own.

denysivano: `connectionManager.shutdown()` doesn't throw `IOException`, but `httpClient.close()` does.

Is there need to catch `RuntimeException` for `connectionManager.shutdown()` ?

spencergibb: you're correct.

denysivano: Wrapped it in e4f3f60addd4e77e567fecc6dd77c96e6a5e0994

denysivano:
@ryanjbaxter Is there any plans to backport this PR to `2.1.x` branch? As I can see from the milestone, this fix had to be included in `2.1.3.RELEASE` (Greenwich.SR3), but it wasn't.

ryanjbaxter:
@denysivano sorry about that I applied it to the 2.1.x branch it will be in Greenwich.SR4

PR #3421, opened 2019-03-15
Don't set content-encoding header when un-gzipping

candrews:
If the original request has Accept-Encoding not set, the server may still send a gzipped response back to the zuul proxy. In that case, zuul will copy the backend response's headers to the zuul response headers (including Content-Encoding: gzip). Therefore, when zuul is intentionally ungzipping the response, it must set the response Content-Encoding to identity - otherwise, zuul's response would have the Content-Encoding: gzip header but the body would not be gzipped.

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/post/SendResponseFilter.java
@@ -125,6 +125,7 @@ public class SendResponseFilter extends ZuulFilter {
 			servletResponse.setCharacterEncoding("UTF-8");
 		}
 		
+		String servletResponseContentEncoding = getResponseContentEncoding(context);
 		OutputStream outStream = servletResponse.getOutputStream();
 		InputStream is = null;
 		try {
@@ -140,12 +141,16 @@ public class SendResponseFilter extends ZuulFilter {
 					// decompress stream before sending to client
 					// else, stream gzip directly to client
 					if (isGzipRequested(context)) {
-						servletResponse.setHeader(ZuulHeaders.CONTENT_ENCODING, "gzip");
+						servletResponseContentEncoding = "gzip";
 					}
 					else {
+						servletResponseContentEncoding = null;
 						is = handleGzipStream(is);
 					}
 				}
+			}
+			if(servletResponseContentEncoding != null) {
+				servletResponse.setHeader(ZuulHeaders.CONTENT_ENCODING, servletResponseContentEncoding);
 			}
 			
 			if (is!=null) {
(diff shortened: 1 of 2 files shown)


candrews:
I've added a test, `runWithOriginContentLength_gzipNotRequested_gzipResponse_content_encoding_header`, that demonstrates this problem.

It's just the existing test `runWithOriginContentLength_gzipNotRequested_gzipResponse` with one line added:
`RequestContext.getCurrentContext().addZuulResponseHeader("Content-Encoding", "gzip");`

ryanjbaxter:
Thanks I think it makes sense to submit this against the 2.0.x branch and then we can merge it forward

candrews:
> Thanks I think it makes sense to submit this against the 2.0.x branch and then we can merge it forward

I've rebased onto 2.0.x, thanks!

ryanjbaxter:
Can you look at the test failures?

candrews:
I can't see how my change could be related to the failing test, which is:
```
[INFO] Running org.springframework.cloud.netflix.turbine.stream.TurbineStreamTests
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.983 s <<< FAILURE! - in org.springframework.cloud.netflix.turbine.stream.TurbineStreamTests
[ERROR] contextLoads(org.springframework.cloud.netflix.turbine.stream.TurbineStreamTests)  Time elapsed: 0.001 s  <<< ERROR!
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.BeanDefinitionStoreException: Failed to process import candidates for configuration class [org.springframework.boot.autoconfigure.integration.IntegrationAutoConfiguration$IntegrationConfiguration]; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.integration.config.IntegrationRegistrar]: Unresolvable class definition; nested exception is java.lang.ExceptionInInitializerError
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.integration.config.IntegrationRegistrar]: Unresolvable class definition; nested exception is java.lang.ExceptionInInitializerError
Caused by: java.lang.ExceptionInInitializerError
Caused by: org.springframework.context.ApplicationContextException: Starting with Spring Integration 5.0, the 'spring-integration-java-dsl' dependency is no longer needed; the Java DSL has been merged into the core project. If it is present on the classpath, it will cause class loading conflicts.
```

This PR doesn't impact Turbine.

mum-viadee:
We had the same problem, because OWASP ZAP by default removes the "Accept-Encoding" header if no supported encoding is left after filtering the given encodings ([see JAVADOC](https://static.javadoc.io/org.zaproxy/zap/2.7.0/org/parosproxy/paros/core/proxy/ProxyParam.html#setRemoveUnsupportedEncodings-boolean-)) .  GZIP seems to be an unsupported encoding in an OWASP ZAP point of view... :-)

After reading the [RFC7231](https://tools.ietf.org/html/rfc7231#section-3.1.2.2) and looking into the [HTTP Content Coding Registry](http://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding) of the IANA I think "identity" is reserved only for the "Accept-Encoding" header as synonym for "no encoding". Besides the RFC says, that the "Content-Encoding" header field only needs to be set by the sender, if one or more encodings have been applied to a representation.  

In my opinion it would be better to remove the "Content-Encoding" header from the response, if no content-encoding was applied.

candrews:
Thank you for that information, @mum-viadee - I agree, so I've updated this PR to not set the `Content-Encoding` response header for the unzipping case.

ryanjbaxter:
@candrews I am not sure I didn't have time to look into it since there are other things on my plate.  I just know the test is not failing in the 2.0.x branch without your code.

candrews:
> @candrews I am not sure I didn't have time to look into it since there are other things on my plate. I just know the test is not failing in the 2.0.x branch without your code.

All tests are now passing, so whatever it was, it's no longer happening and is no longer relevant :)

candrews:
Can this PR now be merged?

candrews:
@ryanjbaxter can you please merge this PR (and tell me what I can do to get it to that point)?

Also, once this is merged, can you please consider making a release? This is a really big problem for my project, and working around it is troublesome.

candrews:
@spencergibb is this something you can merge and consider for a near term release?

Thanks again!

Issue #3418, opened 2019-03-14
Only request bodies with content-type multipart/* and application/x-www-form-urlencoded work


candrews:
Make a request through zuul using a `POST` or `PUT` verb with a content-type that is not `multipart/* or `application/x-www-form-urlencoded`

When Zuul makes the request to the backend, it won't include the request body.

The reason is that the original request body is consumed in `org.springframework.cloud.netflix.zuul.filters.TraceProxyRequestHelper.debug(String, String, MultiValueMap<String, String>, MultiValueMap<String, String>, InputStream)` leaving no request body to consume when `org.springframework.cloud.netflix.zuul.filters.route.SimpleHostRoutingFilter` actually sends the request.

Requests of content-type `application/x-www-form-urlencoded` work because `org.springframework.cloud.netflix.zuul.filters.pre.FormBodyWrapperFilter` gets the body. Requests of type `multipart/*` work because `org.springframework.cloud.netflix.zuul.filters.ProxyRequestHelper.shouldDebugBody(RequestContext)` explicitly excludes them.

To reproduce the issue, just make a request through zuul with a Content-Type of `text/plain`, for example, and you'll always get a time out error as the backend server will keep waiting for response body that it will never get.

I believe this issue was originally reported at https://github.com/spring-cloud/spring-cloud-netflix/issues/2610

spencergibb:
I'm unable to reproduce this with a POST with text/plain.

Can you provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) that reproduces the problem?  It should be available as a GitHub (or similar) project or attached to this issue as a zip file.

candrews:
My apologies if this isn't that completely minimal, but it does reproduce the problem and it's really easy to run.

Grab this:
https://www.integralblue.com/zuul3418.tar.gz

And run: `./mvnw spring-boot:run -Dspring-boot.run.profiles=localhost`

See the problem with:
`curl -H 'Content-Type: abc/xyz' --data "stuff" http://localhost:8080/TEST.php`

Now see how the problem is fixed by editing `src/main/resources/application.properties` and uncommenting the `zuul.trace-request-body=false` line at the bottom.

and trying that same curl command again.

spencergibb:
```
$ curl -H 'Content-Type: abc/xyz' --data "stuff" http://localhost:8080/TEST.php                                                                                                    (9121)[14:19:53]
[...]

PR #3440, opened 2019-03-19
Enable request buffering when request body tracing

candrews:
Request body tracing reads the body then it will be read again later for actual handling.
Therefore, request buffering is required when request body tracing.

Fixes https://github.com/spring-cloud/spring-cloud-netflix/issues/3418

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/Servlet30WrapperFilter.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/Servlet30WrapperFilter.java
@@ -20,6 +20,8 @@ import java.lang.reflect.Field;
 
 import javax.servlet.http.HttpServletRequest;
 
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.netflix.zuul.filters.ZuulProperties;
 import org.springframework.cloud.netflix.zuul.util.RequestUtils;
 import org.springframework.util.Assert;
 import org.springframework.util.ReflectionUtils;
@@ -38,6 +40,8 @@ import static org.springframework.cloud.netflix.zuul.filters.support.FilterConst
  */
 public class Servlet30WrapperFilter extends ZuulFilter {
 
+	@Autowired
+	private ZuulProperties zuulProperties;
 	private Field requestField = null;
 
 	public Servlet30WrapperFilter() {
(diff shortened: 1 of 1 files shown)


spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/pre/Servlet30WrapperFilter.java
```diff
@@ -80,6 +84,10 @@ else if (RequestUtils.isDispatcherServletRequest()) {
 			// If it's going through the dispatcher we need to buffer the body
 			ctx.setRequest(new Servlet30RequestWrapper(request));
 		}
+		else if (zuulProperties.isTraceRequestBody()) {
```
spencergibb: I'm not sure this is exactly what I described.

> turn off trace request body if zuul.use-filter=true.

This enables buffering if `isTraceRequestBody()` which isn't the same.

candrews: > turn off trace request body if zuul.use-filter=true.

What you suggest seems like it'll be really awkward to program and unexpected for users:
if `zuul.use-filter=true`, ignore `zuul.traceRequestBody` and treat it as if it were always false. This implementation would confuse users who expect `zuul.traceRequestBody` to do something, and it's weird for `traceRequestBody` to be a variable that's sometimes ignored. IMHO, making `zuul.traceRequestBody` always work as expected is a better solution for maintainability and users alike, which is the approach I took in this PR.

spencergibb: It also unexpectedly introduces buffering because trace request body is true by default. I won't apply this PR as is.

candrews: Can you provide any guidance on what I can do?
Things I can think of include making some other change that you suggest or rebasing against another branch.

spencergibb: Add another case here https://github.com/spring-cloud/spring-cloud-netflix/blob/eff47e046410dd0b6f3ccd600195e0ba4f1a1827/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/ProxyRequestHelper.java#L266

candrews: Apologies for my density... what case are suggesting be added?

spencergibb: `if zuul.use-filter=true: return false`

PR #3500, opened 2019-04-11
Give users a choice to determine whether the EurekaClient instance can be refreshed or not.

aftersss:
We found that when we trigger `/actuator/refresh` to refresh our program, the `EurekaClient` instance will be destroyed and reconstructed again, so our program will unregister from eurekaServer and after some seconds, it will register to eurekaServer again. 

Consider that when I visit `/actuator/refresh` to all  our `Service A` instances at the same time, there will be some seconds no one instance alive on eurekaServer, if at the same time,`Service B` tries to fetch instances of `Service A`, it will fetch a empty list, this is very bad. And Indeed we meet this problem.

I think `RefreshScope` is useful, but at most cases, `EurekaClient` instance doesn't need to refresh,  this PR give users a choice to determine whether the EurekaClient instance can be refreshed or not, while keeping the ability to refresh other beans.

For the backward compatibility,  default value of `eureka.client.refresh.enable` is set to true.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-netflix.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-netflix.adoc
@@ -324,6 +324,19 @@ eureka.instance.metadataMap.zone = zone2
 eureka.client.preferSameZoneEureka = true
 ```
 
+=== Refresh Scope of Eureka Client
+
+By default, the `EurekaClient` bean is refreshable, so the Eureka client properties can be refreshed, but this has a disadvantage:
+with the default setting, the `EurekaClient` instance will be destroyed and reconstructed again when refreshed, so your program will unregister from eurekaServer and after some seconds, it will register to eurekaServer again,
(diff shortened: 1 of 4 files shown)


ryanjbaxter:
Doesnt this assume that there is only one instance of service A?  If there are multiple instances it shouldnt be a problem right?

aftersss:
We have multiple instances of service A, notice that `refresh` is trigged at the same time, that means  multiple instances will deregister from eurekaServer at the same time.

ryanjbaxter:
@aftersss We are fine with this change, the one thing we would ask is that you add some documentation about the new property and include a warning that if enabled none of the Eureka client properties will be refreshable.

aftersss:
Done, but sorry, my English is not very good, maybe there are some grammar and terms error, If any , please correct it.  ^_^

zhanwh9:
Is this just fix part of the problem?
I think the `EurekaClientConfigurationRefresher` in `EurekaDiscoveryClientConfiguration` should be conditional on property `eureka.client.refresh.enable` too, or it will deregister and reregister by EurekaAutoServiceRegistration during refresh.

Issue #2732, opened 2018-02-17
Randomized server port registers with management port in eureka


pcornelissen:
I encountered a strange behaviour, when I set management and server port to 0 so I can easily start a few local instances, then the eureka client registers the service with the management port.

My Setup: Spring Boot 1.5.10 + Edgeware.SR2

- BookmarksService with server+management port = 0
- Eureka Server
- Zuul: Forwards all requests to BookmarksService instances

Using a fixed IP list for ribbon works.
Enabling DiscoveryClient in the bookmarksservice and Zuul leads to JSON 404 error messages produced by the BookmarksService. After a long while of searching I found that the service name seems to be resolved with the management port instead of the correct one.

I have verified that using the correct port without zuul/ribbon the correct result is returned and using the management port I see the same JSON error message that I see when I am using json.

This can be fixed by not using a management port. But that requires additional access control so no one accesses the actuator stuff via zuul.

This is the relevant part of the application.yaml with a management port set
```
server.port: 0
management:
  port: 10000
  context-path: /internal

eureka:
  client:
    serviceUrl:
      defaultZone: http://localhost:8761/eureka/
  instance:
    instance-id: ${spring.application.name}:${random.value}
```

I also tried to set the management port to a fixed 10000 and this is the result in eureka:
```
<application>
  <name>BOOKMARKS</name>
  <instance>
    <instanceId>bookmarks:96a084044a99b3ea5ce02b05f0c19619</instanceId>
    <hostName>192.168.2.20</hostName>
    <app>BOOKMARKS</app>
    <ipAddr>192.168.2.20</ipAddr>
    <status>UP</status>
    <overriddenstatus>UNKNOWN</overriddenstatus>
    <port enabled="true">10000</port>
    <securePort enabled="false">443</securePort>
    <countryId>1</countryId>
    <dataCenterInfo class="com.netflix.appinfo.InstanceInfo$DefaultDataCenterInfo">
      <name>MyOwn</name>
    </dataCenterInfo>
    <leaseInfo>
      <renewalIntervalInSecs>30</renewalIntervalInSecs>
      <durationInSecs>90</durationInSecs>
      <registrationTimestamp>1518881421504</registrationTimestamp>
      <lastRenewalTimestamp>1518881421504</lastRenewalTimestamp>
      <evictionTimestamp>0</evictionTimestamp>
      <serviceUpTimestamp>1518881420997</serviceUpTimestamp>
    </leaseInfo>
[...]

PR #3514, opened 2019-04-22
Filter WebServerInitializedEvent for web-only.

chriswhite199:
Fixes gh-2732

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
@@ -30,7 +30,6 @@ import org.springframework.context.SmartLifecycle;
 import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.context.event.EventListener;
 import org.springframework.core.Ordered;
-import org.springframework.util.StringUtils;
 
 /**
  * @author Dave Syer
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
@@ -30,6 +30,7 @@ import org.springframework.context.SmartLifecycle;
 import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.context.event.EventListener;
 import org.springframework.core.Ordered;
+import org.springframework.util.StringUtils;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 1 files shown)


chriswhite199:
Test failures appear unrelated to changes, and match those from a vanilla checkout AFAIK

ryanjbaxter:
Master is passing in CircleCi.  Also I would think we would want this submitted against the 2.0.x branch and we can merge it forward.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
```diff
@@ -119,11 +119,14 @@ public int getOrder() {
 	@EventListener(WebServerInitializedEvent.class)
 	public void onApplicationEvent(WebServerInitializedEvent event) {
 		// TODO: take SSL into account
-		int localPort = event.getWebServer().getPort();
-		if (this.port.get() == 0) {
-			log.info("Updating port to " + localPort);
-			this.port.compareAndSet(0, localPort);
-			start();
+		String contextName = event.getApplicationContext().getServerNamespace();
+		if (StringUtils.isEmpty(contextName)) {
```
spencergibb: Is there a guarantee that this will always be true?

chriswhite199: No guarantee from what i can see as the web app context has a `setServerNamespace` method, but this only appears to be called by the actuator.

https://github.com/spring-projects/spring-boot/search?q=serverNamespace&unscoped_q=serverNamespace

chriswhite199: I'll invert the logic to filter out management server namespace

spencergibb: Yeah, if a user sets it, this won't hold either. Need to check with the boot team to see if there is a better way to identify an actuator context.

chriswhite199: Well another way might be to replace the port instance variable:

```
@LocalServerPort
int port;
```

and then the entire `onApplicationEvent` method can be simplified:

```
@EventListener(ApplicationStartedEvent.class)
public void onApplicationEvent(ApplicationStartedEvent event) {
	start();
}
```

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/serviceregistry/EurekaAutoServiceRegistration.java
```diff
@@ -30,6 +30,7 @@
 import org.springframework.context.event.ContextClosedEvent;
 import org.springframework.context.event.EventListener;
 import org.springframework.core.Ordered;
+import org.springframework.util.StringUtils;
```
spencergibb: Please remove unused import.

Issue #3554, opened 2019-06-07
Not able to add ClientFilter for eureka replication client


LittleBaiBai:
Since Netflix Eureka 1.9.10 (https://github.com/Netflix/eureka/releases/tag/v1.9.10), it's possible to add additional filters to an existing JerseyReplicationClient. 

To do so, `RefreshablePeerEurekaNodes` needs to override `createPeerEurekaNode()` and add additional filters once the JerseyReplicationClient is created.

venkatesh-007:
@LittleBaiBai  can you elaborate this?
Any articles or info will be appreciated

venkatesh-007:
@spencergibb  any workaround for this for older eureka versions
I've to use it with eureka-core-1.6.2, any way to fork JerseyReplicationClient implementation

spencergibb:
No articles or info or any workarounds that I know of.

LittleBaiBai:
@venkatesh-007 If you really need to use `eureka-core-1.6.2`, I believe you will have to do some unfortunate reflections like this:
```
	@Bean
	public PeerEurekaNodes peerEurekaNodes(PeerAwareInstanceRegistry registry,
										   ServerCodecs serverCodecs,
										   ApplicationInfoManager applicationInfoManager,
										   EurekaServerConfig eurekaServerConfig,
										   EurekaClientConfig eurekaClientConfig) {
		return new PeerEurekaNodes(registry, eurekaServerConfig,
			eurekaClientConfig, serverCodecs, applicationInfoManager) {
			@Override
			protected PeerEurekaNode createPeerEurekaNode(String peerEurekaNodeUrl) {
				PeerEurekaNode peerEurekaNode = super.createPeerEurekaNode(peerEurekaNodeUrl);
				JerseyReplicationClient replicationClient = getField(PeerEurekaNode.class,
					peerEurekaNode, "replicationClient");
				ApacheHttpClient4 jerseyApacheClient = getField(JerseyReplicationClient.class,
					replicationClient, "jerseyApacheClient");
				jerseyApacheClient.addFilter(new YourCustomClientFilter()));
				return peerEurekaNode;
			}
		};
	}
```

venkatesh-007:
Thanks @LittleBaiBai
Also are there any disadvantages for running a eureka server(service registry) over http (non ssl) in production?

In case my requirement is to use it on a on-premise setup(on premise cloud or vsis)
[...]

PR #3572, opened 2019-06-20
set replication  client filters in RefreshablePeerEurekaNodes

LittleBaiBai:
fixes gh-3554

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
@@ -36,10 +36,12 @@ import com.netflix.discovery.converters.wrappers.CodecWrappers;
 import com.netflix.eureka.DefaultEurekaServerContext;
 import com.netflix.eureka.EurekaServerConfig;
 import com.netflix.eureka.EurekaServerContext;
+import com.netflix.eureka.cluster.PeerEurekaNode;
 import com.netflix.eureka.cluster.PeerEurekaNodes;
 import com.netflix.eureka.registry.PeerAwareInstanceRegistry;
 import com.netflix.eureka.resources.DefaultServerCodecs;
 import com.netflix.eureka.resources.ServerCodecs;
+import com.netflix.eureka.transport.JerseyReplicationClient;
 import com.sun.jersey.api.core.DefaultResourceConfig;
 import com.sun.jersey.spi.container.servlet.ServletContainer;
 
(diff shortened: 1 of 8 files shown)


spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerAutoConfiguration.java
```diff
@@ -275,12 +285,32 @@ public EurekaServerConfig eurekaServerConfig(EurekaClientConfig clientConfig) {
...
+				final ServerCodecs serverCodecs,
+				final ApplicationInfoManager applicationInfoManager,
+				final ReplicationClientAdditionalFilters replicationClientAdditionalFilters) {
+			super(registry, serverConfig, clientConfig, serverCodecs, applicationInfoManager);
+			this.replicationClientAdditionalFilters = replicationClientAdditionalFilters;
+		}
+
+		@Override
+		protected PeerEurekaNode createPeerEurekaNode(String peerEurekaNodeUrl) {
+			JerseyReplicationClient replicationClient = JerseyReplicationClient
```
ryanjbaxter: Does this work when you do not use Jersey and use RestTemplate?

LittleBaiBai: My understanding is that peer replication calls are all handled in Netflix eureka-core with those statically created JerseyReplicationClient for each peer node. (https://github.com/Netflix/eureka/blob/8325086bb6012841ab85e0288d3ceec743b79a76/eureka-core/src/main/java/com/netflix/eureka/cluster/PeerEurekaNodes.java#L197) 
Is there a way to use RestTemplate instead of Jersey?

ryanjbaxter: You can configure the Eureka client to not use jersey, see https://cloud.spring.io/spring-cloud-static/Greenwich.SR1/single/spring-cloud.html#_eurekaclient_without_jersey

I am unsure if this will effect peep to peer replication as well, but I would like to confirm if it does or doesnt

spencergibb: There is, but it looks like it might need to be extended for peer replication. The rest template stuff is mostly for eureka clients. I'm not too worried about it here. If there is ever a need we can certainly open an issue.

LittleBaiBai:
The circleci error message indicates missing environment variables. How can I set those?

LittleBaiBai:
@spencergibb any update on this PR?

spencergibb:
Rebuilding, it was a java compile failure.

spencergibb:
Does this compile for you locally? It doesn't for me. @LittleBaiBai

LittleBaiBai:
@spencergibb Sorry about that. Intellij was hiding the problem from me. It's fixed now.

ryanjbaxter:
@LittleBaiBai would you like this in the next Greenwich SR?  If so you should submit the PR against the 2.1.x branch.

LittleBaiBai:
@ryanjbaxter Thanks! I'll close this one and submit a new one against 2.1.x.

PR #3580, opened 2019-06-28
Update RoutesEndpoint.java

ZengYueWei:
CompositeRouteLocator will sort the routeLocator by their order , so that the higher priority routeLocator will be check first.

But when we checked the route infos (by getting /actuator/routes)，the route infos are different from the result we request.

It turns out that the route infos are returned by map. Because of that, if they are same route infos offering by different routeLocator, the lower priority will rewrite the higher one route info. But when we send the request, the higher one will perform.

So, I think it's something we can change to reduce ambiguity by returing same route infos as we request.

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/RoutesEndpoint.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/RoutesEndpoint.java
@@ -64,7 +64,7 @@ public class RoutesEndpoint implements ApplicationEventPublisherAware {
 	public Map<String, String> invoke() {
 		Map<String, String> map = new LinkedHashMap<>();
 		for (Route route : this.routes.getRoutes()) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ZengYueWei Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3580)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3580) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ZengYueWei Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3580)!

spencergibb:
Can you include a test?

vsasvipul0605 (approved):

PR #3585, opened 2019-07-02
Separate and fix tests in SidecarApplicationTests

### LINKED ISSUE #3584: hidden failing unit tests in spring-cloud-netflix-sidecar
There are several unit tests in spring-cloud-netflix-sidecar that don't run via mvn and that fail when run via intellij. I think they don't run via mvn because the inner classes don't match the failsafe plugin test pattern. I noticed them because I ran unit tests via intellij (right-click spring-cloud-netflix-sidecar/src/test/java and choose "Run All Tests") and saw some failures. The failing unit tests are SidecarApplicationTests.ManagementContextPathStatusAndHealthCheckUrls., SidecarApplicationTests.AcceptAllSslCertificatesContext.testUseRestTemplateWhenHttpClientIsNotAvailable, SidecarApplicationTests.ServerContextPathStatusAndHealthCheckUrls.testStatusAndHealthCheckUrls.

These should probably be fixed and renamed so that they run via mvn, or removed if they're not testing something relevant.

ryanjbaxter:
Fixes #3584


spring-cloud-netflix-sidecar/src/test/java/org/springframework/cloud/netflix/sidecar/AcceptAllSslCertificatesContext.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.RANDOM_PORT;
+
+/**
+ * @author Ryan Baxter
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = SidecarApplication.class, webEnvironment = RANDOM_PORT, value = {
+		"sidecar.accept-all-ssl-certificates=false" })
+public class AcceptAllSslCertificatesContext {
```
spencergibb: Does this need to end of Tests?

PR #3597, opened 2019-07-17
Ribbon default client properties file

sabareeshkkanan:
Default ribbon properties can be now used by specifying ribbon. without specifying client name


ryanjbaxter:
As far as I know this is already possible, you can specify properties without the client name.

spencergibb:
You can, but not for the beans created. I added that feature

sabareeshkkanan:
> As far as I know this is already possible, you can specify properties without the client name.

This is not really happening when tested for this feature

spencergibb:
ribbon is no longer supported

PR #3609, opened 2019-07-29
ConditionalOnMissingBean on formBodyWrapperFilter, debugFilter and servlet3...

rzukow:
What do you think about adding some ConditionalOnMissingBean annotation to pre filters?
The problem we are facing is DebugFilter that is calling request.getParameter(DEBUG_PARAMETER.get()) just to check the value of debug and ignore it, and there is no easy way to disable it. 

I added some other beans as described here: https://www.anaplan.com/blog/using-zuul-in-production/ , but the article is outdated anyway.


rzukow:
Done

rzukow:
> Please remove the name parameter

Done.

PR #3617, opened 2019-08-01
Provide support for docker port mapping

yu-private:
When deploy a project with docker, it is wrong that the spring boot actuator port registered to spring boot admin and spring cloud eurake. 
This port is "server.port", not set in "eurak.instance.non-secure-port" or "eurak.instance.secure-port", This bug will cause many functional exception.

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -165,6 +165,12 @@ public class EurekaClientAutoConfiguration {
 		EurekaInstanceConfigBean instance = new EurekaInstanceConfigBean(inetUtils);
 
 		instance.setNonSecurePort(serverPort);
+		Integer nonSecurePort = env.getProperty("eureka.instance.non-secure-port",
+				Integer.class);
+		if (nonSecurePort != null) {
+			instance.setNonSecurePort(nonSecurePort);
+		}
+		
 		instance.setInstanceId(getDefaultInstanceId(env));
 		instance.setPreferIpAddress(preferIpAddress);
 		instance.setSecurePortEnabled(isSecurePortEnabled);
@@ -174,6 +180,11 @@ public class EurekaClientAutoConfiguration {
 
 		if (isSecurePortEnabled) {
 			instance.setSecurePort(serverPort);
+			Integer securePort = env.getProperty("eureka.instance.secure-port",
+					Integer.class);
+			if (securePort != null) {
+				instance.setSecurePort(securePort);
+			}
 		}
 
 		if (StringUtils.hasText(hostname)) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@yu-private Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3617)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3617) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yu-private Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3617)!

yu-private:
Suppose my program "server.port" has a value of 8080, I put it in a docker and set the port mapping to 20000:8080.
its are now IP:8080 that The values of "instance.statusPageUrl" and "instance.healthCheckUrl". 
In the spring boot Admin will not be able to get the state of the program through them.
Their value should be IP: 20000.

yu-private:
> I'm not sure I want to accept these changes. There are defaults. Getting values out of the environment doesn't make sense when spring boot configuration properties are going to do just that.
> Can you explain your symptoms without trying to provide a solution?


Suppose my program "server.port" has a value of 8080, I put it in a docker and set the port mapping to 20000:8080.
its are now IP:8080 that The values of "instance.statusPageUrl" and "instance.healthCheckUrl".
In the spring boot Admin will not be able to get the state of the program through them.
Their value should be IP: 20000.

yu-private:
> I'm not sure I want to accept these changes. There are defaults. Getting values out of the environment doesn't make sense when spring boot configuration properties are going to do just that.
> Can you explain your symptoms without trying to provide a solution?
> 
> Suppose my program "server.port" has a value of 8080, I put it in a docker and set the port mapping to 20000:8080.
> its are now IP:8080 that The values of "instance.statusPageUrl" and "instance.healthCheckUrl".
> In the spring boot Admin will not be able to get the state of the program through them.
> Their value should be IP: 20000.

 spring boot configuration set "eureka.instance.non-secure-port" is replaced with "server.port" by code

ryanjbaxter:
So why not change those values via configuration properties in a profile or overriding them using environment variables in the docker container?

yu-private:
> So why not change those values via configuration properties in a profile or overriding them using environment variables in the docker container?

spring-boot-admin

I don't know that which properties should be set.
Can you tell me?

ryanjbaxter:
The same properties you set in your application properties, they are just uppercased and the `.` are replaced with `_`

yu-private:
> The same properties you set in your application properties, they are just uppercased and the . are replaced with _

what properties can set value who underline in the images?

spencergibb:
We don't maintain spring boot admin

spencergibb:
I don't know what values those are

spencergibb:
Regardless, I don't think we want to merge this

PR #3653, opened 2019-09-15
Added reactive support for service discovery

TYsewyn:


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaDiscoveryClientConfiguration.java
```diff
@@ -40,19 +42,27 @@
...
 @Configuration
 @EnableConfigurationProperties
 @ConditionalOnClass(EurekaClientConfig.class)
 @ConditionalOnProperty(value = "eureka.client.enabled", matchIfMissing = true)
 @ConditionalOnDiscoveryEnabled
+@ConditionalOnBlockingDiscoveryEnabled
 public class EurekaDiscoveryClientConfiguration {
 
 	@Bean
 	public Marker eurekaDiscoverClientMarker() {
```
OlgaMaciaszek: is this still going to be used anywhere? If not, maybe we should deprecate it?

OlgaMaciaszek: Also I have seen a discussion regarding the marker use is going on [here](https://github.com/spring-cloud/spring-cloud-zookeeper/pull/235/files#r324941423) - whatever the conclusion of this discussion is, please make sure to apply it here as well.

spring-cloud-netflix-eureka-client/src/main/resources/META-INF/spring.factories
```diff
@@ -3,8 +3,14 @@ org.springframework.cloud.netflix.eureka.config.EurekaClientConfigServerAutoConf
...
-org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration
+org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration,\
+org.springframework.cloud.netflix.eureka.reactive.EurekaReactiveDiscoveryClientConfiguration
 
 org.springframework.cloud.bootstrap.BootstrapConfiguration=\
 org.springframework.cloud.netflix.eureka.config.EurekaDiscoveryClientConfigServiceBootstrapConfiguration
 
+## Discovery Client Configuration
+org.springframework.cloud.client.discovery.EnableDiscoveryClient=\
+org.springframework.cloud.netflix.eureka.EurekaDiscoveryClientConfiguration, \
```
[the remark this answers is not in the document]
spencergibb: good catch. I think we need to have a separate auto config that still uses the marker bean

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/reactive/EurekaReactiveDiscoveryClientConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	public EurekaReactiveDiscoveryClient eurekaReactiveDiscoveryClient(
+			EurekaClient client, EurekaClientConfig clientConfig) {
+		return new EurekaReactiveDiscoveryClient(client, clientConfig);
+	}
+
+	@Bean
+	@ConditionalOnClass(ReactiveHealthIndicator.class)
```
OlgaMaciaszek: Apparently `ConditionalOnClass(xxx.class)` should not be added on methods;  move over a `@Configuration` class;

spencergibb: yeah, either make this use the String attribute or move this to a nested class.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/reactive/EurekaReactiveDiscoveryClient.java
```diff
@@ -0,0 +1,139 @@
...
+	@Override
+	public int getOrder() {
+		return clientConfig instanceof Ordered ? ((Ordered) clientConfig).getOrder()
+				: ReactiveDiscoveryClient.DEFAULT_ORDER;
+	}
+
+	/**
+	 * An Eureka-specific {@link ServiceInstance} implementation.
+	 */
+	public static class EurekaServiceInstance implements ServiceInstance {
```
spencergibb: Let's consolidate in a later PR. This is an exact duplicate of the `EurekaDiscoveryClient` one.

PR #3657, opened 2019-09-19
Fix RestTemplateEurekaHttpClient status update endpoint

### LINKED ISSUE #3571: restTemplate for eureka client failed to change status
I use Eureka Client with RestTemplate instead of Jersey, and send a post request to change status by use the rest api. but i find it doesn't work. And I find the client code in `RestTemplateEurekaHttpClient`:
```
public EurekaHttpResponse<Void> statusUpdate(String appName, String id, InstanceStatus newStatus, InstanceInfo info) {
        String urlPath = this.serviceUrl + "apps/" + appName + '/' + id + "?status=" + newStatus.name() + "&lastDirtyTimestamp=" + info.getLastDirtyTimestamp().toString();
        ResponseEntity<Void> response = this.restTemplate.exchange(urlPath, HttpMethod.PUT, (HttpEntity)null, Void.class, new Object[0]);
        return EurekaHttpResponse.anEurekaHttpResponse(response.getStatusCodeValue()).headers(headersOf(response)).build();
    }
```

but the server-side code like this  `InstanceResource`:
```
@PUT
    @Path("status")
    public Response statusUpdate(@QueryParam("value") String newStatus, @HeaderParam("x-netflix-discovery-replication") String isReplication, @QueryParam("lastDirtyTimestamp") String lastDirtyTimestamp) {
        try {
            if (this.registry.getInstanceByAppAndId(this.app.getName(), this.id) == null) {
                logger.warn("Instance not found: {}/{}", this.app.getName(), this.id);
                return Response.status(Status.NOT_FOUND).build();
            } else {
                boolean isSuccess = this.registry.statusUpdate(this.app.getName(), this.id, InstanceStatus.valueOf(newStatus), lastDirtyTimestamp, "true".equals(isReplication));
                if (isSuccess) {
                    logger.info("Status updated: " + this.app.getName() + " - " + this.id + " - " + newStatus);
                    return Response.ok().build();
                } else {
                    logger.warn("Unable to update status: " + this.app.getName() + " - " + this.id + " - " + newStatus);
                    return Response.serverError().build();
                }
            }
        } catch (Throwable var5) {
            logger.error("Error updating instance {} for status {}", this.id, newStatus);
            return Response.serverError().build();
        }
    }
```
I debug the code and find the request like this:
```
http://127.0.0.1:8761/eureka/apps/TEST.SERVICE.LOCAL/127.0.0.1:test.service.local:8106?status=OUT_OF_SERVICE&lastDirtyTimestamp=1560997426052
```
Is this a bug？

ryanjbaxter:
Fixes #3571

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateEurekaHttpClient.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateEurekaHttpClient.java
@@ -113,7 +113,7 @@ public class RestTemplateEurekaHttpClient implements EurekaHttpClient {
 	@Override
 	public EurekaHttpResponse<Void> statusUpdate(String appName, String id,
 			InstanceStatus newStatus, InstanceInfo info) {
-		String urlPath = serviceUrl + "apps/" + appName + '/' + id + "?status="
+		String urlPath = serviceUrl + "apps/" + appName + '/' + id + "/status?value="
 				+ newStatus.name() + "&lastDirtyTimestamp="
 				+ info.getLastDirtyTimestamp().toString();
 
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -144,8 +144,8 @@ public class EurekaClientAutoConfiguration {
 				.parseBoolean(getProperty("eureka.instance.secure-port-enabled"));
 
 		String serverContextPath = env.getProperty("server.servlet.context-path", "/");
-		int serverPort = Integer
-				.parseInt(env.getProperty("server.port", env.getProperty("port", "8080")));
+		int serverPort = Integer.parseInt(
+				env.getProperty("server.port", env.getProperty("port", "8080")));
 
 		Integer managementPort = env.getProperty("management.server.port", Integer.class);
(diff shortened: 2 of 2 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateEurekaHttpClient.java
```diff
@@ -113,7 +113,7 @@ public RestTemplateEurekaHttpClient(RestTemplate restTemplate, String serviceUrl
 	@Override
 	public EurekaHttpResponse<Void> statusUpdate(String appName, String id,
 			InstanceStatus newStatus, InstanceInfo info) {
-		String urlPath = serviceUrl + "apps/" + appName + '/' + id + "?status="
+		String urlPath = serviceUrl + "apps/" + appName + '/' + id + "/status?value="
```
spencergibb: :man_facepalming:

ryanjbaxter:
@spencergibb yes there is

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/http/RestTemplateEurekaHttpClientTest.java#L106

Problem is the way we were making the request before the request was actually hitting the mock endpoint here https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/http/EurekaServerMockApplication.java#L134

yuwzho:
> LGTM. Is there a test so we can catch a regression like this? We can add it later if needed.

Should the wrong URL returns HTTP 404 or 405? Or 400 if has request parameter `status`?

Issue #3679, opened 2019-10-23
Wrong Content-Encoding set when response is coming from `RequestContext::getResponseBody` and origin response is gzipped


jebeaudet:
Hi

When modifying the origin response and setting the client response as a String using `RequestContext::setResponseBody`, the `Content-Encoding` header gets wrongly set to what was set on the origin response. This caused a non-gzipped stream marked as gzip with the `Content-Encoding` header being sent back to the client.

I'll submit a PR shortly.

spencergibb:
Closing in favor of #3680

PR #3680, opened 2019-10-23
Proper handling for Content-Encoding and String response body

jebeaudet:
Fixes gh-3679

I've already signed the contributor's agreement.

I also fixed some test that we're having invalid preconditions and we're using the `streamContent` variable incorrectly for the use case they were testing.


jebeaudet:
I also removed the useless `OutputStream::flush` in `SendResponseFilter`. Section 5.6 of the Servlet spec dictates that the servlet container should do this. This clears out some WARN logs that I've been seeing regurlarly where a client has given up on a connection and the backend tries to write on it.

spencergibb:
will you rebase to get the new circle config so your pr build will run?

PR #3734, opened 2020-01-14
Proper handling for Content-Encoding and String response body

jebeaudet:
Following up on #3680 since I messed it up there sorry.

Fixes gh-3679


spencergibb:
Sorry for the long delay. This needs to be submitted on the 2.2.x branch. I'm a little worried that existing tests needed to be changed for this to work. Can you explain why you had to change them?

jebeaudet:
@spencergibb I've updated the base branch

The fix I made showed that some of the UTs were improperly coded in the first place. When testing against the client response data stream, you need to specify `true` in the `createFilter` method otherwise it was setting the content in the "override" body.

FWIW, I've been running this version of the filter for ~6 months in production with no issue.

jebeaudet:
Closing due to lack of feedback

Issue #3655, opened 2019-09-18
Eureka instance is using non-secure port to construct `secureHealthCheckUrl` 


LittleBaiBai:
Given eureka instance configuration as follows:
```
eureka:
  instance: 
    secure-port: 443
    secure-port-enabled: true
    non-secure-port: 80
    non-secure-port-enabled: false
````
The generated `EurekaInstanceConfigBean` uses the `non-secure-port` to construct both `healthCheckUrl` and `secureHealthCheckUrl`. (https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/metadata/DefaultManagementMetadataProvider.java#L58)

**Comment by spencergibb** [comment]
In fact the url construction only uses `server.port` or `port`. See https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java#L149

**Comment by enhbat** [comment]
I'm using only secure ports:
```
eureka:
  instance:
    nonSecurePortEnabled: false
    securePortEnabled: true
    securePort: ${server.port}
```
this gives exception:
```
java[2580]: 2019-09-19 20:01:01.093 ERROR 2580 --- [nio-8765-exec-3] o.a.coyote.http11.Http11NioProtocol      : Error reading request, ignored
java[2580]: java.lang.NullPointerException: null
java[2580]: at org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper.getSslSupport(NioEndpoint.java:1392) ~[tomcat-embed-core-9.0.24.jar!/:9.0.24]
java[2580]: at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:853) ~[tomcat-embed-core-9.0.24.jar!/:9.0.24]
java[2580]: at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1593) ~[tomcat-embed-core-9.0.24.jar!/:9.0.24]
java[2580]: at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49) ~[tomcat-embed-core-9.0.24.jar!/:9.0.24]
[...]

PR #3739, opened 2020-01-24
secureHealthCheckUrl fix issue-3655

epm-dev-priporov:
fix  for #3655 issue

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java
@@ -147,6 +147,8 @@ public class EurekaClientAutoConfiguration {
 		String serverContextPath = env.getProperty("server.servlet.context-path", "/");
 		int serverPort = Integer.parseInt(
 				env.getProperty("server.port", env.getProperty("port", "8080")));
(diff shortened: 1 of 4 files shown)


epm-dev-priporov:
Also thre is a weird code in class _EurekaClientAutoConfiguration_

            instance.setNonSecurePort(serverPort);
		instance.setInstanceId(getDefaultInstanceId(env));
		instance.setPreferIpAddress(preferIpAddress);
		instance.setSecurePortEnabled(isSecurePortEnabled);
		if (StringUtils.hasText(ipAddress)) {
			instance.setIpAddress(ipAddress);
		}

		if (isSecurePortEnabled) {
			instance.setSecurePort(serverPort);
		}

serverPort  is used in _isSecurePortEnabled ==  true_ case.
Is that ok?

spencergibb:
After a review and trying to polish I'm not sure this is all correct. What is the priority between secureport, port and management port?

epm-dev-priporov:
I would say that each thing called securitySomething has more higher priority than just something) In this particular case it will be not obvious if security port will be overwritten y any other, but this is just my opinion

spencergibb:
except if management port is defined, actuator is on that port. port and secure port only matter if management port isn't defined as far as the health check url is involved.

epm-dev-priporov:
In my opinion that is not obvious behaviour. As a developer I would like to get some info about that during spring boot starting, even maybe I would expect to get an exception with some details

spencergibb:
@epm-dev-priporov I'm unsure how to proceed. The management port has nothing to do with the secure port of the main context.

Issue #3764, opened 2020-04-02
SimpleRouteLocator strips unexpected paths


OskarD:
**Describe the bug**
spring-boot-starter-parent 2.1.13.RELEASE
Inherently:
<spring-cloud-connectors.version>2.0.7.RELEASE</spring-cloud-connectors.version>
spring-cloud-dependencies: Greenwich.SR5

When a route path starts with the same string that happens to be the dispatcher servlet path, that part is stripped when adjusting the path. I don't know, but I presume it is adjusted in case someone accidentally puts their actual servlet path instead of the relative path.

My suggested solution is to check if the full path segment (From / to /) is equal before stripping it from the path.

**Sample**
application.yml
```
server:
  servlet:
    context-path: /foo

zuul:
  routes:
    foobar:
      path: /foo-bar/**
      url: https://example.com/foo-bar-service
```
Log:

> WARN [app,...] 39689 --- [nio-8221-exec-6] o.s.c.n.z.f.pre.PreDecorationFilter      : No route found for uri: /foo-bar/example

Because in https://github.com/spring-cloud/spring-cloud-netflix/blob/baa50c136add5b1be861e04dbf704d3e4ddf6d8c/spring-cloud-netflix-core/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java#L199:
```
// this.dispatcherServletPath = "/foo", path = "/foo-bar/**"
...
if (!this.dispatcherServletPath.equals("/")
		&& path.startsWith(this.dispatcherServletPath)) {
	adjustedPath = path.substring(this.dispatcherServletPath.length());
// adjustedPath = "-bar/**"
```

spencergibb:
This module has entered [maintenance mode](https://github.com/spring-cloud/spring-cloud-netflix#modules-in-maintenance-mode). This means that the Spring Cloud team will no longer be adding new features to the module. We will fix blocker bugs and security issues, and we will also consider and review small pull requests from the community.

OskarD:
@spencergibb I'm trying to figure out how to get past this... Do you have any tips? I upgraded to the latest spring-boot-starter-parent and Hoxton version, but it's still pulling in a version that contains this code and it's blocking my progress

PR #3765, opened 2020-04-03
Avoid unwanted path stripping

OskarD:
#3764

### DIFF
--- a/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
+++ b/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
@@ -216,7 +216,7 @@ public class SimpleRouteLocator implements RouteLocator, Ordered {
 		if (RequestUtils.isDispatcherServletRequest()
 				&& StringUtils.hasText(this.dispatcherServletPath)) {
 			if (!this.dispatcherServletPath.equals("/")
-					&& path.startsWith(this.dispatcherServletPath)) {
+					&& path.startsWith(this.dispatcherServletPath + "/")) {
 				adjustedPath = path.substring(this.dispatcherServletPath.length());
 				log.debug("Stripped dispatcherServletPath");
 			}
--- a/spring-cloud-netflix-zuul/src/test/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocatorTests.java
+++ b/spring-cloud-netflix-zuul/src/test/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocatorTests.java
@@ -16,19 +16,20 @@
 
 package org.springframework.cloud.netflix.zuul.filters;
 
+import com.netflix.zuul.context.RequestContext;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.springframework.boot.test.system.OutputCaptureRule;
+import org.springframework.cloud.netflix.zuul.filters.ZuulProperties.ZuulRoute;
+
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-
-import org.springframework.boot.test.system.OutputCaptureRule;
-import org.springframework.cloud.netflix.zuul.filters.ZuulProperties.ZuulRoute;
-
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hamcrest.Matchers.containsString;
+import static org.springframework.cloud.netflix.zuul.filters.support.FilterConstants.IS_DISPATCHER_SERVLET_REQUEST_KEY;
 
 /**
  * @author Tom Cawley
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@OskarD Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3765)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3765) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@OskarD Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-netflix&pullRequestId=3765)!

OskarD:
I'm guessing the test will fail with something like this:

> Expecting:
 <Route{id='foo-bar', fullPath='/foo-bar/1', path='/1', location='foo', prefix='/foo-bar', retryable=false, sensitiveHeaders=[], customSensitiveHeaders=false, prefixStripped=true}>
to be equal to:
 <Route{id='foobar', fullPath='/foo-bar/1', path='/1', location='foo', prefix='/foo-bar', retryable=false, sensitiveHeaders=[], customSensitiveHeaders=false, prefixStripped=true}>
but was not.

But I don't understand why? Why does the ID suddenly have the hyphen?

OskarD:
> I'm guessing the test will fail with something like this:
> 
> > Expecting:
> > <Route{id='foo-bar', fullPath='/foo-bar/1', path='/1', location='foo', prefix='/foo-bar', retryable=false, sensitiveHeaders=[], customSensitiveHeaders=false, prefixStripped=true}>
> > to be equal to:
> > <Route{id='foobar', fullPath='/foo-bar/1', path='/1', location='foo', prefix='/foo-bar', retryable=false, sensitiveHeaders=[], customSensitiveHeaders=false, prefixStripped=true}>
> > but was not.
> 
> But I don't understand why? Why does the ID suddenly have the hyphen?

It came true!!! But why?

spencergibb:
Looking good. Is there an edge case where just `/foo` (ie context) is routed thru zuul? Maybe test that and see what happens.

OskarD:
> Looking good. Is there an edge case where just `/foo` (ie context) is routed thru zuul? Maybe test that and see what happens.

I don't understand why `SimpleRouteLocator::getRoute` is mocked, but it seems to be working.

I looked closer at the ID thing, and I get it now... although it is a little confusing

OskarD:
> Looking good. Is there an edge case where just `/foo` (ie context) is routed thru zuul? Maybe test that and see what happens.

I don't understand why `SimpleRouteLocator::getRoute` is mocked, but it seems to be working.

I looked closer at the ID thing, and I get it now... although it is a little confusing

spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/filters/SimpleRouteLocator.java
```diff
@@ -216,7 +216,7 @@ private String adjustPath(final String path) {
 		if (RequestUtils.isDispatcherServletRequest()
 				&& StringUtils.hasText(this.dispatcherServletPath)) {
 			if (!this.dispatcherServletPath.equals("/")
-					&& path.startsWith(this.dispatcherServletPath)) {
+					&& path.startsWith(this.dispatcherServletPath + "/")) {
```
spencergibb, line 219: This never gets exercised by your new tests since `RequestUtils.isDispatcherServletRequest()` is always false.

spencergibb: Adding
```
		RequestContext.getCurrentContext().set(IS_DISPATCHER_SERVLET_REQUEST_KEY, true);
// test
		RequestContext.getCurrentContext().remove(IS_DISPATCHER_SERVLET_REQUEST_KEY);
```
reveals that `test_getMatchingRouteFilterWithEqualServletPath()` fails with a null route

OskarD: What should be the expected behavior?
As I understand it, the purpose of this method is to remove the `dispatcherServletPath` from the path if someone puts it there. Isn't it better to just let it stay wrong, so the user can fix their bad configuration?

spencergibb: a dispatcher path of `/bar` and a routepath of `/bar/**` will only match if the request path is `/bar/bar`. The tests need to be updated to add the request context setup I mentioned above and either add another test that matches a default route (ie dispatcher path of `/bar` and routepath of `/**`) or change the test that returns null to the values I described.

OskarD: I have changed the test to expect null

OskarD:
Why is codecov/project failing? When I click details I get "Unable to find report content in the storage archive."

spencergibb:
Don't worry about codecov

OskarD:
@spencergibb will this be merged?

OskarD:
@spencergibb ping

spencergibb:
I haven't had a chance to review it since the updates.

spencergibb:
The new tests still don't fail if I remove the change to `SimpleRouteLocator`. Without this I can't tell if there is a regression or now.

Issue #3737, opened 2020-01-15
Remove RestTemplate eureka client support


spencergibb:
Replace with WebClient as default depends on #3736

PR #3779, opened 2020-04-17
Adds opt-in webclient support for eureka client

Haybu:
Eureka HTTP client to use either `RestTemplate` or `WebClient` depending on a new `eureka.client.webclient-support` property. When the value of this prop is `true` then the `WebClient` would be used otherwise the `RestTemplate` would be used. `RestTemplate` would be the default if the property is not set.

fixes gh-3737

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientConfigBean.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientConfigBean.java
@@ -413,6 +413,13 @@ public class EurekaClientConfigBean implements EurekaClientConfig, Ordered {
 	 */
 	private int order = 0;
 
+	/**
+	 * A temporary property to switch between using either RestTemplate or WebClient
+	 * support.
+	 * @return
+	 */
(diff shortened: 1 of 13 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientConfigBean.java
```diff
@@ -413,6 +413,13 @@
 	 */
 	private int order = 0;
 
+	/**
+	 * A temporary property to switch between using either RestTemplate or WebClient
+	 * support.
+	 * @return
+	 */
+	private boolean webclientSupport = false;
```
spencergibb, line 421: let's make the property `eureka.client.webclient.enabled` and move the definition to https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-client/src/main/resources/META-INF/additional-spring-configuration-metadata.json rather than here in the config bean.

Haybu: Done.

Issue #3795, opened 2016-09-26
Use REST API for registration in Eureka


whboyd:
It appears that in discovery-first mode, the config client is not updating its location information for the config server when that data changes in eureka (i.e. when a config server instance is removed or added in eureka). Is this currently expected behavior, or am I doing something wrong?

my bootstrap config:

```
spring:
    cloud:
        config:
            enabled: true
            failFast: true
            discovery:
                enabled: true
eureka:
    client:
        serviceUrl:
            defaultZone: [[my eureka defaultZone]]
```

When I take down the first config server instance, the configserver healthcheck in the app begins to fail despite there being a second working node registered with eureka. The same occurs if I restart the config server node with a new password in the eureka metadataMap, in which case the client keeps trying to use the old password.

PR #3796, opened 2020-05-14
Uses http to look up config server from eureka

spencergibb:
This eliminates the oddness of having multiple discovery client beans and related behavior.

Fixes gh-3795

See https://github.com/spring-cloud/spring-cloud-gateway/issues/1514


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaDiscoveryClientConfigServiceBootstrapConfiguration.java
```diff
@@ -36,13 +52,52 @@
...
+			search = SearchStrategy.CURRENT)
+	public EurekaClientConfigBean eurekaClientConfigBean() {
+		return new EurekaClientConfigBean();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public WebClientEurekaHttpClient configDiscoveryEurekaHttpClient(
+			EurekaClientConfigBean config) {
+		List<String> urls = EndpointUtils.getServiceUrlsFromConfig(config, "unknown",
```
spencergibb, line 68: @twicksell or @troshko111 I put "unkown" here, is there a better default?

troshko111: Based on https://github.com/Netflix/eureka/blob/master/eureka-client/src/main/java/com/netflix/discovery/endpoint/EndpointUtils.java#L373 seems like there's no functional difference but specifying `default` here will avoid that warn log.

spencergibb: Thanks!

PR #3894, opened 2020-09-21
Set errexit output

OlgaMaciaszek:


spencergibb:
Seems ok. What is the issue this fixes?

OlgaMaciaszek:
Output was not being properly logged during pipeline builds.

spencergibb (approved):

Issue #3973, opened 2021-02-26
Zuul adds "Transfer-Encoding: chunked" header to DELETE requests without body


guybydefault:
**Describe the bug**
Please provide details of the problem, including the version of Spring Cloud that you
are using. 

Spring Cloud Hoxton.SR10
Spring Boot 2.3.9.RELEASE

Zuul adds "Transfer-Encoding: chunked" header to the DELETE request which we don't expect in the DELETE requests: they should not have body so "Transfer-Encoding" is redundant.

Delete request is as simple as this:
curl --location --request DELETE 'zuul_url'
 
Neither header "Transfer-Encoding" nor "Content-Length" headers were present in DELETE request before it is proxied from client to the service by Zuul.

Moreover, since I'm using Payara application server for deploying my service, it throws me 400 BAD_REQUEST due to the presence of "Transfer-Encoding: chunked" header.

iburakov:
Faced the same problem (independently, in a different project). 

After several hours of debugging, I can say that nothing is setting the Transfer-Encoding header in Zuul / Hystrix / Ribbon code. 
[org.apache.http.protocol.RequestContent](https://android.googlesource.com/platform/external/apache-http/+/refs/heads/lollipop-mr1-dev/src/org/apache/http/protocol/RequestContent.java#89) is a "culprit". 

It's even casually mentioned in a comment deep inside spring-cloud-netflix-ribbon: 

https://github.com/spring-cloud/spring-cloud-netflix/blob/92cb1a7aa0471cc3286b9936511dd5f12ff6cdca/spring-cloud-netflix-ribbon/src/main/java/org/springframework/cloud/netflix/ribbon/apache/RibbonApacheHttpRequest.java#L62-L64

I'd be glad if someone explained the reason why they do so. OkHttp seem to behave the same way. WildFly doesn't treat DELETE requests with "transfer-encodning: chunked" as bad after all, so it may be an issue for Payara. 

A workaround is to explicitly set the Zuul request Content-Length to 0 somehow (is it possible with a "pre" filter?) or to switch to the deprecated Ribbon RestClient by setting property `ribbon.restclient.enabled` to `true`.

spencergibb:
PRs welcome

OlgaMaciaszek:
Fixed.

Issue #1042, opened 2016-05-22
411 Content Length required issue


beku8:
For some reason, my home ISP throws ```411 Content length required``` error when I try to post something to the cloud rest endpoint thru locally running zuul proxy. 
So I tried setting ```content-length``` header manually from a custom filter like:
```ctx.addZuulRequestHeader("content-length", ctx.getRequest().getHeader("Content-Length"));```

But it throws 
```
com.netflix.zuul.exception.ZuulException: Forwarding error...
Caused by: org.apache.http.ProtocolException: Content-Length header already present
	at org.apache.http.protocol.RequestContent.process(RequestContent.java:95)
	at org.apache.http.protocol.ImmutableHttpProcessor.process(ImmutableHttpProcessor.java:131)
	at org.apache.http.protocol.HttpRequestExecutor.preProcess(HttpRequestExecutor.java:165)
	at org.apache.http.impl.client.DefaultRequestDirector.execute(DefaultRequestDirector.java:485)
	at org.apache.http.impl.client.AbstractHttpClient.doExecute(AbstractHttpClient.java:863)
```

I'm using Zuul 1.0.28. Am I doing anything wrong? I'm not even sure if the original problem would disappear if I manage to set this header manually. Any suggestion, anything would be very much appreciated.

p.s: ISP allows my post request when its done thru Feign.

This issue is originally posted at https://github.com/Netflix/zuul/issues/223.

**Comment by spencergibb** [comment]
I'm unsure what you are saying is happening? You have some kind of proxy at home that requests go through?

**Comment by beku8** [comment]
No. We have our microservices deployed to the cloudfoundry and we connect to it using locally running zuul to develop the UI section. For some weird reason one of the main ISP in our country (Mongolia) throws 411 error.
Anyways I'm just trying to set the Content-Lenght header manually, hoping this will fix the problem. Probably this is more related to apache's http library?

**Comment by beku8** [comment]
Hello @spencergibb, I just investigated the /trace endpoint to see whats going on with this issue. 

Here's the request from browser to the server:
```
info: {
  method: "PUT",
  path: "/api/core/rest/location/amarbayasgalant",
  headers: {
    request: {
      host: "localhost:9000",
      connection: "keep-alive",
      content-length: "663",
      pragma: "no-cache",
      cache-control: "no-cache",
      authorization: "...authorization omitted",
      accept: "application/json, text/plain, */*",
      origin: "http://localhost:9000",
[...]

PR #4002, opened 2021-05-30
Fix transfer encoding is set to chunked by settin entity contentLength to 0 for DELETE requests

beatfreaker:
Fix for #3973 

When there is no body in request and content-length is not set,
apache http client sets transfer encoding to chunked.

This change will set content-length to 0 for DELETE request with no body
and as a result transfer encoding wont be set for such requests.

See gh-1042,  gh-3973


OlgaMaciaszek (changes requested):
@beatfreaker Thanks for submitting this fix. LGTM. Please add your full name and surname with `@author` tag to the javadocs of the classes you've changed and this will be good to merge.

OlgaMaciaszek:
Fixes gh-3973.

OlgaMaciaszek (approved):
LGTM

PR #4009, opened 2021-07-05
Bump eureka-client version

SeraphC:
Bump eureka-client version to get rid of underlying xstream vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2021-29505


SeraphC:
Unsure what's going on here. Test fails with or without my change. Some guidance would be appreciated, failing that I'll close this PR.

PR #4012, opened 2021-07-09
Fix spring-cloud-netflix-zuul 2.2.x compatibility with Spring Boot 2.5

clementdenis:
Spring Boot 2.5 removed the deprecated method org.springframework.boot.web.servlet.error.ErrorController#getErrorPath, making spring-cloud-netflix-zuul unable to run on it because of [this usage](https://github.com/spring-cloud/spring-cloud-netflix/blob/2.2.x/spring-cloud-netflix-zuul/src/main/java/org/springframework/cloud/netflix/zuul/web/ZuulHandlerMapping.java#L87).

This is fixed by using ServerProperties.getError().getPath() instead.


spencergibb:
The 2.x line is not compatible with boot 2.5 and we have no plans to make it so

clementdenis:
That's too bad, because it seems to be the only thing that is preventing this module to work on Boot 2.5.x

I'll keep working with my fork then!

PR #4019, opened 2021-07-29
Added public modifier for JUnit4 classes and methods

robertmcnees:
Public modifiers were removed from a few classes when upgrading to Unit Jupiter.  When we use the ClassPathExclusions it still requires JUnit4 to run, and thus classes and methods must have the public modifier to be executed.


OlgaMaciaszek (approved):
LGTM.

Issue #4056, opened 2021-12-26
Publish custom port to eureka service registry from eureka client


an0nh4x0r:
Eureka client by default publishes `${server.port}` to service registery i.e. eureka server. There needs to be a property where we can publish custom port to service registry 

There needs to be a property called `${eureka.instance.custom-port}` which can be used to set custom port which will then get published to the eureka server ... instead of `${server.port}`

In my organization , I've override a bean called [`eurekaInstanceConfigBean`](https://github.com/spring-cloud/spring-cloud-netflix/blob/main/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java#L125) basically at this line here [in class `EurekaClientAutoConfiguration`](https://github.com/spring-cloud/spring-cloud-netflix/blob/main/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfiguration.java#L133) I'm checking for custom port property and if it's present I'm using custom port instead of `${server.port}` and publishing that port instead. It works great for me

I would like to get this property support officially from the library

an0nh4x0r:
My pull request got closed. :( 
non-secure-port does not work.

OlgaMaciaszek:
Hello @an0nh4x0r, why do you need this functionality? Could you describe your use-case in more details?

an0nh4x0r:
Hey @OlgaMaciaszek ... We have our own cloud in my company where we run custom built containers(not docker) ... We basically have two ports i.e. first inner port of which spring boot is running, second outer port through which traffic gets rerouted to this inner port in which spring boot app runs ... I have access to outer port number in container through enviornment variable ... I was looking for ways to publish this outer port to eureka service registery. In my company I see people who are using zookeeper as service registery have option to publish custom port to zookeeper ... I was looking to get same functionality natively from Eureka client.

OlgaMaciaszek:
@an0nh4x0r thanks for the details. Will discuss it with the team and get back to you.

OlgaMaciaszek:
[...]

PR #4057, opened 2021-12-26
Added support for eureka to publish custom port to registry

an0nh4x0r:
This pull request is for issue #4056


spencergibb:
server.port is just the default. Use eureka.instance.non-secure-port to customize

an0nh4x0r:
non-secure-port doesn't work. See the code itself.

spencergibb:
I know what the code does. It is used as a default only. 
If you'd like us to spend some time investigating, please take the time to provide a [complete, minimal, verifiable sample](https://stackoverflow.com/help/mcve) (something that we can unzip attached to this issue or git clone, build, and deploy) that reproduces the problem.

Issue #674, opened 2015-11-30
Document the eureka context path.


sfussenegger:
I'm currently trying to get Eureka server running with URL discovery and it proofed to be a bit tricky.

According to the [doc](https://github.com/Netflix/eureka/wiki/Configuring-Eureka-in-AWS-Cloud#configuring-eips-using-dns) the `eureka.eurekaServer.context` (= `eureka.client.eurekaServerURLContext`) should be set to `eureka/v2`. This doesn't work though. The reason is that the Jersey ServletContainer is [mapped](https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaServerConfiguration.java#L169) at `/eureka/*`. Configuring context to `eureka/v2` results in requests for `/eureka/v2/apps/{appId}` which Jersey matches against a pattern of `/{version}/apps/{appId}`. This pattern therefore doesn't match. 

Leaving out the `/eureka` prefix doesn't work as the path won't be filtered by the Jersey filter. A possible workaround is to simply set the context to `eureka`. This results in `eureka` being treated as the `{version}` which interestingly [is allowed](https://github.com/Netflix/eureka/blob/master/eureka-core/src/main/java/com/netflix/eureka/Version.java#L31).

The fix for this issue is to set Jersey's `PROPERTY_FILTER_CONTEXT_PATH` to `/eureka`:

```
    @Bean
    public FilterRegistrationBean jerseyFilterRegistration(
            javax.ws.rs.core.Application eurekaJerseyApp) {
        FilterRegistrationBean bean = new FilterRegistrationBean();
        bean.setFilter(new ServletContainer(eurekaJerseyApp));
        bean.setOrder(Ordered.LOWEST_PRECEDENCE);
        bean.setUrlPatterns(Collections
                .singletonList(EurekaServerConfigBean.DEFAULT_PREFIX + "/*"));
        bean.setInitParameters(Collections.singletonMap(ServletContainer.PROPERTY_FILTER_CONTEXT_PATH, EurekaServerConfigBean.DEFAULT_PREFIX));
        return bean;
    }
```

This is obviously a major change so I'm not sure this is going to happen. Anyway, since I couldn't find any documentation on this use case though, I believe this is at least a documentation issue.

nvj98:
I can work on this documentation issue, thanks!

igarash1:
[...]

Issue #402, opened 2015-06-18
Eureka: server ignores the "version" path param in REST requests


brenuart:
As per Netflix's documentation (https://github.com/Netflix/eureka/wiki/Eureka-REST-operations), Eureka REST operations have the following form:

    /<context>/<version>/<REST operation>

For example:

    POST /eureka/v2/apps/appID
    DELETE /eureka/v2/apps/appID/instanceID
    ...

These examples will fail with SpringCloud's version of the Eureka server:  the _version_ path param CANNOT be specified. However, the following will work (same as above but without the _v2_ version):

    POST /eureka/apps/appID
    DELETE /eureka/apps/appID/instanceID
    ...

In fact, the server considers the first path element (i.e. _/eureka_) as the version argument when handling the request. Since "eureka" is not a valid/supported version (only _v1_ and _v2_ are supported currently), it defaults to version 2 and proceeds with the rest of the request.

REST handling is implemented via Jersey which is integrated via a servlet filter matching requests with pattern `/eureka/*`. When the request is handled over to Jersey, the `/eureka` prefix is not stripped and is therefore considered as the first path element by the Jersey resources. The Jersey resource corresponding to the above example is mapped as follows (see the `@Path` annotation):

    @Path("/{version}/apps")
    @Produces({"application/xml", "application/json"})
    public class ApplicationsResource {
        ...

This problem can be solved by integrating Jersey as a _servlet_ instead of a _servlet filter_.
However, this will force people to update their existing configuration to include the version in the server URL. Change from:

    (1) eureka.client.serviceUrl.defaultZone = http://myeurekaserver:8761/eureka/

to 

    (2) eureka.client.serviceUrl.defaultZone = http://myeurekaserver:8761/eureka/v2/

(Note: maybe we could add some transparent redirection from (1) to (2) in the eureka server if no version is specified)

At the end, the question is **whether or not SpringCloud's version of the Eureka server should expose the _version_ in the urls**. The client is v2-compatible and this is unlikely Netflix releases a version 3 on this code base (they are working on Eureka2 instead)...

What is your opinion?

spencergibb:
That's a great question.

brenuart:
So better to leave it asis and ignore the case?
It works after all...

spencergibb:
Not sure, I'd like to hear @dsyer's opinion.  I was the one who put `/eureka` in the path.

dsyer:
[...]

PR #4077, opened 2022-03-15
Documents the eureka version in the context path (#674)

igarash1:
Currently, specifying the version of the Spring Cloud Eureka server as a path param isn't supported, yet it can be included as the server URL. This means, if you want to set the version to `v2` explicitly, as [the Netflix official doc](https://github.com/Netflix/eureka/wiki/Deploying-Eureka-Servers-in-EC2) explains, you cannot specify the version by setting the context path (`eurekaServerURLContext` in the Sprint Cloud Eureka) to `eureka/v2`, but you can by setting `defaultZone` to `http://localhost:8761/eureka/v2`.

This PR adds the documentation about it.

Fixes gh-674
Related issues: gh-402


OlgaMaciaszek:
Fixes gh-674.

igarash1:
@OlgaMaciaszek Sure!

PR #4079, opened 2022-03-16
Documents the eureka version in the context path (#674)

igarash1:
(This is the revision 2 of gh-4077)
Currently, specifying the version of the Spring Cloud Eureka server as a path param isn't supported, yet it can be included as the server URL. This means, if you want to set the version to `v2` explicitly in path parameters, you cannot specify the version by setting the context path (`eurekaServerURLContext` in the Spring Cloud Eureka) to `eureka/v2` as [the Netflix official doc](https://github.com/Netflix/eureka/wiki/) explains, but you can by setting `defaultZone` to `http://localhost:8761/eureka/v2`.

This PR adds the documentation about it.

Fixes gh-674.


OlgaMaciaszek (approved):
Thanks for the PR, @igarash1. LGTM.

PR #4086, opened 2022-04-10
Fixed 'eureka datacenter and environment not displayed on the dashboard'

### LINKED ISSUE #4032: Eureka datacenter and environment not displayed on the Eureka dashboard
I am using Eureka server - version 3.0.3 and have configured the datacenter and environment information using the below configuration:

```
eureka:
  datacenter: Mumbai
  environment: QA
  server:
    enable-self-preservation: false
  instance:
    lease-renewal-interval-in-seconds: 30
    lease-expiration-duration-in-seconds: 90
  client:
    register-with-eureka: false
    fetch-registry: false
```
I do not see the Datacenter and Environment information in the Eureka dashboard. In the older versions of Eureka, the information used to be displayed. Is there any change in the configuration with the current release?

weixsun:
Fixes #4032

### DIFF
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
@@ -48,6 +48,7 @@ import org.springframework.web.bind.annotation.RequestMethod;
 /**
  * @author Spencer Gibb
  * @author Gang Li
+ * @author Weix Sun
  */
 @Controller
 @RequestMapping("${eureka.dashboard.path:/}")
@@ -58,8 +59,20 @@ public class EurekaController {
 
 	private ApplicationInfoManager applicationInfoManager;
 
+	private final EurekaProperties eurekaProperties;
+
+	/**
+	 * @deprecated in favour of
+	 * {@link EurekaController#EurekaController(ApplicationInfoManager, EurekaProperties)}
+	 */
+	@Deprecated
 	public EurekaController(ApplicationInfoManager applicationInfoManager) {
+		this(applicationInfoManager, null);
+	}
+
+	public EurekaController(ApplicationInfoManager applicationInfoManager, EurekaProperties eurekaProperties) {
 		this.applicationInfoManager = applicationInfoManager;
+		this.eurekaProperties = eurekaProperties;
 	}
 
 	@RequestMapping(method = RequestMethod.GET)
@@ -116,8 +129,14 @@ public class EurekaController {
 	private void populateHeader(Map<String, Object> model) {
 		model.put("currentTime", StatusResource.getCurrentTimeAsString());
 		model.put("upTime", StatusInfo.getUpTime());
-		model.put("environment", "N/A"); // FIXME:
-		model.put("datacenter", "N/A"); // FIXME:
+		if (eurekaProperties != null) {
+			model.put("environment", eurekaProperties.getEnvironment());
+			model.put("datacenter", eurekaProperties.getDatacenter());
+		}
+		else {
+			model.put("environment", "N/A");
+			model.put("datacenter", "N/A");
+		}
 		PeerAwareInstanceRegistry registry = getRegistry();
 		model.put("registry", registry);
 		model.put("isBelowRenewThreshold", registry.isBelowRenewThresold() == 1);
--- a/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaProperties.java
+++ b/spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaProperties.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2013-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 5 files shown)


spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
```diff
@@ -57,6 +58,9 @@
 
 	private ApplicationInfoManager applicationInfoManager;
 
+	@Autowired
+	private EurekaDeploymentProperties eurekaDeploymentProperties;
```
weixsun, line 62: The reason why constructor injection is not used is because I don’t want to break the existing test case code structure, and because I run the current test case, I found that there are some problems in the test case of this project. When these problems are solved, I very happy to use constructor injection and update relevant test code.

OlgaMaciaszek:
Thanks, @weixsun - will take a look today.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDeploymentProperties.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Objects;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Configuration properties for the Eureka deployment.
+ *
+ * @author Weix Sun
+ */
+@ConfigurationProperties("eureka")
```
OlgaMaciaszek, line 28: Please change the suffix to `eureka.deployment`.

OlgaMaciaszek (changes requested):
Hello @weixsun, thanks for submitting the PR. In general, it looks good. However, please submit this PR against `3.1.x` branch instead of `main` so that this change is also included in the `2021.x` release train. Also, please add your full name with `@author` tag to any classes you've modified. Also, I have added some comments - please address them.

OlgaMaciaszek:
Fixes gh-4032.

weixsun:
@OlgaMaciaszek Thanks for the clear and effective code review, I resubmitted as requested.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
```diff
@@ -58,10 +59,19 @@ public class EurekaController {
 
 	private ApplicationInfoManager applicationInfoManager;
 
+	private EurekaDeploymentProperties eurekaDeploymentProperties;
```
OlgaMaciaszek, line 62: Let's add `final`. and set to `null` in the deprecated constructor.

OlgaMaciaszek (changes requested):
Thanks @weixsun - it's nearly good to go -  we just need to be able to handle null `eurekaDeploymentProperties`; have added comments - please address them.

weixsun:
@OlgaMaciaszek It's done.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDeploymentProperties.java
weixsun: Everyone has been using `eureka.environment` and `eureka.datacenter` for a long time, and now that we've changed the prefixes, should we remind everyone in the doc or in the release notes?

weixsun: > Please change the suffix to `eureka.deployment`.

Perhaps we can do the configuration item expiration specification in `spring-configuration-metadata.json` like Spring Boot and recommend the configuration item with the `eureka. deployment` prefix instead.

Is it worth it？😂

OlgaMaciaszek: You're right - sorry for the confusion, but given what  you've pointed out, I think we should switch it back to `eureka` prefix and then rename the class to `EurekaProperties` . We might want to do some additional work on all SC Netflix properties prefixes later on to ensure more consistency, but that will have to be done on the `main` branch.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaController.java
```diff
@@ -58,8 +59,17 @@ public class EurekaController {
 
 	private ApplicationInfoManager applicationInfoManager;
 
+	private final EurekaDeploymentProperties eurekaDeploymentProperties;
+
+	@Deprecated
```
OlgaMaciaszek, line 64: Let's also add a javadoc with `@deprecated in favour of `{@link ...}` tag.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDeploymentProperties.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Objects;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Configuration properties for the Eureka deployment.
+ *
+ * @author Weix Sun
+ */
+@ConfigurationProperties("eureka.deployment")
```
OlgaMaciaszek, line 28: Let's switch as described in this comment: https://github.com/spring-cloud/spring-cloud-netflix/pull/4086#discussion_r849338707 - I'm sorry for asking you to change it twice, but you're right here.

OlgaMaciaszek (changes requested):
@weixsun have added 2 more comments - please take a look.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/EurekaDeploymentProperties.java
weixsun: > I think we should switch it back to `eureka`prefix and then rename the class to `EurekaProperties`

Ok, Let me do a quick fix

OlgaMaciaszek (approved):
LGTM

PR #4118, opened 2022-08-28
Move it around to make sure the correct order

MoonLord-LM:
Move it around to make sure the correct order


spencergibb:
Can you explain why you want to do this

MoonLord-LM:
The code structure:  
image

And the pom.xml:  
image

Maybe I'm a little obsessive-compulsive. 🤣

MoonLord-LM:
> @MoonLord-LM Thanks for submitting the PR, however, the order seems ok as it is - the starter is usually the entry point to using these dependencies.

ok

PR #4142, opened 2022-11-28
Changes made in configbean.java

MallikharjunaTeja:
Made Changes in EurekaConfigBean.java. Only Instance Id is taken into consideration for the log file.


spencergibb:
Can you explain why you want to remove a field from the `toString()`?

ankit-gautam23 (approved):
LGTM

PR #4240, opened 2024-01-22
Remove function and use PropertyResolver

ryanjbaxter:
This relates to https://github.com/spring-cloud/spring-cloud-config/pull/2375

And reverts part of #4175

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaConfigServerBootstrapper.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaConfigServerBootstrapper.java
@@ -17,18 +17,16 @@
 package org.springframework.cloud.netflix.eureka.config;
 
 import java.util.Collections;
-import java.util.List;
 
 import com.netflix.discovery.shared.transport.EurekaHttpClient;
-import org.apache.commons.logging.Log;
 
 import org.springframework.boot.BootstrapContext;
 import org.springframework.boot.BootstrapRegistry;
 import org.springframework.boot.BootstrapRegistryInitializer;
 import org.springframework.boot.context.properties.bind.BindHandler;
 import org.springframework.boot.context.properties.bind.Binder;
-import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.config.client.ConfigClientProperties;
+import org.springframework.cloud.config.client.ConfigServerConfigDataLocationResolver.PropertyResolver;
 import org.springframework.cloud.config.client.ConfigServerInstanceProvider;
 import org.springframework.cloud.configuration.TlsProperties;
 import org.springframework.cloud.netflix.eureka.EurekaClientConfigBean;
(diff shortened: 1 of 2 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaConfigServerBootstrapper.java
```diff
@@ -45,52 +43,39 @@ public void initialize(BootstrapRegistry registry) {
...
-			// This should never be called now but is here for backward
-			// compatibility
-			return apply(serviceId, null, null, null);
-		}
+	private static PropertyResolver getPropertyResolver(BootstrapContext context) {
+		return context.getOrElseSupply(PropertyResolver.class,
+				() -> new PropertyResolver(context.get(Binder.class), context.getOrElse(BindHandler.class, null)));
+	}
 
+	private static Boolean getDiscoveryEnabled(BootstrapContext bootstrapContext) {
```
pantherdd, line 74: Though this does not affect us at the moment, I'd suggest to make every bootstrapper's `isEnabled` function `public` ([similar to the Zookeeper implementation](https://github.com/spring-cloud/spring-cloud-zookeeper/pull/328#discussion_r1458959524)) and give them a consistent name. People who try to customize these will thank you 😊

ryanjbaxter: I can make it public, but I am not going to change the name to avoid more breaking changes

pantherdd: @ryanjbaxter Would it really be a breaking change to rename a method that was `private`, i.e. nothing could have used it outside the class?

pantherdd: @ryanjbaxter Would it really be a breaking change to rename a method that was `private`, i.e. nothing could have used it outside the class until now? I'd think that right now is the ideal point in time to decide what the best name for these `isEnabled`-type methods would be, and to rename all of them to the chosen name, right when they are made `public`.
(However, I understand that this is not really technically necessary for any reason, it'd just be a nice touch to have a consistent name for the same functionality.)

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaConfigServerBootstrapper.java
```diff
@@ -45,52 +43,39 @@ public void initialize(BootstrapRegistry registry) {
...
-			// This should never be called now but is here for backward
-			// compatibility
-			return apply(serviceId, null, null, null);
-		}
+	private static PropertyResolver getPropertyResolver(BootstrapContext context) {
+		return context.getOrElseSupply(PropertyResolver.class,
+				() -> new PropertyResolver(context.get(Binder.class), context.getOrElse(BindHandler.class, null)));
+	}
 
+	public static Boolean getDiscoveryEnabled(BootstrapContext bootstrapContext) {
```
pantherdd, line 74: Unnecessary auto-boxing to `Boolean` (and I now realized, that's why this implementation has a method named `get*` instead of `is*`).

Issue #4236, opened 2023-12-29
Improve documentation on required HTTP client dependencies


jagc93:
I am experiencing issues with the Eureka Client.

Spring Boot version `3.2.1` or `3.2.2`.
Spring cloud `2023.0.0`
java version 17
Maven

I created from Spring Initializr again, 2 basic microservices to check whether the issue lies within my existing microservices or not, and the error persists.

The first one would be the discovery-server

pom.xml

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>

	<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>3.2.1</version>
		<relativePath/> <!-- lookup parent from repository -->
	</parent>

	<groupId>co.prueba</groupId>
	<artifactId>discovery-server</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<name>discovery-server</name>
	<description>Discobery Server</description>

	<properties>
		<java.version>17</java.version>
		<spring-cloud.version>2023.0.0</spring-cloud.version>
	</properties>

	<dependencies>
		<dependency>
			<groupId>org.springframework.cloud</groupId>
			<artifactId>spring-cloud-starter-netflix-eureka-server</artifactId>
		</dependency>

		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-devtools</artifactId>
			<scope>runtime</scope>
			<optional>true</optional>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-configuration-processor</artifactId>
			<optional>true</optional>
		</dependency>
		<dependency>
			<groupId>org.projectlombok</groupId>
			<artifactId>lombok</artifactId>
			<optional>true</optional>
		</dependency>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-test</artifactId>
			<scope>test</scope>
		</dependency>
	</dependencies>

	<dependencyManagement>
		<dependencies>
			<dependency>
[...]

PR #4242, opened 2024-01-25
Improve documentation on HTTP clients used by EurekaClient.

OlgaMaciaszek:
Fixes gh-4236.

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-netflix.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-netflix.adoc
@@ -267,25 +267,31 @@ Do not use the `EurekaClient` in a `@PostConstruct` method or in a `@Scheduled` 
 It is initialized in a `SmartLifecycle` (with `phase=0`), so the earliest you can rely on it being available is in another `SmartLifecycle` with a higher phase.
 ====
 
-==== EurekaClient with Jersey
+==== Underlying HTTP clients
 
-By default, EurekaClient uses Spring's `RestTemplate` for HTTP communication.
+`EurekaClient` uses either `RestTemplate`, `WebClient` or `JerseyClient` under the hood. In order to use the `EurekaClient`, you need to have one of the supported HTTP clients on your classpath.
+
+To use `RestTemplate`, add `spring-boot-starter-web` to your dependencies. To use `WebClient`, add `spring-boot-starter-webflux` to your dependencies. If both `RestTemplate` and `WebClient` are on the classpath, when `eureka.client.webclient.enabled` is set to `true`, `WebClient` will be used. Otherwise, `RestTemplate` will be used.
+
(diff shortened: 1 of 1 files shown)


OlgaMaciaszek:
@Buzzardo please review the changes.

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -267,25 +267,31 @@ Do not use the `EurekaClient` in a `@PostConstruct` method or in a `@Scheduled`
 It is initialized in a `SmartLifecycle` (with `phase=0`), so the earliest you can rely on it being available is in another `SmartLifecycle` with a higher phase.
 ====
 
-==== EurekaClient with Jersey
+==== Underlying HTTP clients
+
+`EurekaClient` uses either `RestTemplate`, `WebClient` or `JerseyClient` under the hood. In order to use the `EurekaClient`, you need to have one of the supported HTTP clients on your classpath.
```
Buzzardo, line 272: Change to:

`EurekaClient` uses one of `RestTemplate`, `WebClient`, or `JerseyClient` under the hood. To use the `EurekaClient`, you need to have one of the supported HTTP clients on your classpath.

That usage of "either" isn't, strictly speaking, wrong, but it is about a century out of date. Also, as a general rule, you rarely need "In order" - you can often leave it out and have the same meaning, as here.

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -267,25 +267,31 @@ Do not use the `EurekaClient` in a `@PostConstruct` method or in a `@Scheduled`
...
+
+To use `RestTemplate`, add `spring-boot-starter-web` to your dependencies. To use `WebClient`, add `spring-boot-starter-webflux` to your dependencies. If both `RestTemplate` and `WebClient` are on the classpath, when `eureka.client.webclient.enabled` is set to `true`, `WebClient` will be used. Otherwise, `RestTemplate` will be used.
```
Buzzardo, line 274: Remove the comma before "when" and change both instances of "will be" to "is".

The comma isn't wrong (the sentence works either way), but we aim to remove anything we don't strictly need. The change from "will be" to "is" is in keeping with our practice of sticking with the indefinite present tense (sometimes called the simple present tense), which generally lets us write shorter sentences. Again, you're not wrong, but our conventions do it differently.

Just for reference, here is the Spring Style Guide: https://github.com/spring-io/spring-asciidoctor-backends/blob/main/guides/style-guide.adoc

Buzzardo (changes requested):

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
OlgaMaciaszek: Thanks, @Buzzardo.

Issue #4133, opened 2022-11-07
Migrate out of deprecated WebClient exchange() method


OlgaMaciaszek:

ayushdubey755:
@OlgaMaciaszek Can I work on this issue? is there anything else should I know for the first time contributing to this project, I would love to work on it since I have previously used the spring framework and currently working on spring boot.

OlgaMaciaszek:
Sure, @ayushdubey755. Go ahead.

PR #4261, opened 2024-02-19
Clean up eureka-client deprecations

ZIRAKrezovic:
Fixes: #4133

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java
@@ -210,11 +210,6 @@ public class RestTemplateTransportClientFactory implements TransportClientFactor
 		}
 
 		@Override
-		public int getRawStatusCode() throws IOException {
-			return response.getRawStatusCode();
-		}
-
-		@Override
 		public String getStatusText() throws IOException {
 			return response.getStatusText();
 		}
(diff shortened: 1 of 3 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java
```diff
@@ -209,11 +209,6 @@ public HttpStatusCode getStatusCode() throws IOException {
 			return response.getStatusCode();
 		}
 
-		@Override
```
OlgaMaciaszek, line 212: We cannot remove public methods in a non-major release. Please revert this change.

OlgaMaciaszek (changes requested):
Hello @ZIRAKrezovic. Thanks for the PR. Have added a comment. Please address.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java
ZIRAKrezovic: There is a default implementation on the interface level that delegates to getStatusCode(), so technically nothing is removed.

ZIRAKrezovic: But in any case, I have reverted this change.

ZIRAKrezovic:
Hi @OlgaMaciaszek. Thanks for the review, I believe I have addressed your feedback.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java
OlgaMaciaszek: You're right. Have missed that the implementation is same anyhow. You can remove it.

ZIRAKrezovic: I have re-removed it.

OlgaMaciaszek (approved):
LGTM.

PR #4270, opened 2024-03-24
Collect metrics for application instances

### LINKED ISSUE #3685: Eureka server micrometer metrics for registered services
Enhancement

Metrics on Eureka server, both `/actuator/metrics` and eg. `/actuator/prometheus` provide only generic jvm and http server metrics. I would like to see metrics of the services registered with Eureka as well, something similar to what is displayed on the Eureka server status page that comes with eureka-server. This would be very useful to include in common dashboards like Grafana to see over time, add alerts for down services etc. 

For example gauges like this (for Prometheus):

`eureka_server_instances{application="my-app", instanceId="host1:my-app:9180", status="DOWN"}
`

Other interesting metrics could be Renews (last min) and status of replicas.

I'm thinking of writing something like this myself for our Eureka server, but if there is some existing implementation somewhere I'd like to know - or get some input whether others would find this useful enough to add to spring-cloud-netflix-eureka-server.

heowc:
Fixes #3685


spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Override
+	public void bindTo(MeterRegistry meterRegistry) {
+		instanceRegistry.getApplications().getRegisteredApplications().stream()
+				.flatMap(application -> application.getInstances().stream()).forEach(instanceInfo -> {
+					Gauge.builder("eureka.server.application.instances", () -> 1L)
+							.description("Information about application instances registered on the Eureka server.")
+							.tags(List.of(Tag.of("application", instanceInfo.getAppName()),
+									Tag.of("id", instanceInfo.getId()), Tag.of("host", instanceInfo.getHostName()),
```
OlgaMaciaszek, line 46: When it comes to actual metrics, I'll request review from @jonatan-ivanov and you should go along with his suggestions. My preliminary thoughts are that "id" should probably be removed, cause we'll have many singular values, as each instance has a unique id. Also, I'm not sure port is going to be helpful as a tag. While, we might want to slice the data based on how many instances are running on each host, I'm not sure what slicing by port would be useful for.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
```diff
@@ -0,0 +1,53 @@
...
+import com.netflix.eureka.registry.PeerAwareInstanceRegistry;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+/**
+ * @author heowc
+ */
+public class EurekaInstanceMetricsBinder implements MeterBinder {
```
OlgaMaciaszek, line 31: Please rename to `EurekaInstanceMeterBinder`.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Objects;
+
+import com.netflix.eureka.registry.PeerAwareInstanceRegistry;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+/**
+ * @author heowc
```
OlgaMaciaszek, line 29: Please add your full name and surname with `@author` tag in the javadoc of any new or modified class. Please add description at the  beginning of the javadoc of any new class. Please add `@since 4.1.1` to the javadoc of every new class.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.cloud.netflix.eureka.server.EurekaServerAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+
+/**
+ * @author heowc
+ */
+@ConditionalOnBean(MeterRegistry.class)
```
OlgaMaciaszek, line 33: This should be `@ConditionalOnClass`.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class,
+		EurekaServerAutoConfiguration.class })
+class EurekaInstanceMetricsAutoConfiguration {
```
OlgaMaciaszek, line 36: This feature should be opt-in, i.e. `@ConditionalOnProperty`.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMetricsBinderTests.java
```diff
@@ -0,0 +1,71 @@
...
+
+	private static final String APP_NAME = "MY-APP-NAME";
+
+	private static final String HOST_NAME = "my-host-name";
+
+	private static final String INSTANCE_ID = "my-host-name:8008";
+
+	private static final int PORT = 8008;
+
+	@SpyBean(PeerAwareInstanceRegistry.class)
```
OlgaMaciaszek, line 49: Why are you using a `@SpyBean` here? Please switch to actual autowired bean.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMetricsBinderTests.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.boot.test.mock.mockito.SpyBean;
+import org.springframework.cloud.netflix.eureka.server.metrics.EurekaInstanceMetricsBinder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.cloud.netflix.eureka.server.InstanceRegistryTests.getInstanceInfo;
+import static org.springframework.cloud.netflix.eureka.server.InstanceRegistryTests.getLeaseInfo;
+
+@SpringBootTest(classes = InstanceRegistryTests.TestApplication.class,
+		webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, value = { "spring.application.name=eureka",
+				"logging.level.org.springframework.cloud.netflix.eureka.server.InstanceRegistry=DEBUG" })
```
OlgaMaciaszek, line 38: Please remove debug logs.

OlgaMaciaszek (changes requested):
Hello @heowc, thanks for providing the PR. I have  added some comments - please address. When it comes to actual metrics, I'll also request review from @jonatan-ivanov who specialises in it. Please also document the feature in `docs/modules/ROOT/pages/spring-cloud-netflix.adoc`.

OlgaMaciaszek:
@jonatan-ivanov Could you please review the metrics-related code?

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsAutoConfiguration.java
jonatan-ivanov: I think it should be both `@ConditionalOnBean` and `@ConditionalOnClass`.
E.g.:
```java
@ConditionalOnBean(MeterRegistry.class)
@ConditionalOnClass(MeterRegistry.class)
@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class, EurekaServerAutoConfiguration.class })
```

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
```diff
@@ -0,0 +1,53 @@
...
+public class EurekaInstanceMetricsBinder implements MeterBinder {
+
+	private final PeerAwareInstanceRegistry instanceRegistry;
+
+	EurekaInstanceMetricsBinder(PeerAwareInstanceRegistry instanceRegistry) {
+		this.instanceRegistry = Objects.requireNonNull(instanceRegistry);
+	}
+
+	@Override
+	public void bindTo(MeterRegistry meterRegistry) {
```
jonatan-ivanov, line 40: This will be called once, when the binder is registered and never again.
How do you expect registrations to be updated?

I _think_ you can simulate this in your tests by doing something like this:
```java
InstanceInfo instanceInfo = getInstanceInfo(...);
binder.bindTo(meterRegistry);
instanceRegistry.register(instanceInfo, false);
assert... // You will see zero instances but there should be one
```

There are multiple ways to resolve this and I can help but first I want to make sure we are on the same page and I'm not missing something.
(We can use a `ScheduledTaskExecutor`, the `Gauge` itself to refresh the list of instances and also a `MultiGauge` can be helpful in some cases.)

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
```diff
@@ -0,0 +1,53 @@
...
+		instanceRegistry.getApplications().getRegisteredApplications().stream()
+				.flatMap(application -> application.getInstances().stream()).forEach(instanceInfo -> {
+					Gauge.builder("eureka.server.application.instances", () -> 1L)
+							.description("Information about application instances registered on the Eureka server.")
+							.tags(List.of(Tag.of("application", instanceInfo.getAppName()),
```
jonatan-ivanov, line 45: You can use `Tags` instead of `List<Tag>` here.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
jonatan-ivanov: I agree with the "id", if there will be lot of instances, it can be a problem.
I think host+port together can be problematic too because of the same reason: together they are unique per instance (just like the id) so both together have the same problem.

On the other hand, some users might need these so I think it would be useful to inject some kind of a `TagsProvider` that can give you a `Tags` instance for an `instanceInfo` object. There is an issue with this: if we will not provide instance-level tags (dimensions) somehow we need to count them since right now `1` is hardcoded (which works right now since this creates a separate Gauge for every instance).

Also, what everyone thinks about adding `status` as a tag (can be `UP`/`DOWN`)?

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/InstanceRegistryTests.java
```diff
@@ -147,14 +147,14 @@ void testRenew() {
 		assertThat(event2.getInstanceInfo()).isEqualTo(instanceInfo2);
 	}
 
-	private LeaseInfo getLeaseInfo() {
+	static LeaseInfo getLeaseInfo() {
```
jonatan-ivanov, line 150: I think it would make sense moving these out to a separate class for tests so that test classes will not depend on each other (after we figured out the binder).

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMetricsBinderTests.java
```diff
@@ -0,0 +1,71 @@
...
+	private InstanceRegistry instanceRegistry;
+
+	@Autowired
+	private MeterRegistry meterRegistry;
+
+	@Autowired
+	private EurekaInstanceMetricsBinder binder;
+
+	@Test
+	void testMetrics() {
```
jonatan-ivanov, line 59: I would add a few more tests:
1. When the instance is registered after the `bindTo` call
2. When multiple instances registered (after the `bindTo` call)
3. When multiple instances registered and at lest one is de-registered (after the `bindTo` call)
4. When multiple instances registered and at least one is de-registered and then re-registered (after the `bindTo` call)

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
OlgaMaciaszek: No, I think it's not been done on purpose. We should probably create these metrics during events related to instance lifetime, such as `EurekaInstanceRegisteredEvent` and `EurekaInstanceCancelledEvent`.

OlgaMaciaszek: or use `Gauge` to update results with information from new `application.getInstances()` periodically.

OlgaMaciaszek:
Thanks a lot @jonatan-ivanov. @heowc please address the comments and don't hesitate to let us know if you need additional help.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
OlgaMaciaszek: Right, the status is there already.

OlgaMaciaszek (changes requested):
Thanks @heowc. Waiting for the rest of the changes we've discussed.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
jonatan-ivanov: Ooops, I did not see `status`, sorry. :)
Btw to increase the readability you can also do this:
```java
.description("Information about application instances registered on the Eureka server.")
.tag("application", instanceInfo.getAppName())
.tag("status", instanceInfo.getStatus().name()))
.register(meterRegistry);
```

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
jonatan-ivanov: >We should probably create these metrics during events related to instance lifetime, such as EurekaInstanceRegisteredEvent and EurekaInstanceCancelledEvent.

This seem to be the most promising one to me.
@heowc do you want to look into this? I think if you listen to these events, you can use `MutiGauge` to do the recording. This might play well with the user-defined tags/aggregation we discussed above: if you pass the list of instances, the user-defined logic can return a list of `MultiGauge.Row` instances that define the tags and the number of instances too. Please let me know if you need help/have questions. You can find the docs here: [Multi Gauge](https://docs.micrometer.io/micrometer/reference/concepts/gauges.html#_multi_gauge) and there is some sample code here: [`MultiGaugeSample`](https://github.com/micrometer-metrics/micrometer/blob/main/samples/micrometer-samples-core/src/main/java/io/micrometer/core/samples/MultiGaugeSample.java).

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
heowc: Added `EurekaInstanceTagProvider` 😄

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
heowc: Ooops..! Thank you.

Let me look into this.

heowc: While examining `EurekaInstanceRegisteredEvent` and `EurekaInstanceCancelledEvent`, I found something somewhat strange. I've written a related PR, please check it. #4274  🤔

OlgaMaciaszek:
Thanks for updates, @heowc. It seems you're still working on it since there are TBD comments and not all the changes we've discussed have been implemented, I will not be reviewing it yet. Could you mark the PR as draft and then turn it back into regular PR once you're ready for review?  Also, make sure to pull the changes from `main` and resolve existing conflicts.  You can also add a comment for us to look at it. If you have additional questions, also let us know. Please make sure you tag me in any comments so that we don't miss them.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMetricsBinderTests.java
heowc: @jonatan-ivanov 
I switched to a different way of collecting metrics. However, I added tests similar to those mentioned.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
heowc: @OlgaMaciaszek @jonatan-ivanov 

I made a lot of changes to use `MultiGuage`.

- I changed it to listen for events like `EurekaInstanceXXXEvent`.
- I changed to using `TagsProvider` to group and count things with equivalent tags.
- However, since `status` is required, I always add it independently of the `TagsProvider`.

If you have any other good ideas, feel free to let me know 🙏

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsBinder.java
heowc: Added 🙇

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,154 @@
...
+	@Autowired
+	private MeterRegistry meterRegistry;
+
+	private InstanceInfo instanceInfo;
+
+	private InstanceInfo instanceInfo2;
+
+	@BeforeEach
+	void setup() {
+		this.instanceRegistry.clearRegistry();
```
OlgaMaciaszek, line 62: Please avoid unnecessary use of  `this.` keyword.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/FixtureEurekaInstances.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.eureka.server;
+
+import com.netflix.appinfo.InstanceInfo;
+import com.netflix.appinfo.LeaseInfo;
+
+public final class FixtureEurekaInstances {
```
OlgaMaciaszek, line 22: Please rename to `EurekaInstancesFixture`

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+ * Auto-configuration for Eureka Instance metrics.
+ *
+ * @author wonchul heo
+ * @since 4.1.1
+ */
+@ConditionalOnClass(MeterRegistry.class)
+@ConditionalOnBean(MeterRegistry.class)
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class,
+		EurekaServerAutoConfiguration.class })
+@ConditionalOnProperty(name = "eureka.instance.metrics.enabled", havingValue = "true")
```
OlgaMaciaszek, line 42: Please rename the property to `eureka.server.metrics.enabled`; Please create an `additional-spring-configuration-metadata.json` file with the corresponding entry.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,154 @@
...
+import static org.springframework.cloud.netflix.eureka.server.FixtureEurekaInstances.getInstanceInfo;
+import static org.springframework.cloud.netflix.eureka.server.FixtureEurekaInstances.getLeaseInfo;
+
+/**
+ * @author Wonchul Heo
+ */
+@SpringBootTest(classes = EurekaInstanceMonitorTests.Application.class,
+		webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+		value = { "spring.application.name=eureka", "eureka.instance.metrics.enabled=true" })
+class EurekaInstanceMonitorTests {
```
OlgaMaciaszek, line 46: Please add tests for more than one application. Please add tests for other instance statuses.

OlgaMaciaszek (changes requested):
Thanks @heowc. This looks better. Have added some comments. Please address.

OlgaMaciaszek:
@jonatan-ivanov Could you please review the reworked metrics?

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMonitor.java
```diff
@@ -0,0 +1,80 @@
...
+	public void onApplicationEvent(ApplicationEvent event) {
+		final Map<Tags, Long> aggregatedCounts = instanceRegistry.getApplications().getRegisteredApplications().stream()
+				.flatMap(application -> application.getInstances().stream())
+				.collect(Collectors.groupingBy(this::tags, Collectors.counting()));
+		eurekaInstanceStatuses.register(aggregatedCounts.entrySet().stream()
+				.map(it -> MultiGauge.Row.of(it.getKey(), it.getValue())).collect(Collectors.toList()), true);
+	}
+
+	private Tags tags(InstanceInfo instanceInfo) {
+		return tagProvider.eurekaInstanceTags(instanceInfo).and("status", instanceInfo.getStatus().name());
```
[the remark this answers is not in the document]
heowc: So how about checking whether the `Tags` created in `tagsProvider` include `status`?

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: I thought it was a convention because it was included in other test codes 🤣

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/FixtureEurekaInstances.java
heowc: Fixed

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMetricsAutoConfiguration.java
heowc: Added

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,154 @@
...
+		final Map<Tags, Long> counts = Map.of(tags(instanceInfo), 1L, tags(instanceInfo2), 1L);
+		assertEurekaInstance(counts, counts);
+	}
+
+	private static Tags tags(InstanceInfo instanceInfo) {
+		return Tags.of(Tag.of("application", instanceInfo.getAppName()),
+				Tag.of("status", instanceInfo.getStatus().name()));
+	}
+
+	private void assertEurekaInstance(Map<Tags, Long> instanceRegistryCounts, Map<Tags, Long> meterRegistryCounts) {
```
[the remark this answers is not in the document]
heowc: On second thought, validating the state of `instanceRegistry` seems a bit overkill, let's remove the related code.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: Rewrote the existing tests by adding the application to them.

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMonitor.java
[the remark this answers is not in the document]
OlgaMaciaszek: If it isn't, then we do not tag by it. That's what we're providing in the default. If the users just want to gauge by number of all instances by the app or whichever other attributes they choose.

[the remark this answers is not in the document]
heowc: So what do you think about changing the multi-gauge name? Or is it still enough?


eureka.server.instance.status -> eureka.server.instance


Previously, I wanted status to be included somehow, like its current name.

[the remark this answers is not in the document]
OlgaMaciaszek: maybe `eureka.server.instances` ?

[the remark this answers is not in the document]
heowc: Fixed 👍

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+
+	@Override
+	public void onApplicationEvent(ApplicationEvent event) {
+		final Map<Tags, Long> aggregatedCounts = instanceRegistry.getApplications().getRegisteredApplications().stream()
+				.flatMap(application -> application.getInstances().stream())
+				.collect(Collectors.groupingBy(tagProvider::eurekaInstanceTags, Collectors.counting()));
+		eurekaInstances.register(aggregatedCounts.entrySet().stream()
+				.map(it -> MultiGauge.Row.of(it.getKey(), it.getValue())).collect(Collectors.toList()), true);
+	}
+}
```
jonatan-ivanov, line 74: If you run `./mvnw verify` won't the auto-formatter inserts a newline here?
I think it should, please check and if it does update the PR with that newline since it will cause problems for us in the future.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,166 @@
...
+import static org.springframework.cloud.netflix.eureka.server.EurekaInstancesFixture.getInstanceInfo;
+import static org.springframework.cloud.netflix.eureka.server.EurekaInstancesFixture.getLeaseInfo;
+
+/**
+ * @author Wonchul Heo
+ */
+@SpringBootTest(classes = EurekaInstanceMonitorTests.Application.class,
+		webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+		value = { "spring.application.name=eureka", "eureka.server.metrics.enabled=true" })
+class EurekaInstanceMonitorTests {
```
jonatan-ivanov, line 45: I'm pretty sure the auto-formatter should format this file, could you please run `./mvnw verify` and update the PR with the changes?

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstancesFixture.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.netflix.eureka.server;
+
+import com.netflix.appinfo.InstanceInfo;
+import com.netflix.appinfo.LeaseInfo;
+
+public final class EurekaInstancesFixture {
```
jonatan-ivanov, line 22: Shouldn't this be singular `EurekaInstanceFixture` since it is for one instance?

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -604,6 +604,18 @@ when running a Eureka server you must include these dependencies in your POM or
...
+Starting with version 4.1.1, `MultuGuage` listens events related to Eureka instances and creates and updates Eureka instance information in the metrics registry. By default, this setting is disabled, if you want to enable it, you need to change `eureka.server.metrics.enabled` to `true`.
+
+The multi-gauge is named `eureka.server.instance.status` and has the following tags (value are merged based on tags)
+
+- `application`: application name
+
+- `status`: instance status
+
+You can add additional tags by implementing `EurekaInstanceTagsProvider` separately. However, status is added separately because it is always needed.
+
```
jonatan-ivanov, line 618:
```suggestion
=== Metrics

`EurekaInstanceMonitor` listens to events related to Eureka instance registration and creates/updates `Gauge`s for Eureka instance information Micrometer's `MeterRegistry`. By default, this behavior is disabled. Ff you want to enable it, you need to set `eureka.server.metrics.enabled` to `true`.

By default, the `Gauge`s are named `eureka.server.instances` and have the following tags:

- `application`: application name
- `status`: instance status (`UP`, `DOWN`, `STARTING`, `OUT_OF_SERVICE`, `UNKNOWN`, see: `com.netflix.appinfo.InstanceInfo.InstanceStatus`)

You can add additional tags by injecting your own implementation of `EurekaInstanceTagsProvider`.
```

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/DefaultEurekaInstanceTagsProvider.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.netflix.appinfo.InstanceInfo;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+
+/**
+ * Default implementation for {@link EurekaInstanceTagsProvider}.
+ *
+ * @author Wonchul Heo
+ * @since 4.1.1
```
jonatan-ivanov, line 27: Please fix all the `@since` tags in this PR. Check the issue, Olga set the milestone to `4.1.2` so the `@since` tags should say the same (4.1.1 is already out).

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,166 @@
...
+		fooInstanceInfo2 = getInstanceInfo(FOO_APP_NAME, "my-host-name", "my-host-name:8009", 8009, getLeaseInfo());
+		barInstanceInfo = getInstanceInfo(BAR_APP_NAME, "my-host-name", "my-host-name:8010", 8010, getLeaseInfo());
+		barInstanceInfo2 = getInstanceInfo(BAR_APP_NAME, "my-host-name", "my-host-name:8011", 8011, getLeaseInfo());
+	}
+
+	@Test
+	void testNoRegistration() {
+		assertThatThrownBy(() -> {
+			meterRegistry.get("eureka.server.instances").meter();
+		}).isInstanceOf(MeterNotFoundException.class);
```
jonatan-ivanov, line 75:
```suggestion
		assertThat(meterRegistry.find("eureka.server.instances").gauge()).isNull();
```

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,166 @@
...
+		assertEurekaInstance(counts);
+	}
+
+	private static Tags tags(InstanceInfo instanceInfo) {
+		return Tags.of(Tag.of("application", instanceInfo.getAppName()),
+				Tag.of("status", instanceInfo.getStatus().name()));
+	}
+
+	private void assertEurekaInstance(Map<Tags, Long> meterRegistryCounts) {
+		meterRegistryCounts.forEach((tags, count) -> {
```
jonatan-ivanov, line 154: `{` and `}` are unnecessary here.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
```diff
@@ -0,0 +1,166 @@
...
+				InstanceInfo.InstanceStatus.STARTING, null, false);
+		instanceRegistry.renew(fooInstanceInfo.getAppName(), fooInstanceInfo.getInstanceId(), false);
+		instanceRegistry.renew(barInstanceInfo2.getAppName(), barInstanceInfo2.getInstanceId(), false);
+
+		final Map<Tags, Long> counts = Map.of(tags(fooInstanceInfo), 1L, tags(fooInstanceInfo2), 1L,
+				tags(barInstanceInfo), 1L, tags(barInstanceInfo2), 1L);
+		assertEurekaInstance(counts);
+	}
+
+	private static Tags tags(InstanceInfo instanceInfo) {
```
jonatan-ivanov, line 148: How hard would it be adding an extra test with a custom, user provided mapping that adds a tag that makes every instance unique (e.g.: `port`)?

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+		this.eurekaInstances = MultiGauge.builder("eureka.server.instances")
+				.description("Count of application instances registered with the Eureka server.")
+				.register(meterRegistry);
+	}
+
+	@Override
+	public boolean supportsEventType(Class<? extends ApplicationEvent> eventType) {
+		return EurekaInstanceCanceledEvent.class.isAssignableFrom(eventType)
+				|| EurekaInstanceRegisteredEvent.class.isAssignableFrom(eventType)
+				|| EurekaInstanceRenewedEvent.class.isAssignableFrom(eventType);
```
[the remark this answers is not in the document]
heowc: good idea 👍

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/EurekaInstanceMonitor.java
heowc: I missed it 🤔

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: ditto

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
heowc: Thanks for the great suggestion 😄

spring-cloud-netflix-eureka-server/src/main/java/org/springframework/cloud/netflix/eureka/server/metrics/DefaultEurekaInstanceTagsProvider.java
heowc: 👍

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: Oh.. there was `find`. thank you 👍

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: I will write additional tests 🙏

OlgaMaciaszek: @heowc  Waiting for this one still. Once you've added it, we can go ahead with the merge.

OlgaMaciaszek:
Thanks a lot @jonatan-ivanov.

docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -604,6 +604,17 @@ when running a Eureka server you must include these dependencies in your POM or
 
 Spring Cloud Netflix Eureka Server does not support Spring AOT transformations or native images.
 
+=== Metrics
+
+`EurekaInstanceMonitor` listens to events related to Eureka instance registration and creates/updates `Gauge`s for Eureka instance information Micrometer's `MeterRegistry`. By default, this behavior is disabled. Ff you want to enable it, you need to set `eureka.server.metrics.enabled` to `true`.
```
jonatan-ivanov, line 609: I made typos when I was suggesting docs changes here:
```suggestion
`EurekaInstanceMonitor` listens to events related to Eureka instance registration and creates/updates `Gauge`s for Eureka instance information in Micrometer's `MeterRegistry`. By default, this behavior is disabled. If you want to enable it, you need to set `eureka.server.metrics.enabled` to `true`.
```

jonatan-ivanov (changes requested):

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorTests.java
heowc: Added 😄

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorWithCustomTagsProviderTests.java
```diff
@@ -0,0 +1,102 @@
...
+
+	private InstanceInfo fooInstanceInfo;
+
+	private InstanceInfo fooInstanceInfo2;
+
+	@BeforeEach
+	void setup() {
+		instanceRegistry.clearRegistry();
+		meterRegistry.clear();
+		fooInstanceInfo = getInstanceInfo(FOO_APP_NAME, "my-host-name", "my-host-name:8008", 8008, getLeaseInfo());
```
OlgaMaciaszek, line 62: Let's rename this to `firstInstanceInfo` and `secondInstanceInfo`, and also rename `FOO_APP_NAME` to `APP_NAME`.

OlgaMaciaszek (changes requested):
Thanks @heowc. Have added a comment. Please address. Also, please remove the typos from doc entry that @jonatan-ivanov has indicated.

spring-cloud-netflix-eureka-server/src/test/java/org/springframework/cloud/netflix/eureka/server/EurekaInstanceMonitorWithCustomTagsProviderTests.java
heowc: @OlgaMaciaszek 
Fix it quickly 😄

OlgaMaciaszek (approved):
Thanks @heowc. LGTM.

PR #4163, opened 2023-03-05
Encode url path segments & query parameters

NewAgeCZ:
- Extend test suite with instanceId value that requires encoding
- Fix url encoding in both RestTemplate & WebClient

PR #4291, opened 2024-07-15
Encode url path segments & query parameters

NewAgeCZ:
Rebased upon main, added author tag and edited date strings in licenses
New PR of the old https://github.com/spring-cloud/spring-cloud-netflix/pull/4163

- Extend test suite with instanceId value that requires encoding
- Fix url encoding in both RestTemplate & WebClient


OlgaMaciaszek (approved):
Thanks, @NewAgeCZ. LGTM.

Issue #4294, opened 2024-07-22
Incorrect Default Values in Eureka Appendix Documentation


555cider:
I've discovered incorrect default values in the Eureka's [Common application properties](https://docs.spring.io/spring-cloud-netflix/docs/current/reference/html/appendix.html) documentation.

The documentation file is located at "docs/modules/ROOT/partials/_configprops.adoc". I've attached a corrected version of the file(
[_configprops.adoc.txt](https://github.com/user-attachments/files/16334889/_configprops.adoc.txt)) saved as .txt for uploading.

The corrected items are as follows:

| Name                                                        | Default (before) | Default (after) |
| ----------------------------------------------------------- | ---------------- | --------------- |
| eureka.client.eureka-service-url-poll-interval-seconds      | 0                | 300             |
| eureka.server.a-s-g-update-interval-ms                      | 0                | 300000          |
| eureka.server.a-s-g-cache-expiry-timeout-ms                 | 0                | 600000          |
| eureka.server.e-i-p-binding-retry-interval-ms               | 0                | 300000          |
| eureka.server.e-i-p-binding-retry-interval-ms-when-unbound  | 0                | 60000           |
| eureka.server.renewal-threshold-update-interval-ms          | 0                | 900000          |
| eureka.server.response-cache-update-interval-ms             | 0                | 30000           |
| eureka.server.route53-binding-retry-interval-ms             | 0                | 300000          |
| eureka.server.retention-time-in-m-s-in-delta-queue          | 0                | 180000          |
| eureka.server.delta-retention-timer-interval-in-ms          | 0                | 30000           |
| eureka.server.wait-time-in-ms-when-sync-empty               | 0                | 300000          |
| eureka.client.rest-template-timeout.connect-timeout         | 0                | 180000          |
| eureka.client.rest-template-timeout.connect-request-timeout | 0                | 180000          |
| eureka.client.rest-template-timeout.socket-timeout          | 0                | 180000          |
| eureka.client.rest-template-timeout.idle-timeout            | (absent)         | 1800000         |
| eureka.instance.secure-port-enabled                         | false            | null            |

`eureka.client.rest-template-timeout.idle-timeout` is missing from the original documentation. A description for this property is "Indicates how long a connection can be idle before it is evicted from the connection pool." The behavior of `eureka.instance.secure-port-enabled` set to null is unclear. In the declaration, it's not initialized with false.

OlgaMaciaszek:
[...]

PR #4321, opened 2024-09-06
Fix incorrect default values in docs (#4294)

555cider:


spencergibb:
That file is auto generated, so editing it directly is not the right way to do this

OlgaMaciaszek:
Right. @555cider what seems to happen is that the default values in `@ConfigurationProperties` classes defining those props (for example, `EurekaClientConfigBean`, `EurekaServerConfigBean`), are not explicitly specified, but using math equations with constants instead, for instance: https://github.com/spring-cloud/spring-cloud-netflix/blob/57dd4f06744dba5af4fe4ac8dfb971665fd73f30/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/EurekaClientConfigBean.java#L99, so what you'd need to do is instead of changing that `configprops` file, change the default values in those `@ConfigurationProperties` classes to explicit values (you could add comments such as "5 * MINUTES" where helpful.

OlgaMaciaszek (changes requested):
Thanks @555cider, please adjust the changes as described here: https://github.com/spring-cloud/spring-cloud-netflix/pull/4321#issuecomment-2374741265

OlgaMaciaszek (changes requested):
Thanks, @555cider - the changes look good; have you regenerated the configprops file with the new changes (in order to do that, remove any manual changes in the file and run `./mvnw clean install -Pdocs` locally and the file will be regenerated)? Also, please add your full name and surname with `@author` tag in the javadocs of the classes you've modified and update the dates in the years in the license comment on top of the class to  `-2024`. Finally, could you submit your changes against the `4.1.x` branch instead of `main`?

555cider:
I have applied all the changes you requested except for the regeneration of the configprops file. I was unable to regenerate the file because it involves removing any manual changes and running `./mvnw clean install -Pdocs` locally.

Just in case, I've reverted the _configprops file back to its original state.

Also, I accidentally merged the main branch while clicking around the site. I force pushed to clean up the git history, but if this action is not allowed, I will create a new pull request. In that case, I will modify the branch based on the "4.1.x" branch and submit the request.

OlgaMaciaszek:
Thanks @555cider - waiting for the new PR.

555cider:
Creating a new pull request (#4999) and closing this pull request (#4321).

PR #4363, opened 2024-11-06
Set up RestClient-based TransportClientFactories

OlgaMaciaszek:


docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -98,13 +98,18 @@ NOTE: Because of a limitation in Eureka, it is not possible to support per-serve
 
 If you want to customize the RestTemplate used by the Eureka HTTP Client you may want to create a bean of `EurekaClientHttpRequestFactorySupplier` and provide your own logic for generating a `ClientHttpRequestFactory` instance.
 
-All default timeout-related properties for RestTemplate used by the Eureka HTTP Client are set to 3 minutes (in keeping with Apache HC5 default `RequestConfig` and `SocketConfig`). Therefore, to specify the timeout values, you must specify the value directly with the properties in `eureka.client.rest-template-timeout`. (All timeout properties are in milliseconds.)
+All default timeout-related properties for RestTemplate and RestClient used by the Eureka HTTP Client are set to 3 minutes (in keeping with Apache HC5 default `RequestConfig` and `SocketConfig`). Therefore, to specify the timeout values, you must specify the value directly with the properties in `eureka.client.rest-template-timeout` or `eureka.client.restclient.timeout`. All timeout properties are in milliseconds.
```
ryanjbaxter, line 101: If the idea is to remove `RestTemplate` support in the major we should probably call that out in the documentation.

Also I think it would them make sense to have the new properties be `eureka.client.timeout` rather than use the `rest-template` or `restclient` prefix.

OlgaMaciaszek: Right, will change that.

OlgaMaciaszek: After a second look, I think it's better to keep the `restclient.` prefix for now. We still need to support `RestTemplate`-specific ones for `RestTemplate`, since we can't remove them in a non major and we need something for `RestClient` that people will then not think it's going to work with `RestTemplate` as well, since it will not (if we wanted it to also work for `RestTemplate`, we would then need to have some kind of handling both prefixes for `RestTemplate`, along with an overriding logic, which is probably only going to be confusing for the users), but I will also deprecate the `RestClient`-specific class and prefix and indicate there what the new one will be in the major; wdyt?

spencergibb: Sounds good

ryanjbaxter: Sounds basically switch the prefix in the major then?

OlgaMaciaszek: I think it would be more messy otherwise - if  `restclient` is not in the prefix right now, they may think whatever they change will also apply to `RestTemplate`, but we can't remove the `rest-template` one at this point either, so there we use a different bean altogether.

ryanjbaxter (approved):

Issue #4391, opened 2024-12-12
Disable protocol upgrades for eureka client and server HTTP clients


mamachanko:
**Is your feature request related to a problem? Please describe.**

I would like to disable protocol upgrades for Eureka client and server HTTP client, because with Spring Boot 3.4.0 ...

> Apache HTTP Components have changed defaults in the HttpClient relating to HTTP/1.1 TLS upgrades. Most proxy servers handle upgrades without issue, however, you may encounter issues with Envoy or Istio.

[There's a workaround](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes#apache-http-components-and-envoy
) suggested. However, providing a custom `HttpComponentsClientHttpRequestFactoryBuilder` bean does not affect Eureka's client and server HTTP clients. Afaik that's because they do not consider that bean, but create their own.

**Describe the solution you'd like**

As a first step, I would love a recommendation for how to customize Eureka server and client HTTP clients to disable protocol upgrades. Eventually, maybe in future versions, it would be great if there was a property one could set.

**Describe alternatives you've considered**

I have tried to provide my own `EurekaClientHttpRequestFactorySupplier` which is essentially a copied of the `DefaultEurekaClientHttpRequestFactorySupplier` with:

```java

	@Override
	public ClientHttpRequestFactory get(SSLContext sslContext, @Nullable HostnameVerifier hostnameVerifier) {
		HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
		if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
			httpClientBuilder
					.setConnectionManager(buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
		}
		if (timeoutProperties != null) {
			httpClientBuilder.setDefaultRequestConfig(buildRequestConfig());
		}

		CloseableHttpClient httpClient = httpClientBuilder.build();
		
		// disable protocol upgrades
		HttpComponentsClientHttpRequestFactory requestFactory = ClientHttpRequestFactoryBuilder
				.httpComponents()
				.withDefaultRequestConfigCustomizer((builder) -> builder.setProtocolUpgradeEnabled(false))
				.build();
		
		requestFactory.setHttpClient(httpClient);
		return requestFactory;
	}


```

but calls to the Eureka server still try to upgrade the protocol

```
Accept: application/json, application/*+json
Accept-Encoding: gzip, x-gzip, deflate
[...]

PR #4396, opened 2025-01-27
Control Apache HttpComponents' protocol upgrades w/ a property

mamachanko:
Provides a configuration property

```yaml
eureka:
  client:
    http-components:
      enable-protocol-upgrades: false
```

to make https://github.com/spring-cloud/spring-cloud-netflix/issues/4391 even easier building on top of https://github.com/spring-cloud/spring-cloud-netflix/pull/4394

PR #4399, opened 2025-01-31
Control Apache HttpComponents' protocol upgrades w/ a property

mamachanko:
Provides a configuration property

```yaml
eureka:
  client:
    http-components:
      enable-protocol-upgrades: false
```

to make https://github.com/spring-cloud/spring-cloud-netflix/issues/4391 even easier building on top of https://github.com/spring-cloud/spring-cloud-netflix/pull/4394

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
+++ b/spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2022 the original author or authors.
+ * Copyright 2013-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,10 +27,10 @@ import com.netflix.discovery.AbstractDiscoveryClientOptionalArgs;
 import com.netflix.discovery.EurekaClient;
 import com.netflix.discovery.EurekaClientConfig;
 import com.netflix.discovery.shared.transport.jersey.TransportClientFactories;
+import org.apache.hc.client5.http.config.RequestConfig;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
-
 import org.springframework.aop.framework.Advised;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.boot.SpringApplication;
@@ -51,6 +51,7 @@ import org.springframework.cloud.context.config.ContextRefreshedWithApplicationE
 import org.springframework.cloud.context.refresh.ContextRefresher;
 import org.springframework.cloud.context.scope.GenericScope;
 import org.springframework.cloud.netflix.eureka.config.DiscoveryClientOptionalArgsConfiguration;
+import org.springframework.cloud.netflix.eureka.http.EurekaClientHttpRequestFactorySupplier;
 import org.springframework.cloud.netflix.eureka.serviceregistry.EurekaServiceRegistry;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationListener;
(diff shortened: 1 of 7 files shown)


spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
```diff
@@ -543,6 +545,105 @@ void shouldCloseDiscoveryClient() throws Exception {
...
+
+	@Configuration
+	protected static class TestEmptyRequestCustomizerConfiguration {
+
+		@Bean
+		EurekaClientHttpRequestFactorySupplier.RequestConfigCustomizer emptyRequestCustomizer() {
+			return builder -> {};
+		}
+
+	}
```
mamachanko, line 645: **[thought]:** 

As suggested by you, @OlgaMaciaszek, in https://github.com/spring-cloud/spring-cloud-netflix/pull/4396#discussion_r1935427425 user-provided request customizers override HTTP components configuration by way of `@ConditionalOnMissingBean`.

_(for simplicity I am referring to `EurekaClientHttpRequestFactorySupplier.RequestConfigCustomizer` as plain ærequest customizer")_

Assuming if I have understood your remarks correctly, then I see the following issues. When user-provided request customizers override HTTP components configuration, ...

* Then _any_ request customizer - regardless of whether it controls protocol upgrades or not - will override `eureka.client.http-components.enable-protocol-upgrades`. That's because request customizers are opaque. We can't easily tell what they customize. The test  `TODO_shouldOverrideProtocolUpgrades` demonstrates the issue with an _empty_ request customizer which overrides.

* Then it seems as if it's breaking the priority order respective [externalized configuration](https://docs.spring.io/spring-boot/reference/features/external-config.html). Conventionally, I would expect a configuration property to have the last word, not a bean. Arguably, at a lower level the two become beans all the same. But I find it unintuitive. It is making the mental model more involved, doesn't it? Furthermore, if there are users who have - knowingly or not - have a request customizer, then they may set `eureka.client.http-components.enable-protocol-upgrades` and wonder why its not working as expected.

OlgaMaciaszek: I see your point. Actually, given that it's a customizer type bean, we may want to change it so that we expect >= 1 customiser beans, then we'd add ours, but also pick the user's beans, and we should ensure ours have lower order of precedence than the user-created ones (so possibly should be applied first and then overwritten by user changes where appropriate); wdyt?  CC @spencergibb

OlgaMaciaszek (changes requested):
Thanks @mamachanko. Have added some comments.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
```diff
@@ -543,6 +545,105 @@ void shouldCloseDiscoveryClient() throws Exception {
...
+
+		@Bean
+		EurekaClientHttpRequestFactorySupplier.RequestConfigCustomizer disableProtocolUpgrades() {
+			return builder -> builder.setProtocolUpgradeEnabled(false);
+		}
+
+	}
+
+	@Test
+	void TODO_shouldOverrideProtocolUpgrades() {
```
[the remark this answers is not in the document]
mamachanko: @OlgaMaciaszek  I'd like to keep it, until we resolve https://github.com/spring-cloud/spring-cloud-netflix/pull/4399/files#r1937493242 because it's exemplifying the problem.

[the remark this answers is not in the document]
OlgaMaciaszek: Sure, the review is done with the final version in thought.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/EurekaClientAutoConfigurationTests.java
OlgaMaciaszek: I have consulted with @spencergibb and he's approved the ordered list approach.

mamachanko:
@OlgaMaciaszek Unfortunately I have to abandon this PR for now due to a shift of priorities. I am really sorry about that. Hopefully, we will be able to pick it up again eventually. Thank you for your review and feedback! 🙇🏻

OlgaMaciaszek:
No worries, @mamachanko, thanks for letting us know.

Issue #4457, opened 2025-07-14
com.netflix.discovery.DiscoveryClient keeps using org.apache.commons.lang.exception.ExceptionUtils with commons-configuration2


nvalencia-veradat:
Hi!
For vulnerabilities we update from commons-configuration to commons-configuration2 but com.netflix.discovery.DiscoveryClient is using a reference to org.apache.commons.lang.exception.ExceptionUtils and throws a java.lang.ClassNotFoundException.
Are there any other configuration to update commons-lang dependency?.

**Sample**
com.netflix.discovery.DiscoveryClient    : Cannot fetch registry from server
 
java.lang.NoClassDefFoundError: org/apache/commons/lang/exception/ExceptionUtils
	at com.netflix.discovery.DiscoveryClient.fetchRegistry(DiscoveryClient.java:969) ~[eureka-client-2.0.4.jar:2.0.4]
	at com.netflix.discovery.DiscoveryClient.refreshRegistry(DiscoveryClient.java:1473) ~[eureka-client-2.0.4.jar:2.0.4]
	at com.netflix.discovery.DiscoveryClient$CacheRefreshThread.run(DiscoveryClient.java:1441) ~[eureka-client-2.0.4.jar:2.0.4]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572) ~[na:na]
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317) ~[na:na]
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:1583) ~[na:na]
Caused by: java.lang.ClassNotFoundException: org.apache.commons.lang.exception.ExceptionUtils
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641) ~[na:na]
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188) ~[na:na]
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526) ~[na:na]
[...]

Issue #4446, opened 2025-06-04
spring-cloud-starter-netflix-eureka-client:4.3.0 has vulnerability with dependency commons-configuration:1.10


ziad-saade:
![Image](https://github.com/user-attachments/assets/920e0b79-c087-49bc-9b5b-1ee5672539d1)

commons-configuration:1.10 has vulnerabilities:

![Image](https://github.com/user-attachments/assets/247aa9a7-48de-497d-a327-d50a01f30d6d)

https://mvnrepository.com/artifact/commons-configuration/commons-configuration/1.10

to fix the issue upgrade to commons-configuration2:2.12.0

@OlgaMaciaszek

OlgaMaciaszek:
Hello @ziad-saade, this is a transitive dependency not governed by our project, but brought about by Netflix/Eureka, which is an external project, not maintained by us. Switching to `commons-configuration2` conflicts with the [Netflix/Eureka]https://github.com/Netflix/archaius project which is used internally by [Netflix/Eureka](https://github.com/Netflix/eureka). Please create an issue in both of those projects and link here to track when it's done.

ph33rtehgd:
Apologies if I'm getting ahead of myself, but I created a PR for this: https://github.com/spring-cloud/spring-cloud-netflix/pull/4468

If this PR is accepted I think it should be merged back into 4.2.x and 4.3.x.

PR #4468, opened 2025-07-31
Exclude commons-configuration

ph33rtehgd:
This PR is meant to primarily resolve this issue: https://github.com/spring-cloud/spring-cloud-netflix/issues/4446

My comments in the issue give some more detailed analysis, but from what I see is that while commons-configuration is used by eureka-client there isn't a flow through Spring Cloud Netflix that actually utilizes any classes from eureka-client that will actually utilize commons-configuration. Given this dependency's age, the fact that it's EOL and the fact the Netflix very likely won't be updating this I believe the most prudent thing to do is exclude this dependency as it should not negatively impact this Spring module.

It should also resolve this issue: https://github.com/spring-cloud/spring-cloud-netflix/issues/4457

### DIFF
--- a/spring-cloud-netflix-dependencies/pom.xml
+++ b/spring-cloud-netflix-dependencies/pom.xml
@@ -75,6 +75,10 @@
 						<groupId>org.glassfish.jersey.core</groupId>
 						<artifactId>jersey-client</artifactId>
 					</exclusion>
+					<exclusion>
+						<artifactId>commons-configuration</artifactId>
+						<groupId>commons-configuration</groupId>
+					</exclusion>
 				</exclusions>
 			</dependency>
 			<dependency>


spencergibb:
Unfortunately, running the build reveals
```
 Caused by: java.lang.IllegalStateException: java.lang.NoClassDefFoundError: org/apache/commons/lang/exception/ExceptionUtils
 ```
So this will require a new release of eureka with my merged PR mentioned above

ph33rtehgd:
@spencergibb Got it, apologies, this is the first time I've raised a PR against an open source project, I thought it might actually go the other way :) (master down to older branches). Do you want me to close this PR and open a new one starting at 4.2.x (or if there's a different branch that's lower and still supported I can start there)?

As for the commons-lang dependency, yes, you're right that is needed. When I was testing out my change locally I had the latest Netflix 2.x branch that already contained your PR merged locally so I didn't notice it. Hopefully since your change is minor on Eureka that can get released in the near future and then we can merge this change.

cdprete:
> Unfortunately, running the build reveals
> 
> ```
>  Caused by: java.lang.IllegalStateException: java.lang.NoClassDefFoundError: org/apache/commons/lang/exception/ExceptionUtils
> ```
> 
> So this will require a new release of eureka with my merged PR mentioned above

There is indeed no fix for it from the Spring Cloud Netflix project.

I tried the same where I work, but then I stumbled on the same issue and I had to add `commons-lang` as dependency which is also vulnerable.
The only solution I think may work is to, during the build of `spring-cloud-netflix-eureka`, is to:

1. download the sources of `commons-lang`
2. unpack them
3. delete all the sources which are not needed (i.e.: keep only this `ExceptionUtils` ideally)
4. attach the new sources as additional sources to the build
5. package the JAR with the `commons-configuration` dependency excluded

cdprete:
> This PR has been merged and we just need a release of eureka [Netflix/eureka#1601](https://github.com/Netflix/eureka/pull/1601)

Just out of curiosity, why the string interpolation there has been replaced with `String.format`?
Like this, all the benefits to not evaluate the parameters if the used logging level is disabled are lost.

spencergibb:
Because the other doesn't take an exception as an argument

spencergibb:
TIL.

PR #4551, opened 2026-04-02
Add support for custom actuator health status mapping in EurekaHealthCheckHandler

yashdotdev13:
This PR introduces configurable mapping between Spring Boot Actuator health statuses and Eureka InstanceStatus.

Currently, custom actuator statuses are mapped to UNKNOWN by default. This change allows users to define mappings via configuration:

eureka.client.status.mapping.*

Example:
eureka.client.status.mapping.fatal=OUT_OF_SERVICE

Key points:
- Supports custom actuator health statuses
- Case-insensitive mapping
- Preserves existing default behavior as fallback
- Maintains backward compatibility

Tests:
- Added unit tests for custom status mapping
- Verified null safety handling
- All existing tests pass successfully

Fixes #4551

### DIFF
--- a/docs/modules/ROOT/pages/spring-cloud-netflix.adoc
+++ b/docs/modules/ROOT/pages/spring-cloud-netflix.adoc
@@ -389,134 +389,15 @@ the ability to refresh Eureka clients.  To do this set `eureka.client.refresh.en
 
 === Using Eureka with Spring Cloud LoadBalancer
 
-Spring Cloud Netflix integrates with Spring Cloud LoadBalancer to provide
-client-side load balancing for services registered with Eureka.
+We offer support for the Spring Cloud LoadBalancer `ZonePreferenceServiceInstanceListSupplier`.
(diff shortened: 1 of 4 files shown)


docs/modules/ROOT/pages/spring-cloud-netflix.adoc
```diff
@@ -389,15 +389,134 @@ the ability to refresh Eureka clients.  To do this set `eureka.client.refresh.en
 
 === Using Eureka with Spring Cloud LoadBalancer
 
-We offer support for the Spring Cloud LoadBalancer `ZonePreferenceServiceInstanceListSupplier`.
```
spencergibb: Why did you rewrite the documentation? Please revert to focus on one thing

spencergibb (changes requested):

98001yash:
You're right — that was unintended. I've reverted the unrelated documentation changes and kept the PR focused only on the status mapping feature.
Thanks for pointing that out.

Issue #4275, opened 2024-04-09
When shutting down after 4.1.1, an exception occurs while unregistering


heowc:
I recently uploaded the spring cloud version and started using spring-cloud-netflix 4.1.1.

Since then, whenever the application is deployed, `DiscoveryClient` is shut down and an exception is repeatedly reported indicating that deregistration failed.

The problem I think occurred while fixing the problem below.
- https://github.com/spring-cloud/spring-cloud-netflix/pull/4258

If I replace it with the new `EurekaClientHttpRequestFactorySupplier` it doesn't throw any exception. 
(Remove the code related to `AtomicReference<CloseableHttpClient>`)

The question I am curious about here is is there a reason to manage `CloseableHttpClient` separately rather than Spring Bean?

9you:
Same issue, attached some logs which may help diagnose the problem.
```
2024-04-09 16:14:11.719  INFO [] 42115 --- [ionShutdownHook] c.n.d.DiscoveryClient                    : Shutting down DiscoveryClient ...
2024-04-09 16:14:14.728  INFO [] 42115 --- [ionShutdownHook] c.n.d.DiscoveryClient                    : Unregistering ...
2024-04-09 16:14:14.730 DEBUG [] 42115 --- [ionShutdownHook] o.s.w.c.RestTemplate                     : HTTP DELETE http://localhost:8761/eureka/apps/API-GATEWAY/192.168.10.228:api-gateway:8082
2024-04-09 16:14:14.730 DEBUG [] 42115 --- [ionShutdownHook] o.s.w.c.RestTemplate                     : Accept=[application/json, application/*+json]
2024-04-09 16:14:14.733  INFO [] 42115 --- [ionShutdownHook] c.n.d.s.t.d.RedirectingEurekaHttpClient  : Request execution error. endpoint=DefaultEndpoint{ serviceUrl='http://localhost:8761/eureka/} exception=Connection pool shut down stacktrace=java.lang.IllegalStateException: Connection pool shut down
	at org.apache.hc.core5.util.Asserts.check(Asserts.java:38)
	at org.apache.hc.core5.pool.StrictConnPool.lease(StrictConnPool.java:176)
	at org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager.lease(PoolingHttpClientConnectionManager.java:298)
[...]

PR #4258, opened 2024-02-19
Enable Connection Evictor in HttpClient (#4103)

ZIRAKrezovic:
Fixes: #4103

Issue #4103, opened 2022-06-15
RestTemplateEurekaHttpClient not closing HttpClient on shutdown 


BenEfrati:
**Spring Cloud 2021.0.0
Spring Boot 2.6.8**

**Bug Description:**
`RestTemplateEurekaHttpClient` is not closing `HttpClient` on shutdown. This leads to TCP `CLOSE_WAIT` connections to eureka server.

`EurekaClient` will shutdown when an exception occurs on an http request, but not shutdown `HttpClient`.

This bug is related to https://github.com/spring-cloud/spring-cloud-netflix/blob/27ac3379f6b391ccd605d212512f15c439825ecb/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateEurekaHttpClient.java#L203

In case of exception here:
https://github.com/Netflix/eureka/blob/ed0da19ca1c049c87e3dbf75b6015c1861d5c2d0/eureka-client/src/main/java/com/netflix/discovery/shared/transport/decorator/RedirectingEurekaHttpClient.java#L96
new `HttpClient` will be created without closing the existing one - this causes `CLOSE_WAIT` connections

This supplier creates new `CloseableHttpClient` for every call to https://github.com/spring-cloud/spring-cloud-netflix/blob/27ac3379f6b391ccd605d212512f15c439825ecb/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/RestTemplateTransportClientFactory.java#L103

```
public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClientHttpRequestFactorySupplier {

	@Override
	public ClientHttpRequestFactory get(SSLContext sslContext, @Nullable HostnameVerifier hostnameVerifier) {
		HttpClientBuilder httpClientBuilder = HttpClients.custom();
		if (sslContext != null) {
			httpClientBuilder = httpClientBuilder.setSSLContext(sslContext);
		}
		if (hostnameVerifier != null) {
			httpClientBuilder = httpClientBuilder.setSSLHostnameVerifier(hostnameVerifier);
		}
		CloseableHttpClient httpClient = httpClientBuilder.build();
		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
		requestFactory.setHttpClient(httpClient);
		return requestFactory;
	}

}
```

[...]

PR #4585, opened 2026-08-14
gh-4569: Deterministically close shared HTTP client on TransportClientFactory#shutdown()

### LINKED ISSUE #4569: Eureka deregistration fails on graceful shutdown: "Connection pool shut down"
I am experiencing an issue where my Spring Boot microservices fail to deregister from Eureka during a graceful shutdown. Because the deregistration HTTP request fails,

During shutdown, it seems the HTTP connection pool is closed before the DiscoveryClient attempts to send the final DELETE request to the Eureka server, resulting in an IllegalStateException: Connection pool shut down.
Could you help me understand how to configure my applications so that the deregistration completes successfully before the connection pools are destroyed?

Environment
Spring Boot: 3.2.1
Spring Cloud: 2023.0.0 (eureka-client 2.0.2)
Deployment: Docker Swarm (using stop-first update config)

Configuration
Eureka Server (application.yml):
```
yaml
spring:
  application:
    name: service-registry
  cloud:
    inetutils:
      preferred-networks:
        - XXX.XXX.XXX
eureka:
  server:
    eviction-interval-timer-in-ms: 60000 
    response-cache-update-interval-ms: 10000 
  client:
    register-with-eureka: false
    fetch-registry: false
    service-url:
      defaultZone: http://${eureka.instance.hostname}:${server.port}/eureka/
  instance:
    hostname: localhost
```
Eureka Client / Microservice (eureka-properties.yml via Config Server):
```
eureka:
  client:
    service-url:
      defaultZone: ${EUREKA_SERVICE_ADDRESS:http://localhost:8761/eureka/}
    register-with-eureka: true
    fetch-registry: true
  instance:
    prefer-ip-address: true
    instance-id: ${spring.application.name}:${spring.application.instance_id:${random.value}}
spring:
  cloud:
    inetutils:
      preferred-networks:
        - XXX.XXX.XXX
```
Shutdown Logs
Here are the logs during the shutdown of one of the microservices (order-service). It shows the sequence of events leading up to the failure:

```
13:50:42.773  INFO --- [ionShutdownHook] o.s.c.n.e.s.EurekaServiceRegistry        : Unregistering application ORDER-SERVICE with eureka with status DOWN
13:50:42.773  INFO --- [ionShutdownHook] com.netflix.discovery.DiscoveryClient    : Saw local status change event StatusChangeEvent [timestamp=1779112242773, current=DOWN, previous=UP]
13:50:42.773  INFO --- [nfoReplicator-0] com.netflix.discovery.DiscoveryClient    : DiscoveryClient_ORDER-SERVICE/order-service:373b0e4bac...: registering service...
13:50:42.777  INFO --- [nfoReplicator-0] com.netflix.discovery.DiscoveryClient    : DiscoveryClient_ORDER-SERVICE/order-service:373b0e4bac... - registration status: 204
13:50:42.781  INFO --- [ionShutdownHook] j.LocalContainerEntityManagerFactoryBean : Closing JPA EntityManagerFactory for persistence unit 'default'
13:50:42.782  INFO --- [ionShutdownHook] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Shutdown initiated...
13:50:42.786  INFO --- [ionShutdownHook] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Shutdown completed.
13:50:42.786  INFO --- [ionShutdownHook] com.netflix.discovery.DiscoveryClient    : Shutting down DiscoveryClient ...
13:50:45.787  INFO --- [ionShutdownHook] com.netflix.discovery.DiscoveryClient    : Unregistering ...
13:50:45.791  INFO --- [ionShutdownHook] c.n.d.s.t.d.RedirectingEurekaHttpClient  : Request execution error. endpoint=DefaultEndpoint{ serviceUrl='http://registry-dev:8761/eureka/} exception=Connection pool shut down stacktrace=java.lang.IllegalStateException: Connection pool shut down
13:50:45.792  WARN --- [ionShutdownHook] c.n.d.s.t.d.RetryableEurekaHttpClient    : Request execution failed with message: Connection pool shut down
13:50:45.793 ERROR --- [ionShutdownHook] com.netflix.discovery.DiscoveryClient    : DiscoveryClient_ORDER-SERVICE/order-service:373b0e4bac... - de-registration failedCannot execute request on any known server
13:50:45.797  INFO --- [ionShutdownHook] com.netflix.discovery.DiscoveryClient    : Completed shut down of DiscoveryClient

```

[order.logs.txt](https://github.com/user-attachments/files/28010358/order.logs.txt)

Is there a specific configuration I am missing to ensure the deregistration happens before the connection pools are closed? Any guidance on how to resolve this would be greatly appreciated.

PRAHLAD09-dev:
Fixes #4569

### Problem

`RestClientTransportClientFactory.newClient()` builds a brand-new `CloseableHttpClient` (with its own dedicated connection pool) on every single call. Neither `TransportClientFactory#shutdown()` nor `EurekaHttpClient#shutdown()` ever closes these clients - cleanup relied entirely on GC. During graceful shutdown, the final `unregister()` DELETE call (made from `DiscoveryClient.shutdown()`, right after `cancelScheduledTasks()`) can hit a connection pool that's already been torn down, throwing:

    java.lang.IllegalStateException: Connection pool shut down

This causes deregistration to fail silently during graceful shutdown, leaving stale instances in the Eureka registry until the lease expires.

### History

This overlaps with #4103, which documented the same "new HTTP client per call, never closed" issue. That was addressed in #4258 by caching a shared `CloseableHttpClient` in `DefaultEurekaClientHttpRequestFactorySupplier` and making the supplier a Spring `DisposableBean` to close it on shutdown. That approach was reverted (79a2eb8c4) after it caused #4275: making the supplier a `DisposableBean` introduced an *independent* Spring bean-destroy callback with no ordering guarantee relative to `CloudEurekaClient`'s own `@Bean(destroyMethod = "shutdown")` - so the shared client could be closed *before* `CloudEurekaClient.shutdown()` reached its `unregister()` call, breaking deregistration outright.

### Fix

Same idea as #4258 - share one `CloseableHttpClient` per supplier instance instead of building one per call - but close it through the Eureka transport lifecycle instead of an independent Spring bean-destroy callback:

- `EurekaClientHttpRequestFactorySupplier` gets a new `default void close() {}` method (backward compatible for existing custom implementations).
- `DefaultEurekaClientHttpRequestFactorySupplier` lazily builds and caches a single `CloseableHttpClient`, reused across all `get()` calls, and closes it in `close()`. It does **not** implement `DisposableBean`.
- `RestClientTransportClientFactory#shutdown()` now calls `eurekaClientHttpRequestFactorySupplier.close()`.

Since `TransportClientFactory#shutdown()` is invoked synchronously by Netflix's `DiscoveryClient.shutdown()` - which calls `unregister()` before `eurekaTransport.shutdown()` - the pool is now guaranteed to close *after* the final deregistration request completes, in the same thread, with no race against an unrelated Spring bean-destroy path.

### Testing

- `DefaultEurekaClientHttpRequestFactorySupplierTests`: verifies the client is reused across `get()` calls, `close()` is safe to call before any `get()` and safe to call twice, and - as an explicit regression guard for #4275 - that this class does not implement `DisposableBean`.
- `RestClientTransportClientFactoryShutdownTests`: verifies `shutdown()` delegates to `supplier.close()`.
- Full `spring-cloud-netflix-eureka-client` test suite passes locally (216/216, excluding one pre-existing Docker-dependent Testcontainers test unrelated to this change).

### DIFF
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
@@ -55,9 +55,7 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	private final Set<RequestConfigCustomizer> requestConfigCustomizers;
 
-	private final Object lock = new Object();
-
-	private volatile CloseableHttpClient sharedHttpClient;
+	private CloseableHttpClient sharedHttpClient;
 
 	public DefaultEurekaClientHttpRequestFactorySupplier(TimeoutProperties timeoutProperties,
 			Set<RequestConfigCustomizer> requestConfigCustomizers) {
@@ -66,20 +64,18 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 	}
 
 	@Override
-	public ClientHttpRequestFactory get(SSLContext sslContext, @Nullable HostnameVerifier hostnameVerifier) {
-		CloseableHttpClient httpClient;
-		synchronized (this.lock) {
-			httpClient = this.sharedHttpClient;
-			if (httpClient == null) {
-				HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
-				if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
-					httpClientBuilder
-						.setConnectionManager(buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
-				}
-				httpClientBuilder.setDefaultRequestConfig(buildRequestConfig());
-				httpClient = httpClientBuilder.build();
-				this.sharedHttpClient = httpClient;
+	public synchronized ClientHttpRequestFactory get(SSLContext sslContext,
+			@Nullable HostnameVerifier hostnameVerifier) {
+		CloseableHttpClient httpClient = this.sharedHttpClient;
+		if (httpClient == null) {
+			HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
+			if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
+				httpClientBuilder
+					.setConnectionManager(buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
 			}
+			httpClientBuilder.setDefaultRequestConfig(buildRequestConfig());
+			httpClient = httpClientBuilder.build();
+			this.sharedHttpClient = httpClient;
 		}
 		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
 		requestFactory.setHttpClient(httpClient);
@@ -87,17 +83,15 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 	}
 
 	@Override
-	public void close() {
-		synchronized (this.lock) {
-			CloseableHttpClient httpClient = this.sharedHttpClient;
-			this.sharedHttpClient = null;
-			if (httpClient != null) {
-				try {
-					httpClient.close();
-				}
-				catch (IOException ex) {
-					// best-effort close during shutdown; nothing actionable if it fails
-				}
+	public synchronized void close() {
+		CloseableHttpClient httpClient = this.sharedHttpClient;
+		this.sharedHttpClient = null;
+		if (httpClient != null) {
+			try {
+				httpClient.close();
+			}
+			catch (IOException ex) {
+				// best-effort close during shutdown; nothing actionable if it fails
 			}
 		}
 	}
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
@@ -55,9 +55,9 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	private final Set<RequestConfigCustomizer> requestConfigCustomizers;
 
-	private volatile CloseableHttpClient sharedHttpClient;
-
 	private final Object lock = new Object();
+
+	private volatile CloseableHttpClient sharedHttpClient;
 
 	public DefaultEurekaClientHttpRequestFactorySupplier(TimeoutProperties timeoutProperties,
 			Set<RequestConfigCustomizer> requestConfigCustomizers) {
@@ -67,20 +67,18 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	@Override
 	public ClientHttpRequestFactory get(SSLContext sslContext, @Nullable HostnameVerifier hostnameVerifier) {
-		CloseableHttpClient httpClient = this.sharedHttpClient;
-		if (httpClient == null) {
-			synchronized (this.lock) {
-				httpClient = this.sharedHttpClient;
-				if (httpClient == null) {
-					HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
-					if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
-						httpClientBuilder.setConnectionManager(
-								buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
-					}
-					httpClientBuilder.setDefaultRequestConfig(buildRequestConfig());
-					httpClient = httpClientBuilder.build();
-					this.sharedHttpClient = httpClient;
+		CloseableHttpClient httpClient;
+		synchronized (this.lock) {
+			httpClient = this.sharedHttpClient;
+			if (httpClient == null) {
+				HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
+				if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
+					httpClientBuilder
+						.setConnectionManager(buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
 				}
+				httpClientBuilder.setDefaultRequestConfig(buildRequestConfig());
+				httpClient = httpClientBuilder.build();
+				this.sharedHttpClient = httpClient;
 			}
 		}
 		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
@@ -90,13 +88,16 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	@Override
 	public void close() {
-		CloseableHttpClient httpClient = this.sharedHttpClient;
-		if (httpClient != null) {
-			try {
-				httpClient.close();
-			}
-			catch (IOException ex) {
-				// best-effort close during shutdown; nothing actionable if it fails
+		synchronized (this.lock) {
+			CloseableHttpClient httpClient = this.sharedHttpClient;
+			this.sharedHttpClient = null;
+			if (httpClient != null) {
+				try {
+					httpClient.close();
+				}
+				catch (IOException ex) {
+					// best-effort close during shutdown; nothing actionable if it fails
+				}
 			}
 		}
 	}
--- a/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
+++ b/spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.netflix.eureka.http;
 
+import java.io.IOException;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
@@ -34,6 +35,7 @@ import org.apache.hc.core5.http.io.SocketConfig;
 import org.apache.hc.core5.util.Timeout;
 
 import org.springframework.cloud.netflix.eureka.TimeoutProperties;
+import org.springframework.cloud.netflix.eureka.http.EurekaClientHttpRequestFactorySupplier.RequestConfigCustomizer;
 import org.springframework.http.client.ClientHttpRequestFactory;
 import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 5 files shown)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -61,19 +67,40 @@ public DefaultEurekaClientHttpRequestFactorySupplier(TimeoutProperties timeoutPr
...
-		CloseableHttpClient httpClient = httpClientBuilder.build();
 		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
 		requestFactory.setHttpClient(httpClient);
 		return requestFactory;
 	}
 
+	@Override
+	public void close() {
+		CloseableHttpClient httpClient = this.sharedHttpClient;
+		if (httpClient != null) {
```
ryanjbaxter, line 89: should the lock be used here as well?

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -53,6 +55,10 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	private final Set<RequestConfigCustomizer> requestConfigCustomizers;
 
+	private volatile CloseableHttpClient sharedHttpClient;
```
ryanjbaxter: I am a little concerned about this in 2 cases:

First the refresh case when `eureka.client.refresh.enable=true`.  A refresh would close his but never nulls it out.

Second when if a second Eureka client is created, then it would reuse this sharedHttpClient, if one calls close if destroys the second instance.

ryanjbaxter (changes requested):

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
PRAHLAD09-dev: Fixed — close() now runs inside synchronized (this.lock), same as get().

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
PRAHLAD09-dev: Good catch. The refresh case is now handled by clearing sharedHttpClient after closing it, so a subsequent get() creates a fresh client instead of reusing the closed instance.

For the second scenario, where multiple Eureka clients might share the same supplier instance, I wanted to confirm whether that can actually happen with the current RestClientTransportClientFactory lifecycle before introducing additional reference-counting/lifecycle management.

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplierTests.java
```diff
@@ -0,0 +1,92 @@
...
+		supplier.close();
+	}
+
+	@Test
+	void getAfterCloseShouldStillReturnARequestFactory() {
+		supplier.get(null, null);
+		supplier.close();
+
+		// A get() call racing just after shutdown must not throw; the returned factory
+		// wraps a closed client and will fail on actual use, which is expected during
```
ryanjbaxter, line 86: I don't think this is true anymore since the shared client is nulled

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
ryanjbaxter: > For the second scenario, where multiple Eureka clients might share the same supplier instance, I wanted to confirm whether that can actually happen with the current RestClientTransportClientFactory lifecycle before introducing additional reference-counting/lifecycle management.

It is not something we do out of the box, but in theory someone could create a second client on their own and that case they would run into a problem if one of those clients gets closed while the other one is still being used.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -53,6 +55,10 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	private final Set<RequestConfigCustomizer> requestConfigCustomizers;
 
+	private final Object lock = new Object();
+
+	private volatile CloseableHttpClient sharedHttpClient;
```
wind57: what is the point to make it `volatile` here? you are reading and writing it under the _same_ lock, and the JLS already guarantees visibility.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -61,19 +67,41 @@ public DefaultEurekaClientHttpRequestFactorySupplier(TimeoutProperties timeoutPr
...
-
-		CloseableHttpClient httpClient = httpClientBuilder.build();
 		HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory();
 		requestFactory.setHttpClient(httpClient);
 		return requestFactory;
 	}
 
+	@Override
+	public void close() {
+		synchronized (this.lock) {
```
wind57: this is overly complicated also, I think you are confusing what `volatile` is supposed to do and this confusion is "dragged" into this method: that is why you do a volatile read via the _local_ `httpClient`. But again, this is not correct, you do not even need it, and the method can be simplified to : 

```
@Override
public void close() {
	synchronized (this.lock) {
		if (this.sharedHttpClient == null) {
			return;
		}

		try {
			this.sharedHttpClient.close();
		}
		catch (IOException ex) {
			// Best-effort close during shutdown.
		}
		finally {
			this.sharedHttpClient = null;
		}
	}
}
```

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -61,19 +67,41 @@ public DefaultEurekaClientHttpRequestFactorySupplier(TimeoutProperties timeoutPr
 
 	@Override
 	public ClientHttpRequestFactory get(SSLContext sslContext, @Nullable HostnameVerifier hostnameVerifier) {
-		HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
-		if (sslContext != null || hostnameVerifier != null || timeoutProperties != null) {
-			httpClientBuilder
-				.setConnectionManager(buildConnectionManager(sslContext, hostnameVerifier, timeoutProperties));
+		CloseableHttpClient httpClient;
```
wind57: same goes for this method, it can be simplified: 

```
@Override
public ClientHttpRequestFactory get(
		SSLContext sslContext,
		@Nullable HostnameVerifier hostnameVerifier) {

	synchronized (this.lock) {
		if (this.sharedHttpClient == null) {
			HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();

			if (sslContext != null
					|| hostnameVerifier != null
					|| timeoutProperties != null) {
				httpClientBuilder.setConnectionManager(
						buildConnectionManager(
								sslContext,
								hostnameVerifier,
								timeoutProperties));
			}

			this.sharedHttpClient = httpClientBuilder.build();
		}

		return new HttpComponentsClientHttpRequestFactory(
				this.sharedHttpClient);
	}
}
```

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
```diff
@@ -53,6 +55,10 @@ public class DefaultEurekaClientHttpRequestFactorySupplier implements EurekaClie
 
 	private final Set<RequestConfigCustomizer> requestConfigCustomizers;
 
+	private final Object lock = new Object();
```
wind57: and now, if you look at the other comments, you will realize that this is not even needed anymore, you can simply make the methods synchronized...

PRAHLAD09-dev: Good point. I simplified the synchronization by making get() and close() synchronized methods and removed the explicit lock.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/http/DefaultEurekaClientHttpRequestFactorySupplier.java
ryanjbaxter: This is still a concern of mine

PRAHLAD09-dev: Thanks for clarifying. I understand the concern around a supplier instance being shared by multiple Eureka clients. Since this is not the default lifecycle but is still a possible use case, would you prefer us to handle shared supplier instances explicitly, for example with reference counting, or document this as a limitation?

ryanjbaxter: The right fix IMO is to actually no cache the `ClosableHttpClient` in `EurekaClientHttpRequestFactorySupplier` and instead put it in `RestClientTransportClientFactory` which is already creaed whenever `newTransportClientFactory()` is called.

PRAHLAD09-dev: Done — moved the caching to RestClientTransportClientFactory as suggested. The supplier is now completely stateless (close() removed from the interface entirely), and RestClientTransportClientFactory lazily builds and owns its own cached client, closing it directly in shutdown(). This resolves both the refresh and multi-instance sharing concerns since there's no more shared mutable state in the supplier.

Updated tests accordingly — full suite passes locally (216/216, excluding the pre-existing Docker-dependent test noted earlier).

ryanjbaxter (changes requested):
This looks good!  Thanks.

One small change, RestClientTransportClientFactoryShutdownTests should verify that after shutdown is called that a new instance is created.

PR #4602, opened 2026-09-13
Exclude @LoadBalanced RestClient/WebClient builders from Eureka transport

### LINKED ISSUE #4591: @LoadBalanced` with `RestClient` — the problem and the fix
# `@LoadBalanced` with `RestClient` — the problem and the fix

**Spring Boot 4.1.1 · Spring Cloud 2025.1.2 · Eureka**

## The problem

You add a `@LoadBalanced RestClient.Builder`, expecting it to work the way `@LoadBalanced RestTemplate`
always did. Instead your application never registers with Eureka, and the log repeats:

```
java.lang.IllegalStateException: No instances available for localhost
WARN  ... DiscoveryClient : registration failed Cannot execute request on any known server
ERROR ... DiscoveryClient : was unable to send heartbeat!
```

The application still starts, which is what makes it confusing — it just silently never appears in
the registry.

## Why it happens

The Eureka client needs an HTTP client for its own calls to the registry, and it looks for a
`RestClient.Builder` in your application context:

```java
// DiscoveryClientOptionalArgsConfiguration.RestClientConfiguration
ObjectProvider<RestClient.Builder> restClientBuilderProvider
...
() -> restClientBuilderProvider.getIfAvailable(RestClient::builder);
```

If the only `RestClient.Builder` bean you have is the `@LoadBalanced` one, Eureka finds and uses it.
Its calls to `http://localhost:8761/eureka` then go through the load balancer, which tries to resolve
`localhost` as a service id and fails.

Two details make this easy to hit:

- `LoadBalancerRestClientBuilderBeanPostProcessor` has already added the load balancer interceptor to
  that builder instance, so anything resolving the bean inherits the interceptor.
- Declaring *any* `RestClient.Builder` bean suppresses Boot's auto-configured one
  (`RestClientAutoConfiguration#restClientBuilder` is `@ConditionalOnMissingBean`), which removes the
  plain fallback candidate and leaves the load-balanced builder as the only match.

## The fix

Keep a second, plain `@Primary` builder so the Eureka transport has an ordinary candidate to find,
and ask for the load-balanced one explicitly by qualifier:

```java
@Configuration
public class RestClientConfig {

    /** Plain and @Primary — this is the one the Eureka transport picks up. */
    @Bean
    @Primary
    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
    public RestClient.Builder restClientBuilder(RestClientBuilderConfigurer configurer) {
        return configurer.configure(RestClient.builder());
    }

    /** Load-balanced — gets the interceptor from the bean post processor. */
    @Bean
    @LoadBalanced
    public RestClient.Builder loadBalancedRestClientBuilder(RestClientBuilderConfigurer configurer) {
        return configurer.configure(RestClient.builder());
    }

    /** Note the @LoadBalanced on the parameter — see rule 2 below. */
    @Bean
    public RestClient serviceARestClient(@LoadBalanced RestClient.Builder builder) {
        return builder.baseUrl("http://SERVICE-A").build();
    }
}
```

With this in place, registration succeeds and requests round-robin correctly across instances.

## Two rules to remember

**1. Don't delete the plain builder.** It looks like dead code. It is the entire fix — it exists so
the Eureka transport has something non-load-balanced to resolve.

**2. `@LoadBalanced` must also be on the injection point.**

```java
// WRONG — compiles, runs, quietly does no load balancing
public RestClient serviceARestClient(RestClient.Builder loadBalancedRestClientBuilder) { ... }

// RIGHT
public RestClient serviceARestClient(@LoadBalanced RestClient.Builder builder) { ... }
```

Naming the parameter after the bean does not select it. `DefaultListableBeanFactory#determineAutowireCandidate`
checks for a `@Primary` candidate **first** and only falls back to parameter-name matching afterwards,
so the plain builder wins. `@LoadBalanced` is itself a `@Qualifier`, which is why putting it on the
parameter works.

## Why `RestTemplate` never had this problem

| | `@LoadBalanced RestTemplate` | `@LoadBalanced RestClient.Builder` |
| --- | --- | --- |
| Load balancing works | yes | yes |
| Eureka registration affected | no | **yes** — needs the extra plain builder |
| Why | Eureka's transport doesn't consume `RestTemplate` beans | Eureka's transport resolves `RestClient.Builder` by type |

`WebClient.Builder` is resolved the same way as `RestClient.Builder`, but only when
`eureka.client.webclient.enabled=true`, so a default setup never reaches that path.

## Dependency note

`RestClientBuilderConfigurer` is `org.springframework.boot.restclient.autoconfigure.RestClientBuilderConfigurer`
and comes from the `spring-boot-restclient` module. On Boot 4, `spring-boot-starter-webmvc` does **not**
bring it in, so add:

```xml
<dependency>
    <artifactId>spring-boot-starter-restclient</artifactId>
</dependency>
```

Without it there is no auto-configured `RestClient.Builder` at all, and the snippet above won't compile.

## Upstream

Reported in [spring-cloud-netflix#4524](https://github.com/spring-cloud/spring-cloud-netflix/issues/4524)
and [spring-cloud-commons#1633](https://github.com/spring-cloud/spring-cloud-commons/issues/1633).
Both open at the time of writing — check them before assuming this workaround is still needed.

arimu1:
## Summary

- Resolve Eureka registry `RestClient.Builder` / `WebClient.Builder` candidates by excluding beans annotated with `@LoadBalanced`, so calls to the Eureka server URL are not routed through the load balancer (fixes `No instances available for localhost` when only a load-balanced builder is registered).
- Fall back to `RestClient.builder()` / `WebClient.builder()` when every registered builder is load-balanced.
- Add context tests covering RestClient-only, plain+load-balanced, and WebClient-only setups.

Fixes gh-4591 (same root cause as gh-4524).

## Test plan

- [x] `./mvnw -pl spring-cloud-netflix-eureka-client -Dtest=LoadBalancedEurekaHttpClientBuilderConfigurationTests test` (JDK 21)
- [x] `./mvnw -pl spring-cloud-netflix-eureka-client -am test` — module suite passes except pre-existing `EurekaConfigServerBootstrapperIntegrationTests` (Docker not available locally)


spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaClientBuilderSuppliers.java
```diff
@@ -0,0 +1,56 @@
...
+		return () -> resolveNonLoadBalancedBuilder(beanFactory, RestClient.Builder.class, RestClient::builder);
+	}
+
+	static Supplier<WebClient.Builder> webClientBuilder(ConfigurableListableBeanFactory beanFactory) {
+		return () -> resolveNonLoadBalancedBuilder(beanFactory, WebClient.Builder.class, WebClient::builder);
+	}
+
+	private static <T> T resolveNonLoadBalancedBuilder(ConfigurableListableBeanFactory beanFactory, Class<T> type,
+			Supplier<T> fallback) {
+		for (String name : beanFactory.getBeanNamesForType(type)) {
```
ryanjbaxter, line 48: The selected bean ignores `@Order`/`@Priority` of beans

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaClientBuilderSuppliers.java
```diff
@@ -0,0 +1,56 @@
...
+
+	static Supplier<RestClient.Builder> restClientBuilder(ConfigurableListableBeanFactory beanFactory) {
+		return () -> resolveNonLoadBalancedBuilder(beanFactory, RestClient.Builder.class, RestClient::builder);
+	}
+
+	static Supplier<WebClient.Builder> webClientBuilder(ConfigurableListableBeanFactory beanFactory) {
+		return () -> resolveNonLoadBalancedBuilder(beanFactory, WebClient.Builder.class, WebClient::builder);
+	}
+
+	private static <T> T resolveNonLoadBalancedBuilder(ConfigurableListableBeanFactory beanFactory, Class<T> type,
```
ryanjbaxter, line 46: This only needs to happen once, and not once right now it happens on every EurekaClient creation

spring-cloud-netflix-eureka-client/src/test/java/org/springframework/cloud/netflix/eureka/config/LoadBalancedEurekaHttpClientBuilderConfigurationTests.java
```diff
@@ -0,0 +1,136 @@
...
+					.getBean(RestClientDiscoveryClientOptionalArgs.class);
+				RestClient.Builder loadBalancedBuilder = context.getBean("loadBalancedRestClientBuilder",
+						RestClient.Builder.class);
+				Supplier<RestClient.Builder> supplier = getRestClientBuilderSupplier(args);
+				assertThat(supplier.get()).isNotSameAs(loadBalancedBuilder);
+			});
+	}
+
+	@Test
+	void restClientTransportPrefersPlainRestClientBuilderOverLoadBalanced() {
```
ryanjbaxter, line 59: WebClient should have a similar test

ryanjbaxter (changes requested):

arimu1:
Thanks for the review @ryanjbaxter — addressed all three points in eb419ef:

1. **@Order / @Priority** — After filtering out `@LoadBalanced` candidates, the remaining bean names are sorted with `DefaultListableBeanFactory.getOrder(beanName)` (with `@Order` / `@Priority` / `Ordered` fallback when the factory is not a `DefaultListableBeanFactory`), so the highest-precedence non–load-balanced builder is chosen.

2. **Resolve once** — `restClientBuilder` / `webClientBuilder` now resolve the selected builder when the supplier is created and return a supplier that always hands back that same instance (callers still `.clone()` per transport client as before).

3. **WebClient test** — Added `webClientTransportPrefersPlainWebClientBuilderOverLoadBalanced`, mirroring the existing RestClient plain-vs-load-balanced test.

**Tests:** `./mvnw -pl spring-cloud-netflix-eureka-client -Dtest=LoadBalancedEurekaHttpClientBuilderConfigurationTests test` — 4 tests, BUILD SUCCESS (Temurin 21).

arimu1:
Force-pushed amended commit (removed erroneous trailer): tip is now `3f39a858c` — same changes as noted above.

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaClientBuilderSuppliers.java
```diff
@@ -0,0 +1,102 @@
...
+		if (candidates.size() == 1) {
+			return candidates.get(0);
+		}
+		candidates.sort(
+				Comparator.comparingInt((String name) -> orderFor(beanFactory, name)).thenComparing(String::compareTo));
+		return candidates.get(0);
+	}
+
+	private static int orderFor(ConfigurableListableBeanFactory beanFactory, String beanName) {
+		if (beanFactory instanceof DefaultListableBeanFactory defaultListableBeanFactory) {
```
ryanjbaxter, line 84: This will cause the bean to be instantiated.  IMO we can just use `beanFactory.findAnnotationOnBean` here and if there isn't an annotation then treat it as `Ordered.LOWEST_PRECEDENCE`

spring-cloud-netflix-eureka-client/src/main/java/org/springframework/cloud/netflix/eureka/config/EurekaClientBuilderSuppliers.java
```diff
@@ -0,0 +1,102 @@
...
+		return () -> builder;
+	}
+
+	static Supplier<WebClient.Builder> webClientBuilder(ConfigurableListableBeanFactory beanFactory) {
+		WebClient.Builder builder = resolveNonLoadBalancedBuilder(beanFactory, WebClient.Builder.class,
+				WebClient::builder);
+		return () -> builder;
+	}
+
+	private static <T> T resolveNonLoadBalancedBuilder(ConfigurableListableBeanFactory beanFactory, Class<T> type,
```
ryanjbaxter, line 58: This will end up picking one non-loadbalanced bean when there is more than one and there is no `@Order/@Priority/@Primary`.  IMO in this case we should throw a `NoUniqueBeanDefinitionException`

ryanjbaxter (changes requested):

